PR #6, opened 2015-06-14
generate a client for a swagger spec

casualjim:
Like the server this needs to support an entire swagger spec.
Unlike the server this doesn't need an untyped version.


casualjim:
This branch will be sleeping for a while.
If you wanted experimental, this is it. It will generate a client for simple things.

In the mean time I'm working on the branch template-refactor, where I'm completing the generation of schema's to more idiomatic go. This is a bit more complex than I had anticipated, anyway it will go something like this:

* defintion of primitive => type alias/name
* defintion of array => type alias/name
* definition of map => type alias/name
* definition of object with properties => struct
* definition of ref => type alias/name
* object with only additional properties => map[string]T
* object with additional properties and properties => custom serializer
* schema with schema array in items => tuple (struct with properties {p1,p2,p3,...}, custom serializer)
* schema with all of => struct
  * all of schema with ref => embedded value
  * all of schema with properties => properties are included in struct
  * adding an all of schema with just "x-isnullable": true turns the schema into a pointer
* schema with additional items and a single schema, nothing special happens it's an invalid schema
* schema with additional items and multiple schema's in items => struct (2 properties with different types, custom serializer)
		

I've got quite a bit of this working on that branch, but it breaks other templates at this moment.
I have the custom structs, serializers and validators for this

casualjim:
A quick update:

I'm finally through most of the work for schemas. Currently adding support for enum validations in all allowed positions.  
supporting additionalItems and additionalProperties specifically across the board has been less than trivial. 
So I hope to finally be able to merge and close this issue in a couple of days

c4milo:
nice!, I'm really looking forward to having this generator.

PR #25, opened 2015-06-27
Fix -target path injected at end of operations instead of beginning

gbjk:
I found that when I used -target, I was still getting restapi stuff ditched into my gopath.

I believe the fix below covers that issue.


casualjim:
Thanks a lot, there are 2 test failures here. 
One is related to the path change. 
The other one I don't know why it does that but might be because of the same reason.

PR #31, opened 2015-07-07
removes `configureapi.gotmpl`

midoblgsm:
Hi,

In this PR we included the change to just remove the `configureapi.gotmpl` since when moved it to `main.gotmpl` the server generation will still create a `configureapi.go` which is redundant.

Of course, you may have other plans for `configureapi.gotmpl` in addition to including the function `configureAPI`. If that's the case then feel free to close this PR, otherwise please merge this as well.

Note if you accept both PRs then we may need to rebase this one and resubmit since this one includes both changes.

We are flexible, just let us know your preference.

Cheers.


casualjim:
I think this is no longer an issue so I'm going to close this one

PR #34, opened 2015-07-15
Fix package path handling on Windows

magnushiie:
The package handling code currently assumes that the path separator is the slash, and also that Go package paths and file paths use the same separator. This PR has small fixes go make it work on Windows but I wonder if there is some better way (e.g. something from go/build) to handle the imports.


magnushiie:
Seems go generate was not run after the package namespace change from casualjim->go-swagger.

casualjim:
I think somebody else submitted that PR earlier (the go generate bindata one)

magnushiie:
Sorry, yes, of course, I didn't check after rebasing (and the generated files were dirty because of line endings in the encoded data). Removed that commit now.

casualjim:
thanks!

PR #78, opened 2015-10-26
Allow http status 201 to have a response body

florindragos:
201 Responses should be able to contain a body. There is no rule enforcing HTTP 201 status code to return an empty body. The current code does not allow a 201 response to contain a body.

According to HTTP 1.1 spec:
>201 Created

>The request has been fulfilled and resulted in a new resource being created. The newly created resource can be referenced by the URI(s) returned in the entity of the response, with the most specific URI for the resource given by a Location header field. The response SHOULD include an entity containing a list of resource characteristics and location(s) from which the user or user agent can choose the one most appropriate. The entity format is specified by the media type given in the Content-Type header field. The origin server MUST create the resource before returning the 201 status code. If the action cannot be carried out immediately, the server SHOULD respond with 202 (Accepted) response instead.


casualjim:
thanks!

PR #157, opened 2015-12-07
Support commonly used x-nullable in addition to x-isnullable

ecnahc515:


casualjim:
can you add a test in the type resolvers tests that makes sure the isNullable method works for both keys?
Mostly to have this behavior described by tests so that people don't accidentally delete it in the future as part of unrelated work.

Issue #153, opened 2015-12-04
Add appveyor build


casualjim:
Also build on appveyor to ensure windows also works.

casualjim:
Added an appveyor build. 
The build runs but there are all kinds of test failures at this point. https://ci.appveyor.com/project/casualjim/go-swagger

I think some have to do because of the filepath helper usage when resolving refs.

Please help with getting this tool to work on windows if you require it

casualjim:
thanks to @faguirre1 for making the builds work on windows!

PR #164, opened 2015-12-11
fix url path generation in windows are using backslash instead of  /

faguirre1:


casualjim:
:+1: Thanks a lot  

contributes to https://github.com/go-swagger/go-swagger/issues/153

There are probably a few more issues for running this on windows

PR #168, opened 2015-12-14
fix test failing on windows.

faguirre1:
replaced some hardcoded ":" to Listseparator and changed some filepath to path (slash path)


casualjim:
Thanks :+1:

PR #173, opened 2015-12-16
Add default value support for response headers, fixes #172

### LINKED ISSUE #172: Add server support for default header values in responses
If response header's default value is specified in spec, generated server should have code to set this default.

aleksandr-vin:
New response objects should be created with `NewXXXX()` function call
instead of `XXXX{}` struct instantiation. Then default values for
headers will be pre filled.

Signed-off-by: Aleksandr Vinokurov <aleksandr.vin@gmail.com>


casualjim:
thanks @aleksandr-vin :+1:

PR #177, opened 2015-12-23
Refactor strfmt/time tests, add strfmt.NewDateTime function

aleksandr-vin:
* Tests in strfmt/time_test.go were separated to logical groups
* Excessive tests were removed
* Some assertions were normalized

* strfmt.NewDateTime function was added to represent zero value for DateTime
* strfmt.DateTime UnmarshalText was reduced, delegating decision to ParseDateTime


casualjim:
thanks :+1:

Issue #196, opened 2016-01-07
Generated BindRequest has call to Validate, even if no validation is specified


rossnanop:
Say I have a simple specification:

```yaml
swagger: '2.0'
info:
  title: Validation Issue
  description: Endpoints
  version: 0.0.1
basePath: /v1
consumes:
  - application/json
produces:
  - application/json

paths:
  /events:
    post:
      parameters:
        - name: event
          in: body
          description: Event to post
          required: true
          schema:
            $ref: '#/definitions/Event'
      responses:
        '201':
          $ref: '#/responses/Created'

definitions:
  Event:
    type: string

responses:
  CreatedEventResponse:
    description: Successfully created an event
    schema:
      $ref: '#/definitions/Event'

```
A Validate method is not generated for the Event model; however, it is referenced in the BindRequest method for the path that uses it, even though I did not specify validation i.e. body.Validate(route.Formats).

Generated models/event.go
```go
package models

// This file was generated by the swagger tool.
// Editing this file might prove futile when you re-run the swagger generate command

/*Event Event event

swagger:model Event
*/
type Event string
```

Generated restapi/operations/post_events_parameters.go
```go
package operations

// This file was generated by the swagger tool.
// Editing this file might prove futile when you re-run the swagger generate command

import (
	"net/http"

	"github.com/go-swagger/go-swagger/errors"
	"github.com/go-swagger/go-swagger/httpkit/middleware"

	"github.com/rossnanop/swagger-issue/models"
)

// NewPostEventsParams creates a new PostEventsParams object
// with the default values initialized.
func NewPostEventsParams() PostEventsParams {
	return PostEventsParams{}
}

// PostEventsParams contains all the bound params for the post events operation
// typically these are obtained from a http.Request
//
// swagger:parameters PostEvents
type PostEventsParams struct {
	/*Event to post
	  Required: true
	  In: body
	*/
	Event models.Event
}

// BindRequest both binds and validates a request, it assumes that complex things implement a Validatable(strfmt.Registry) error interface
// for simple values it will use straight method calls
func (o *PostEventsParams) BindRequest(r *http.Request, route *middleware.MatchedRoute) error {
	var res []error

	var body models.Event
	if err := route.Consumer.Consume(r.Body, &body); err != nil {
		res = append(res, errors.NewParseError("event", "body", "", err))
	} else {
		if err := body.Validate(route.Formats); err != nil {
			res = append(res, err)
		}

		if len(res) == 0 {
[...]

PR #218, opened 2016-01-20
Don't validate base go types

### LINKED ISSUE #217: Go swagger does not support plain values in bodies.
The following `swagger.yml` creates invalid server code when using `swagger generate server`. From what I can tell this is a valid spec according to the swagger spec and tested using both the online swagger editor/viewer as well as the provided go-swagger `validate` command. 

```yml
swagger: '2.0'
info:
  version: 1.0.0
  title: 'string-body-api'
paths:
  /echo:
    post:
      summary: Post a string, get the same string back.
      parameters:
        - name: body
          in: body
          description: this argument is a string
          schema:
            type: string
      responses:
        '200':
          description: this return value is a string
          schema:
            type: string
```
When running the generated code (no changes) the following error occurs:
```
$ go run cmd/string-body-server/*.go
# ./restapi/operations
restapi/operations/post_echo_parameters.go:40: body.Validate undefined (type string has no field or method Validate)
```

Looking at that file the error becomes apparent

```go
var body string
if err := route.Consumer.Consume(r.Body, &body); err != nil {
  res = append(res, errors.NewParseError("body", "body", "", err))
} else {
  if err := body.Validate(route.Formats); err != nil { // <- Calling validate without defining it
    res = append(res, err)
  }

  if len(res) == 0 {
    o.Body = &body
  }
}
```

Not sure what the best fix is here. We could define a custom type and add a Validate method, but that may make it painful to deal with non string types being passed around. Alternatively we could possibly turn off validation for plain types like strings and numbers. 

_Note that the above example uses `string`, but `number` and `boolean` have the same issue._

cozmo:
This appears to fix #217. Should probably add some tests, but I haven't looked into them in depth yet. 

This causes a regression on #196 which makes sense since the example in that issue is type `string` and  as per the changes here non complex objects won't get validated. 

One thing I'm still trying to figure out how to do is check the type of array items, since currently this fails if you do

```yml
type: array
items:
  type: object
```
since `.Validate` is called on each of the items in the array.


generator/templates/server/parameter.gotmpl
```diff
@@ -173,7 +173,7 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) BindRequest(r *http.Requ
         break
       }
     }
-    {{ else }}if err := body.Validate(route.Formats); err != nil {
+    {{ else if or .IsComplexObject .IsTuple }}if err := body.Validate(route.Formats); err != nil {
```
casualjim: I think if you make this be: `not (or .IsInterface .IsPrimitive .IsCustomFormatter)`  then you don't need to edit the operation.go file

cozmo: As far as I can tell `isInterface` is true with

```yaml
type: object
properties:
  name:
    type: string
```

which is what lead me to using these properties instead.

casualjim:
I've got a fix in this PR for this issue, so I'm closing this PR

PR #221, opened 2016-01-20
#101 Custom templates

### LINKED ISSUE #101: Allow for client templates to be overridden with local versions
Allow for customization of the generated code by allowing people to override templates with local files.

pieter-lazzaro:
Fixes #101 

This is a very basic implementation of template overriding. It will scan the directory provided and if there is a file matching an existing template it will replace it with that.


casualjim:
Excellent contribution :+1: 

Do you think it would be possible to defer loading the template until it is used for the first time?
I think it would be nice if you ended up with a workflow where you just have to provide the templates or partials you wanted to override and it would use the embedded ones for the things that don't exist on disk.

If you think that's too much work for now that's fine too.

casualjim:
LGTM

pieter-lazzaro:
I had a look at making it a bit more customizable in terms of adding templates that don't exist etc. It would be nice to be able to not just overwrite the builtin templates but also extend without having to use the exact directory structure. If I get some time on the weekend I will have a look again and see if I can get something a bit more robust working.

casualjim:
Thanks, I figured I'd merge this already so that at least there is a way to do it. And then it can be improved with another PR

Issue #235, opened 2016-01-26
Make generated client use `consumes` in schema for Accept headers, rather than transport consumers


chancez:
Currently Accept headers are gathered from the registered consumers on a client's `httpkit/client/(*Runtime).Consumers` map. Instead the MIME types in the operation's `consumes` field should be used.

This looks like it would be a matter of updating the template https://github.com/go-swagger/go-swagger/blob/master/generator/templates/client/client.gotmpl, the operation struct https://github.com/go-swagger/go-swagger/blob/master/client/client.go#L18, and the client submit method https://github.com/go-swagger/go-swagger/blob/master/httpkit/client/runtime.go#L117-L122

casualjim:
i think this can be closed

PR #236, opened 2016-01-27
Use schemas consumes to determine accept headers

chancez:
Addresses #235

### DIFF
--- a/generator/operation.go
+++ b/generator/operation.go
@@ -55,6 +55,10 @@ func GenerateServerOperation(operationNames, tags []string, includeHandler, incl
 		if defaultScheme == "" {
 			defaultScheme = "http"
 		}
+		defaultProduces := opts.DefaultProduces
+		if defaultProduces == "" {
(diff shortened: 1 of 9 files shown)


generator/operation.go
```diff
@@ -398,9 +409,17 @@ func (b *codeGenOpBuilder) MakeOperation() (GenOperation, error) {
 		SuccessResponse: successResponse,
 		ExtraSchemas:    extra,
 		Schemes:         schemeOrDefault(schemes, b.DefaultScheme),
+		Produces:        producesOrDefault(produces, b.DefaultProduces),
 	}, nil
 }
 
+func producesOrDefault(produces []string, defaultConsumes string) []string {
```
casualjim: super tiny nitpick: I think `defaultConsumes` here wants to be called anything that doesn't have consumes in the name

casualjim:
same comment wrt to naming, from #238

https://github.com/go-swagger/go-swagger/pull/238#issuecomment-175397407

PR #237, opened 2016-01-27
Better custom template solution

pieter-lazzaro:
I have reworked the way templates are compiled so when specifying custom templates you are no longer restricted to just the template names that are embedded. I have added a doc page on it as well.

I will clean up the documentation a bit, hopefully when I get some more time, adding some documentation on the data type the templates are called with but I wanted to just put this up there.


pieter-lazzaro:
whoops, forgot that I was using protected templates in the tests

casualjim:
thanks :100:

PR #238, opened 2016-01-27
httpkit/client: Only set the content-type if the body isnt empty

chancez:
Based on top of my PR #236, can separate out if necessary

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -16,13 +16,14 @@ package client
 
 // Operation represents the context for a swagger operation to be submitted to the transport
 type Operation struct {
-	ID          string
-	Method      string
-	PathPattern string
-	Schemes     []string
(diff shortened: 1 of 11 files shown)


casualjim:
I think this has a merge conflict because I can't merge it, github won't let me.

After properly thinking this one through, I think the public interface for the client (contrary that what I said in the slack channel) should use consumers for what is the consumer on the client side, similarly to the producer. 

However when we look at the spec to get the defined consumes and produces values the produces values apply to the consumers property of the client and the consumes values apply to the producers property of the client. 

This removes surprising behavior on the client side and still captures intent from a client perspective, but the spec is defined from the server perspective so there they are inverted in naming. Also we should put this in a comment in the file so that people don't read that code and go wth?

PR #267, opened 2016-02-09
Fixes #263.

### LINKED ISSUE #263: server consuming `application/x-www-form-urlencoded` doesn't work
With the following spec: 
```json
{
  "consumes": [
    "application/x-www-form-urlencoded"
  ],
  "info": {
    "name": "formData test"
  },
  "paths": {
    "/testing": {
      "post": {
        "parameters": [
          {
            "in": "formData",
            "name": "names",
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "Success message"
          }
        }
      }
    }
  },
  "produces": [
    "application/json",
    "application/json; charset=utf-8"
  ],
  "swagger": "2.0"
}
```

generating a server from this one and than running it. Currently (latest tip e4c8b3c09d4ed2b35e6b77e57ab77169e56243fc) you will get: 

```shell
$ curl -X POST -H "Content-Type: application/x-www-form-urlencoded" "http://127.0.0.1:8282/testing" -d 'names=test'
{"code":500,"message":"request Content-Type isn't multipart/form-data"}
$ # obviously this next one doesn't work as well
$ curl -X POST -H "Content-Type: multipart/form-data" "http://127.0.0.1:8282/testing" -d 'name=test'
{"code":415,"message":"unsupported media type \"multipart/form-data\", only [application/x-www-form-urlencoded] are allowed"}
```

This appears to be coming from the std lib and more accurately from the call at
`restapi/operations/post_testing_parameters.go:37`
```go
29	func (o *PostTestingParams) BindRequest(r *http.Request, route *middleware.MatchedRoute) error {
30		var res []error
31		if err := r.ParseMultipartForm(32 << 20); err != nil {
32			return err
33		}
```

pieter-lazzaro:
request.ParseMultipartForm will return an error if the Content-Type is wrong. If the Content-Type is not a multipart this will just call ParseForm instead.

### DIFF
--- a/generator/templates/server/parameter.gotmpl
+++ b/generator/templates/server/parameter.gotmpl
@@ -121,7 +121,11 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) BindRequest(r *http.Requ
   var res []error
   {{ if .HasQueryParams }}qs := httpkit.Values(r.URL.Query())
   {{ else if .HasFormParams }}if err := r.ParseMultipartForm(32 << 20); err != nil {
-		return err
+		if err != http.ErrNotMultiPart {
+            return err
+        } else if err := r.ParseForm(); err != nil {
+            return err
+        }
 	}
 	fds := httpkit.Values(r.Form)
   {{ end }}
--- a/generator/templates/server/parameter.gotmpl
+++ b/generator/templates/server/parameter.gotmpl
@@ -120,11 +120,17 @@ type {{ pascalize .Name }}Params struct {
 func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) BindRequest(r *http.Request, route *middleware.MatchedRoute) error {
   var res []error
   {{ if .HasQueryParams }}qs := httpkit.Values(r.URL.Query())
-  {{ else if .HasFormParams }}if err := r.ParseMultipartForm(32 << 20); err != nil {
-		return err
-	}
-	fds := httpkit.Values(r.Form)
-  {{ end }}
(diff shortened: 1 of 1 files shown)


generator/templates/server/parameter.gotmpl
```diff
@@ -120,11 +120,17 @@ type {{ pascalize .Name }}Params struct {
 func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) BindRequest(r *http.Request, route *middleware.MatchedRoute) error {
   var res []error
   {{ if .HasQueryParams }}qs := httpkit.Values(r.URL.Query())
-  {{ else if .HasFormParams }}if err := r.ParseMultipartForm(32 << 20); err != nil {
-		return err
-	}
-	fds := httpkit.Values(r.Form)
-  {{ end }}
+  {{ else if .HasFormParams }}if r.Header.Get("Content-Type") == "multipart/form-data" {
```
casualjim: should this be a call to a helper function in httpkit that parses the header into a media type format?
I think at the very least it wants to be a prefix check.

pieter-lazzaro: Yep, you are right. I will change to use ContentType from httpkit. and also a prefix check. That reminds me is there a current way to test these kinds of issues?

pieter-lazzaro: Actually, I just checked the ParseMultipart function. It looks like it will return ErrNotMultipart. Maybe just check if the returned err matches that?

casualjim:
lgtm

PR #272, opened 2016-02-10
Bugfix for POST Ops not setting Content Length

akutz:
This patch sets the content length of the request or else a chunked transfer is declared, and this corrupts outgoing payloads. the content's length must be set prior to the body being written per the spec at https://golang.org/pkg/net/http

     If Body is present, Content-Length is <= 0 and
     TransferEncoding hasn't been set to "identity", Write
     adds "Transfer-Encoding: chunked" to the header. Body
     is closed after it is sent.

To that end a temporary buffer, b, is created to produce the payload body, and then its size is used to set the request's content length.


casualjim:
lgtm

PR #274, opened 2016-02-11
Change generated handlers' method and file names

sflxn:
The old scheme of handler method naming caused problems during regeneration.  For our project, we were generating a REST server and committing only the server code to our repo and discarding the models and operations generated by go-swagger.  Our intent was to regenerate those for every build, but the handler method names were not the same on code regeneration, even when we did not update the swagger spec.

By changing the method and file names to be a combo of Swagger Method + Path, it worked much better for us.


generator/operation_test.go
```diff
@@ -27,35 +27,11 @@ import (
...
-
-func TestEmptyOperationNames(t *testing.T) {
-	doc, err := spec.Load("../fixtures/codegen/todolist.simple.yml")
-	if assert.NoError(t, err) {
-		sp := doc.Spec()
-		sp.Paths.Paths["/tasks"].Post.ID = ""
-		sp.Paths.Paths["/tasks"].Post.AddExtension("origName", "createTask")
-		sp.Paths.Paths["/tasks/{id}"].Put.ID = ""
-		sp.Paths.Paths["/tasks/{id}"].Put.AddExtension("origName", "updateTask")
-
```
casualjim: where did those tests go? I can't seem to find tests that continue to test for this? Are they no longer required?

sflxn: TestUniqueOperationNames still runs, but since the name no longer relies on Post.ID, I removed them.  The exist assertions checks for uniqueness in the names for the path /tasks with different methods.

TestEmptyOperationNames didn't seem valid anymore since the new naming convention doesn't rely on Post.ID (which can be empty) anymore.  Paths are required, and method is also (...hopefully).

casualjim: oic so operation.ID is completely ignored now? 
Can we add it back so that people can provide pretty names if they really want to do so?  (I want to use that for the CLI etc)

sflxn: I do like the idea of pretty names, and there is a solution to keeping them.  However, that solution is too big to implement now.  We can certainly provide a CLI flag to exercise the former naming convention, but that will bring back the code regeneration breakage.  In effect, we would tell users that you can use this flag to use your IDs for pretty names, but then you cannot have code regeneration.

If the handler names were protected with a namespace (e.g. an interface), then the regeneration breakage gets solved and people can keep their pretty handler names.  It would really clean up configureAPI() and the generated server code.  I really want this, but it's too big for me to tackle now.

casualjim:
It breaks codegen for quay.io
I'll look into fixing this before monday

generator/operation_test.go
sflxn: To document what this code regeneration breakage looks like in the former naming convention, here is an example.  This is some handwaving, but you will get the idea.

1. generate a server with the docker spec in the fixture folder
2. look in configureAPI().  notice handlers with names like create1Container and create5Image.  (that's the handwaving.  recalling from memory.)
3. delete the models/ and restapi/ folders
4. regenerate.  This shouldn't touch the server code from gen 1 but will generate new models and operations
5. Try to rebuild.  It will fail because the new handler names have different unique indices.  For example create10Container and create20Image.  This is a mismatch from the names used in configureAPI() from gen 1.

casualjim:
superseded by #293

PR #285, opened 2016-02-14
adds swagger:file annotation, fixes #253

### LINKED ISSUE #253: Panic: interface is spec.Schema, not spec.Parameter for formField file
First up, thanks for the great tool :)

I'm trying to figure out how to use annotations for generating my client SDK for the case when the API call expects a form file. I've tried a few things, and this particular case gives me a panic:

```
panic: interface conversion: interface is spec.Schema, not spec.Parameter
```

Note: I'm not sure if the panic is just expected behaviour, but it goes unchecked in the code so I expect not. I guess I'd like it to tell me what I'm doing wrong.

Here's the full main.go to replicate this:

```go
//go:generate swagger generate spec -o ./swagger.json

// Package main MyCompany API.
//
//     Schemes: https
//     Host: localhost
//     BasePath: /
//     Version: 0.0.1
//     License: MIT http://opensource.org/licenses/MIT
//
//     Consumes:
//     - application/json
//
//     Produces:
//     - application/json
//
//
// swagger:meta
package main

import (
	"bytes"
	"net/http"
)

// swagger:parameters myOperation
type MyStruct struct {
	// MyFormFile desc.
	//
	// name: file
	// in: formData
	// type: file
	MyFormFile *bytes.Buffer `json:"myFormFile"`
}

func main() {
	// swagger:route POST /api/my-op/ myOperation
	//
	//     Consumes:
	//     - multipart/form-data
	//
	//     Produces:
	//     - application/json
	//
	//     Security:
	//       api_key:
	http.HandleFunc("/api/my-op/", func(w http.ResponseWriter, r *http.Request) {

	})
}
```

And obviously just run this

```
go generate
swagger generate client
```

Instead of the `*bytes.Buffer` I've tried `*os.File` and I get this:

```
unable to resolve embedded struct for: &{280192 file}
```

and `httpkit.File` and I get this:

```
unknown primitive "File"
```

Edit: I've also tried `multipart.File` and get

```
unknown primitive "File"
```

If it helps (and if you have any thoughts on if this is possible with go-swagger!), essentially I'm trying to use annotations to replicate my call that allows *both* a JSON-encoded `body` and also a `file` form fields, equivalent of this CURL request:

```curl
curl http://localhost/api/my-op/ \
  -F file=@myfile.txt \
  -F body='{"someKey":"someVal"}'
```

casualjim:


casualjim:
lgtm

casualjim:
lgtm

PR #292, opened 2016-02-15
Preserve trailing slash in URL path in runtime

### LINKED ISSUE #289: httpkit/client/runtime strips trailing slash from request path causing 301
At https://github.com/go-swagger/go-swagger/blob/master/httpkit/client/runtime.go#L138

It uses `path.Join` which strips the trailing slash from my path, e.g. https://play.golang.org/p/qIVI8ANGZt

My server 301s, but the redirect is not followed. Apparently 10 redirects are followed per https://github.com/go-swagger/go-swagger/blob/master/httpkit/client/runtime.go#L153 so not sure if this is a bug or not, but ideally I feel that the redirects should not be necessary in this case because the runtime should not change the given path - at least so far as not changing the trailing slash choice that the server makes.

jonathaningram:
Given a path like "/api/tasks/" the runtime uses `path.Join` to join the
base path to it which strips any trailing slash. If the path previously
ended with a trailing slash ("/"), it is reinstated after adding the
base path prefix because the behaviour of the server's path handling
cannot be guaranteed.

Fixes #289

Signed-off-by: Jonathan Ingram <jonathan.b.ingram@gmail.com>

### DIFF
--- a/httpkit/client/runtime.go
+++ b/httpkit/client/runtime.go
@@ -135,7 +135,14 @@ func (r *Runtime) Submit(operation *client.Operation) (interface{}, error) {
 	}
 	req.URL.Scheme = r.pickScheme(operation.Schemes)
 	req.URL.Host = r.Host
+	var reinstateSlash bool
+	if req.URL.Path != "" && req.URL.Path[len(req.URL.Path)-1] == '/' {
+		reinstateSlash = true
+	}
 	req.URL.Path = path.Join(r.BasePath, req.URL.Path)
+	if reinstateSlash {
+		req.URL.Path = req.URL.Path + "/"
+	}
 
 	r.clientOnce.Do(func() {
 		r.client = &http.Client{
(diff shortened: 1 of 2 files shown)


httpkit/client/runtime.go
```diff
@@ -135,7 +135,14 @@ func (r *Runtime) Submit(operation *client.Operation) (interface{}, error) {
 	}
 	req.URL.Scheme = r.pickScheme(operation.Schemes)
 	req.URL.Host = r.Host
+	var reinstateSlash bool
+	if req.URL.Path[len(req.URL.Path)-1] == '/' {
```
casualjim: I think you want to check if the string is not empty before doing this. I think it panics when the url path is empty

jonathaningram: Yep just seeing that too.

casualjim:
lgtm

casualjim:
Thanks a lot for this btw

jonathaningram:
You're welcome @casualjim it's the least I can do! You're the one that created the great lib :) Still hitting some curve balls to get it working for our API, but hoping we will get there.

PR #297, opened 2016-02-16
fix file type parameter when the parameter is required

adasescu:
The generated code does not compile if a file parameter has the property "required": true.
For example:
`"parameters": [
    {
        "name": "file",
        "in": "formData",
        "description": "file to upload",
        "required": true,
        "type": "file"
    }
],
"consumes": [
    "multipart/form-data"
],`


casualjim:
lgtm

PR #309, opened 2016-02-20
Add support for cookies in the client runtime

### LINKED ISSUE #308: Support for cookies in the client runtime
It would be great if support for setting the `http.CookieJar` could be added to `httpkit/client/runtime.go`, similar to the existing support for setting the `http.RoundTripper`. The default jar could be `nil` as before (no support for cookies), this is ok as far as it could be set to something else. 

We need this for one of our projects at SAP, since our existing services use session cookies. It's a very small and straightforward change, I would like to contribute a PR for it.

stoyanr:
Adds support for setting the `http.CookieJar` to `httpkit/client/runtime.go`, similar to the existing support for setting the `http.RoundTripper`. The default jar is nil as before (no support for cookies). Fixes #308.

### DIFF
--- a/httpkit/client/runtime.go
+++ b/httpkit/client/runtime.go
@@ -38,6 +38,7 @@ type Runtime struct {
 	Producers             map[string]httpkit.Producer
 
 	Transport http.RoundTripper
+	Jar       http.CookieJar
 	//Spec      *spec.Document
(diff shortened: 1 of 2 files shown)


casualjim:
lgtm

PR #314, opened 2016-02-22
fixes #222 server optionally embeds spec

### LINKED ISSUE #222: support not embedding the schema into the server generated code
The same as #190 but for the server generated code.

casualjim:


casualjim:
lgtm

PR #317, opened 2016-02-23
fix #306 also recognize application/octet-stream in generator

### LINKED ISSUE #306: Generated server is misinterpreting request type as application/octet-stream
When I do the following against a server that only supports application/json and text/plain at the moment, I get odd behavior.

Request:

curl -X POST -i http://127.0.0.1:49000/v1.21/images/create?fromImage=golang\&tag=latest

What the server sees:

POST /v1.21/images/create?fromImage=golang&tag=latest HTTP/1.1
Host: 127.0.0.1:49000
User-Agent: curl/7.43.0
Accept: */*

What the server responds:

HTTP/1.1 415 Unsupported Media Type
Content-Type: application/json
Date: Sat, 20 Feb 2016 00:00:55 GMT
Content-Length: 124

{"code":415,"message":"unsupported media type \"application/octet-stream\", only [application/json text/plain] are allowed"}

The handler for the operation never gets called.  This interpretation of the request is happening much earlier.

casualjim:


casualjim:
lgtm

PR #328, opened 2016-02-26
Fix client package name

jakedsouza:


jakedsouza:
I'm a little concerned with the amount of changes that occurred after running ./hack/regen-samples.sh

casualjim:
most of them are due to go's maps being unordered, they aren't actual changes. And you're right I should run regen-samples more often

PR #339, opened 2016-02-29
fixes #330 package for client is client package

### LINKED ISSUE #330: API client has wrong package name
Run ```swagger generate client -A TodoList  -f swagger.yml``` on a sample swagger spec.

```yml
swagger: "2.0"
paths:
  /:
    get:
      parameters:
        - name: since
          in: query
          type: integer
          format: int64
      responses:
        default:
          description: generic item
          schema:
            $ref: "#/definitions/item"
definitions:
  item:
    type: object
    required:
      - description
    properties:
      id:
        type: integer
        format: int64
        readOnly: true
```

The generated client is under package ```operations``` while it should be under package ```client``` but is 

Also verified by passing the ```--client-package``` flag but does not seem to make a difference to the generated code.

casualjim:


casualjim:
lgtm

PR #347, opened 2016-03-02
generator: Use a JSON decoder with UseNumber enabled for unmarshaling

chancez:


chancez:
I've partially fixed the tests, just need to take some time to finish the rest up and push the changes. I've got a patch in the code I'm using to work around this particular issue, so it's not an immediately pressing thing. Regardless I'll try to wrap this up soon.

PR #362, opened 2016-03-03
Add XML consumer and producer

### LINKED ISSUE #359: No consumers and producers for "application/xml" in the client runtime
By default, there are currently no consumers and producers associated with "application/xml" in the client runtime. These should be rather trivial to add. In our project, we use the following very simple implementations:

```go
// XMLConsumer creates a new XML consumer
func XMLConsumer() httpkit.Consumer {
	return httpkit.ConsumerFunc(func(reader io.Reader, data interface{}) error {
		dec := xml.NewDecoder(reader)
		return dec.Decode(data)
	})
}

// XMLProducer creates a new XML producer
func XMLProducer() httpkit.Producer {
	return httpkit.ProducerFunc(func(writer io.Writer, data interface{}) error {
		enc := xml.NewEncoder(writer)
		return enc.Encode(data)
	})
}
```

stoyanr:
Add XML consumer and producer and use it in client runtime and generator support. Fixes #359.


casualjim:
lgtm

PR #364, opened 2016-03-04
Add plain text consumer and producer

### LINKED ISSUE #360: No consumers and producers for "text/plain" in the client runtime
By default, there are currently no consumers and producers associated with "text/plain" in the client runtime. These should be rather easy to add.

stoyanr:
Add plain text consumer and producer and use it in client runtime and generator support. Fixes #360.


casualjim:
lgtm

Issue #159, opened 2015-12-09
The validation does not work for the reference.


feikiss:
The validation feature does not support to validate the reference.  In the "$ref" element
![incorrect referenct](https://cloud.githubusercontent.com/assets/2208663/11685168/01917d6c-9eb2-11e5-9968-ba52d0d7f9be.png)

, I input an incorrect value, the validation can still pass. Actually, it should throw error and report. 

Although there is a ticket https://github.com/go-swagger/go-swagger/issues/16, but that one is just an enhancement, not a bug ticket.

I believe this is a validation bug and we should  fix it.

Thanks,
Xiaofei

elgris:
Also, I noticed that validator treats references as validation errors. For instance, let's put a reference to parameter list:
```
  /foo:
    get:
      summary: Some endpoint
      parameters:
        - name: some_id
          in: query
          description: some ID
          required: true
          type: integer
        - $ref: '#/parameters/some_shared_param'
```

Validator output for such case is:
```
- paths./foo.get.parameters.$ref in body is a forbidden property
- paths./foo.get.parameters.name in body is required
- paths./foo.get.parameters.in in body is required
- paths./foo.get.parameters.type in body is required
```

As specification says, using references in parameter list is valid: http://swagger.io/specification/#operationObject (`parameters` is expected to hold an array `[Parameter Object | Reference Object]`)

Is there any plans to fix? Perhaps I overlooked something :)

casualjim:
this is a new issue or at least it's different from the one that's reported. But it is a bug

elgris:
I thought that it looks similar because of the same reason: incorrect processing of `$ref` :) Let me open a new one...

PR #382, opened 2016-03-06
adds validation of reference path

elgris:
According to JSON reference spec http://tools.ietf.org/html/draft-pbryan-zyp-json-ref-02#section-4 reference must be represented by valid URI which is not only a "link" with schema, but also a path to local file.

This PR supports #159

### DIFF
--- a/fixtures/bugs/159/swagger.yml
+++ b/fixtures/bugs/159/swagger.yml
@@ -1,0 +1,21 @@
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: 'Test'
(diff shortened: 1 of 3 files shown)


casualjim:
thanks for the contribution but this breaks the build. There is a panic with a stackoverflow

casualjim:
superseded by: #387

elgris:
@casualjim thanks! I'm sorry, didn't notice the problem on my local, because it didn't reproduce at the first time. Following test runs revealed the problem.

PR #391, opened 2016-03-07
Fix issue with unknown primitive.

gaplyk:


casualjim:
Could you add a test that proves it works?

casualjim:
:+1:

PR #392, opened 2016-03-07
split restapi.NewServer

caglar10ur:
NewServer no longer configures the API but instead relies on
ConfigureAPI call to do it.

Also add a new ConfigureFlags method so that we can add command line
flags from the handlers.

This way we can get additional flags from the handlers and initialize
them once

Signed-off-by: S.Çağlar Onur <conur@vmware.com>


casualjim:
this needs an update

casualjim:
lgtm

PR #396, opened 2016-03-08
Proper handle of array of pointers to struct

gaplyk:
Handling of the following field:

```go
type MyStruct struct {
  Field string
}

type MyStructs []*MyStruct

type SomethingHere struct {
   MyStr MyStructs
}
```


casualjim:
:+1: thanks!

PR #401, opened 2016-03-09
generate all models if name not specified

easeway:
In command `generate model ...`, It's not necessary to force `--name` as required. If we don't specify a name, it should generate all models.


casualjim:
lgtm

PR #405, opened 2016-03-11
Add setter to generated response writer

anfernee:
While the With methods are more readable and less boring to write,
the return type of With methods make it impossible to abstract common
interface out. Then it's hard to have a factory method to generate these
objects. Though most of the time, we are generating similar response,
we cannot write generic code because there's no common interface.

For example, Every response type has WithPayload method like this:

`
func (r *ResponseType) WithPayload(payload *models.Error) *ResponseType 
`

This patch adds a setter method for each field, then the method becomes:

` 
func (r *ResponseType) SetPayload(payload *models.Error)
`

For all the response type that has models.Error payload, we can have
common interface. Then it's possible to have a factory method that
creates such response type with specific error payload.


casualjim:
this needs an update before it can get merged

casualjim:
lgtm

PR #408, opened 2016-03-11
Add *raw-stream mimetype mapping

sflxn:
Added a mapping for application/.*raw-stream to bin mapping for
mimetypes.  We have a specific vendor mimetype that ends in
raw-stream that we would like to get handle.  If this isn't there,
we can get a panic during runtime in some code generation.  If
the generated file restapi/operations/xxx_api.go refers to the
mimetype as the defaultConsumes or defaultProduces, our server
will panic.  Adding this resolves the panic.


casualjim:
lgtm

PR #411, opened 2016-03-13
Remove dependency heavy goconvey

casualjim:
I removed a bunch of dependencies from the vendor directory and removed the use of goconvey
it pulled in a bunch of dependencies and the tests were very wordy

This makes go-swagger hopefully feel a lot lighter


casualjim:
lgtm

PR #416, opened 2016-03-15
fix Panic expanding responses that are a schema ref with circular

### LINKED ISSUE #415: Panic expanding responses that are a schema ref with circular refs
## Swagger specification

https://raw.githubusercontent.com/APIs-guru/api-models/master/APIs/clickmeter.com/v2/swagger.yaml

## Steps to reproduce

Load above document and try to expand it.

pytlesk4:
closes #415


casualjim:
:shipit:

PR #417, opened 2016-03-15
resolves refs for parameters and responses, fixes #384

### LINKED ISSUE #384: No data fields in the response object
I have a definition of a response data structures defined as this:

And responses section defined as this:
```
responses:
  default_error:
    description: Error response
    schema:
      type: object
      required: [request_id]
      properties:
        request_id:
          type: string
```

```
paths:
  /something:
    get:
      tags:
        - something
      parameters:
        - name: some
          in: query
          type: string
          format: string
      responses:
        default:
          $ref: "#/responses/default_error"

```

The problem is in the generated default response code that has only one private field: _statusCode
There is no struct member such as request_id.

```
/*GetSomethingDefault get new object default

swagger:response getSomethingDefault
*/
type GetSomethingDefault struct {
	_statusCode int
}
```

casualjim:


casualjim:
lgtm

PR #425, opened 2016-03-17
Add a serve-ui command 

jakedsouza:
This PR adds a serve-ui command to swagger

The command is pretty simplistic. 
1. Given a swagger spec path, it loads the spec 
2. Host the spec.json on a server  
3. Open petstore.swagger.io with the locally hosted spec. 

This would allow users to get a quick view of their specification

---
This change is [Reviewable](https://reviewable.io/reviews/go-swagger/go-swagger/425)

### DIFF
--- a/cmd/swagger/commands/serve_ui.go
+++ b/cmd/swagger/commands/serve_ui.go
@@ -1,0 +1,57 @@
+package commands
+
+import (
+	"errors"
+	"log"
+	"net/http"
+	"net/url"
+
+	"github.com/go-swagger/go-swagger/spec"
+	"github.com/rs/cors"
+	"github.com/toqueteos/webbrowser"
+)
+
+//ServeUI Show given swagger spec in swagger-ui
+type ServeUI struct {
+}
+
+// Execute this command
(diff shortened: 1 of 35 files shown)


cmd/swagger/commands/serve_ui.go
```diff
@@ -0,0 +1,57 @@
...
+	// serve the swagger.json
+	mux := http.NewServeMux()
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Write([]byte(doc.Raw()))
+	})
+
+	handler := cors.Default().Handler(mux)
+
+	http.ListenAndServe(":8080", handler)
```
casualjim: can you make this a random port so that there aren't any collisions?
Also print out the full url to petstore.swagger.io?url=specurl 
That allows for people to cmd+click or ctrl click the url to go there even fit he browser didn't open it. Might also be good to make the auto browser opening a flag as well as a flag for the base url for swagger ui

casualjim:
functionality included by #683

PR #426, opened 2016-03-19
make http.Request available to hanlders via params

saurabhsurana:
This gives handlers access to the http.Request object itself via Params.

I liked this approach better than just supplying the RequestURL, because now handlers can build the URLs themselves or use the http.Request as per their need.


casualjim:
lgtm

casualjim:
I've cherry-picked this commit onto another branch, so closing this PR

PR #427, opened 2016-03-19
Improve error message

abclogin:
Error messages improved. By example:

Old message is:
&{%!s(token.Pos=1478849) %!s(*ast.FieldList=&{1478853 [0xc8217db440] 1478870}) %!s(*ast.FieldList=&{1478872 [0xc8217db480 0xc8217db4c0] 1478888})} is unsupported for a schema

New message is:
Expr (/path/src/gopkg.in/mgo.v2/session.go:386:13) is unsupported for a schema


casualjim:
:shipit:

PR #448, opened 2016-03-30
Fixed bug in httpkit/client/runtime.go Submit()

ritchida:
The functionality to determine the value of reinstateSlash would
set it to true in the case where req.URL.Path is "/".  This would
result in an extra "/" being appended to req.URL.Path after it was
joined with the BasePath, causing requests to fail.  The path.Join()
function calls path.Clean() on each path element to be joined.
Clean() will return a trailing "/" on a path element only if it is
"/".


casualjim:
lgtm

Issue #449, opened 2016-03-31
Swagger generate oauth2 scope


galeone:
Hi, I'm trying to use swagger to generate a swagger.json file.

I've defined a route following [the referece](http://goswagger.io/generate/spec/route/), in that way:

```go
// Posts handles the request and returns the required posts written by the specified user
func Posts() echo.HandlerFunc {

	// swagger:route GET /users/{id}/posts user posts getUserPosts
	//
	// List posts on user board, filtered by some parameters.
	//
	// This will show the last posts on the user board by default.
	// You can personalize the request via query string parameters
	//
	//	Security:
	//		oauth: profile:read,base:read
	//
	//	Responses:
	//		default: apiResponse

//..... other code
}
```

Afther that, I generate the file with `generate swagger generate spec -o swagger.json`

But the ":" character in the oauth2 scope definition is missing as you can see:
```json
    "paths": {
        "/users/{id}/posts": {
            "get": {
                "description": "This will show the last posts on the user board by default.\nYou can personalize the request via query string parameters",
                "tags": ["user", "posts"],
                "summary": "List posts on user board, filtered by some parameters.",
                "operationId": "getUserPosts",
                "security": [{
                    "oauth": ["profileread", "baseread"]
                }],
                "responses": {
                    "default": {
                        "$ref": "#/responses/apiResponse"
                    }
                }
            }
        }
```
In my API, scopes follows the following syntax: `<resource>[:[read,][write]]`
Thus something like `profile:read,write base:read` or `comments:read profile:write` is fine in my specification (and in OAuth2 specification too, because the list of scopes con contain anything and scopes must be separated by a single space). 

Is there something I'm doing wrong or it's a swagger bug / limitation?

Thank you!

casualjim:
this is a bug in the scanner, it should consider other chars.

https://github.com/go-swagger/go-swagger/blob/master/scan/validators.go#L488

that should probably use a different regex to capture those values.

PR #451, opened 2016-03-31
Fix https://github.com/go-swagger/go-swagger/issues/449

galeone:
This PR aims to fix:  https://github.com/go-swagger/go-swagger/issues/449

I changed the `/fixtures/goparsing/classification/operations/todo_operation.go` file in order to test more general scopes (For example Google uses URL as scopes).

I changed the  `scan/routes_test.go` accordingly.

I removed the useless regexp that I replaced with a simple TrimAfter followed by a TrimSpace. This is possibile because following the specifications scopes can't contain spaces (must be url encoded if present).

Signed-off-by: Paolo Galeone <nessuno@nerdz.eu>


casualjim:
lgtm thanks!

PR #459, opened 2016-04-03
X nullable first

vburenin:
This change introduces a precedence for a user defined attribute x-nullable.
x-nullable is an extension that actually gives user a freedom to optionally define code generator behavior for nullable and non nullable values.


casualjim:
lgtm

PR #464, opened 2016-04-05
Use spec defined basePath when available.

nikhilm:


nikhilm:
Apologies for not adding tests in the original PR. Will update this soon.

generator/support.go
```diff
@@ -671,7 +671,7 @@ func (a *appGenerator) makeCodegenApp() (GenApp, error) {
 
 	basePath := "/"
 	if sw.BasePath != "" {
-		basePath = "/"
+		basePath = basePath
```
casualjim: this should be sw.BasePath no?

nikhilm: Sorry, between github's outage and something I was working on, I'm clearly not thinking straight.

casualjim: no worries that's what PR's are for :)

PR #466, opened 2016-04-05
Use spec BasePath when it is non-empty.

nikhilm:
This fixes the issue, but I'd like to know a good place to put a test for this (or modify an existing test).
In addition, I'm opening the PR so I know the existing test-suite passes because
1) I couldn't find instructions on how to run all tests locally
2) The Circle process I ran on my fork failed - https://circleci.com/gh/nikhilm/go-swagger/3

### DIFF
--- a/fixtures/bugs/465/swagger.yml
+++ b/fixtures/bugs/465/swagger.yml
@@ -1,0 +1,22 @@
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: 'Test'
+schemes:
+  - http
+basePath: /v1/fancyAPI
+produces:
+  - application/json
+consumes:
+  - application/json
+paths:
+  /key/{id}:
+    delete:
+      parameters:
+        - name: id
+          in: path
+          type: integer
+          required: true
+      responses:
+        '200':
+          description: OK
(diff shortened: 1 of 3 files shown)


casualjim:
I run tests locally like this:

```
go test $(go list ./... | grep -v vendor)
```

Another way to do it is:

```
docker build -t go-swagger:local -f Dockerfile.ci .
docker run --rm -it go-swagger:local ./hack/build-docker.sh
```

And the last way to do it is:

```
drone exec
```

nikhilm:
existing test suite passed locally, where should I put the new test and
what can I use as an example for this kind of test? Thanks!

On Tue, Apr 5, 2016 at 4:46 PM, Ivan Porto Carrero <notifications@github.com

casualjim:
I think writing a test for this would only matter for the client generation stuff. The BasePath property is used in the client facade


it looks a bit like these tests: https://github.com/go-swagger/go-swagger/blob/master/generator/operation_test.go#L301-L375

the template for that is: `clientFacadeTemplate`
and you're checking if this bit renders correctly: `transport := httptransport.New({{ printf "%#v" .Host }}, {{ printf "%#v" .BasePath }}, {{ printf "%#v" .Schemes }})`

nikhilm:
The CircleCI failures don't seem to be my fault.

casualjim:
lgtm 

I go by the drone one, I should remove the circle one I won't be using it anymore going forward.

casualjim:
Thanks a lot!

PR #467, opened 2016-04-06
generator: fix package assumption on server generation

jwmaag:
Just a little thing I noticed when i had the -a flag set to something other than 'operations'


casualjim:
lgtm

PR #469, opened 2016-04-07
Added support for complex objects (array of objects).

zxchris:
When expanding a schema, go-swagger does not correctly process a response which is an array of objects. For example:

```json
 "Pets": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/Pet"
      }
}
```

This pull request adds detection and correct expansion of this case.


casualjim:
this was fixed differently, but the functionality exists now

PR #473, opened 2016-04-08
Required flag is reset for array items.

vburenin:
For some reason the schema for array item is enforced in such way that array items become required.
I created a copy of the originating schema and reset required flag for it. Test update was necessary.


casualjim:
lgtm

Issue #420, opened 2016-03-15
Refactor package structures


casualjim:
There are a number of packages that are more like libraries, they would be better extracted.

Perhaps it's a good thing to extract these into their own repos. The CLI command can then vendor each of those repos. Most of them are stable at this point.

specs, jsonreference, jsonpointer, scan, swag and the httpkit

pytlesk4:
Would love to see this happen, and help out where I can.  We use go-swagger currently at [http://stoplight.io/](url), and when I was evaluating it, I almost passed it up because we were looking for a way to just parse and load a swagger file.

If everything were to be broken up into libraries, I feel like it would be a lot more approachable.  Go-Swagger can do a lot of cool things, and navigating the source code was daunting.  One day Stoplight might do code generation, for apis, but because we don't need that now I almost passed on this awesome project.  I have submitted a few pull requests to fix some issues in the spec package, and even though all the tests were passing in the spec package, my changes broke some tests in another packaged that I knew nothing about.

That is all for now.

phedoreanu:
It would be nice to serve swagger-ui in a more user-friendly way. 

Also, for better compatibility with the Docker `go-wrapper` command, can `main.go` be moved in the root `$projectFolder` i.e. `test/test-api/main.go` from `test/test-api/cmd/test-server/main.go`?

Docker build errors: 
 * `can't load package: package test/test-api: no buildable Go source files in /go/src/test/test-api`
 * `can't load package: package test/test-api/cmd/test-server: no buildable Go source files in /go/src/test/test-api/cmd/test-server`

emicklei:
I am the author of the [swagger2](https://github.com/emicklei/swagger2) package which is intended to be a rewrite/upgrade of [go-restful/swagger](https://github.com/emicklei/go-restful/tree/master/swagger).

The goal of this swagger2 project is to separate the Swagger part from the REST part, allowing others to use the package independently by provinding only the specification objects. I was planning to create another (sub?) package for building model objects from Go types (as I already created 1.2).

As discussed offline with one of the authors of go-swagger, I would like to see go-swagger to be split into multiple repos for the most part because of handling the dependencies. Applications that import go-swagger (using vendoring) should not be forced to include the whole package deal (e.g. generators).

If the go-swagger project would separate the spec part into a new repo then I will probably drop my swagger2 project and collaborate on that instead.

casualjim:
[...]

PR #485, opened 2016-04-16
move analysis out of spec package

casualjim:
refactoring as part of #420 

This step extracts analysis from the spec document
Moves yaml into its own packages for opt-in behavior
Moves loading into its own package


analysis/analyzer.go
```diff
@@ -0,0 +1,621 @@
...
+	allRefs    map[string]spec.Ref
+	referenced struct {
+		schemas    map[string]SchemaRef
+		responses  map[string]*spec.Response
+		parameters map[string]*spec.Parameter
+	}
+}
+
+func (r *referenceAnalysis) addRef(key string, ref spec.Ref) {
+	if r.allRefs == nil {
```
vburenin: Is there any particular reason to initialize it like this?

analysis/analyzer.go
```diff
@@ -339,7 +353,7 @@ func (s *specAnalyzer) SecurityDefinitionsFor(operation *Operation) map[string]S
 }
 
 // ConsumesFor gets the mediatypes for the operation
-func (s *specAnalyzer) ConsumesFor(operation *Operation) []string {
+func (s *Spec) ConsumesFor(operation *spec.Operation) []string {
 	cons := make(map[string]struct{})
```
vburenin: This might make sense to specify a capacity since it is already known.

spec/spec.go
```diff
@@ -147,20 +147,20 @@ func New(data json.RawMessage, version string) (*Document, error) {
...
-		specAnalyzer: specAnalyzer{
-			spec:        spec,
-			consumes:    make(map[string]struct{}),
-			produces:    make(map[string]struct{}),
-			authSchemes: make(map[string]struct{}),
-			operations:  make(map[string]map[string]*Operation),
-			allSchemas:  make(map[string]SchemaRef),
-			allOfs:      make(map[string]SchemaRef),
-		},
+		// specAnalyzer: specAnalyzer{
```
vburenin: Do we need this commented code?

casualjim:
lgtm

PR #486, opened 2016-04-18
420 package refactor

### LINKED ISSUE #420: Refactor package structures

casualjim:
moves a bunch of packages around

all the library like packages have moved to the go-openapi org. 

closes #420


casualjim:
lgtm

Issue #494, opened 2016-04-22
Parameters template bug leads to errors in header validation


galaxie:
## Required headers params like `X-Request-ID` are never valid

It's cased by incorrect access to `request.Header` values by non-canonicalized key.

## Swagger specification
Imagine this specification
```
parameters:
  requestID:
    name: X-Request-ID
    in: header
    required: true
    type: string
```
## Steps to reproduce
Any requests with filled `X-Request-ID`  to a server generated according to the specification

Requests are never valid because of [this string](https://github.com/go-swagger/go-swagger/blob/master/generator/templates/server/parameter.gotmpl#L151)  . Keys are canonicalized in `request.Header`, so there are no `X-Request-ID` there, but `X-Request-Id`, so the incorrect map access makes `X-Request-ID in header is required` error.

PR #497, opened 2016-04-22
Add missed go code generation for #494

galaxie:
I didn't notice `.githooks` folder and that I was supposed to use `git-hooks` to install hooks manually (am I right?).  It would be great to mention this in the contribution guidelines. 

So I add missed `bindata` regeneration for #494


casualjim:
💯 lgtm

PR #498, opened 2016-04-22
Prevents parsing structs that are needed when parsing dicovered schemas

seanbrant:
When parsing discovered schemas it includes all structs including ones that should not be included. This patch makes the schemaParser aware of the discovered schema so we can use that information to exclude schema that are not needed.


casualjim:
lgtm thanks!

Issue #520, opened 2016-05-17
Skip parameter fields with json tag `json:"-"`


jlburkhead:
## Problem statement

From the generation docs:
```
Annotation syntax

If you want to exclude something from the spec generation process you can try with the struct tag: json:"-"
```

I'm not sure which generation parsing rules follow that, but `swagger:parameters` do not skip fields with that json tag. Would be nice to be able to hide fields from swagger without making them unexported. 

## Steps to reproduce

```go
// Params struct
//
// swagger:parameters someOperation
type Params struct {
        // The value
        Value string `json:"-"`
        // Some other value
        // in: query
        OtherValue string `json:"other_value"`
}
```

running `$ swagger generate spec ...` will still include Value in the generated `swagger.json`

jlburkhead:
closing since the pr for this was merged

Hectorhammett:
Is it possible to add to Swagger an json:"-" field? Basically the other way around of this post.

fredbi:
@Hectorhammett try extension x-go-custom-tag (applies to schema object properties)

Hectorhammett:
@fredbi  Thanks for the response! I will try it on the next days and will response the results. Thanks!

PR #521, opened 2016-05-17
skip fields in parameter structs with struct tag `json:"-"`

jlburkhead:
Signed-off-by: Jake Burkhead <jake.b@socialcodeinc.com>

relates to https://github.com/go-swagger/go-swagger/issues/520

### DIFF
--- a/fixtures/goparsing/classification/operations/noparams.go
+++ b/fixtures/goparsing/classification/operations/noparams.go
@@ -52,6 +52,7 @@ type ComplexerOneParams struct {
 	mods.NotSelected
 	mods.Notable
 	CreatedAt strfmt.DateTime `json:"createdAt"`
+	Secret    string          `json:"-"`
 
 	// in: formData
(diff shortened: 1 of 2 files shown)


casualjim:
lgtm

rookie7799:
Howdy, 
Is it possible to skip a field with some special syntax under `swagger:parameters` and not just with `json:"-"` ? Eg:

```
// swagger:parameters listBars
type BarParam struct {
// skip: true
Field string `json:"field"`
...
```

casualjim:
if you're sending it with the json doesn't it need to be included in your contract?

rookie7799:
Right, but I'm using the same struct elsewhere that is not part of swagger
and it needs that 'json' tag.

casualjim:
you could teach go-swagger to also recognize `swagger:` as struct tag perhaps as alternative? Should be fairly straightforward because it's the same as this PR

PR #526, opened 2016-05-21
Fixed bug for not generated validator call if there is an array of complex objects.

vburenin:


casualjim:
lgtm, that one build failure is a appveyor fart

Issue #545, opened 2016-05-27
Concurrency issue on the validation code for string enums


bfreis:
## Problem statement

The code generated for validating string enums might contain an issue with concurrent code, that can be fixed very easily.

There's no synchronization when initializing the global `[]interface{}` var that contains the possible values of an enum. So it is possible that one goroutine initializes it with some values, and then another different goroutine might see and use a partially initialized list of values, so the validation code could fail.

The easy fix is to use a sync.Once and move the entire initialization block inside an once.Do wrapper. This is detailed below.

## Swagger specification

```
definitions:
  buggy:
    type: object
    properties:
      broken:
        type: string
        enum:
        - MIGHT_WORK
        - MIGHT_FAIL

```

## Steps to reproduce
**Caveat**: it is extremely difficult to reproduce the issue (or, in other words, it is impossible to create a test case that will reliably fail). The reason is that, because of the lack of synchronization, the runtime behavior is unspecified. So it *might* work as expected sometimes, or it *might* fail as described below sometimes.

Generate models using the spec snippet above, with `swagger generate model -f swagger.yml`. The following code is created in `buggy.go` for the validation logic:

```
var buggyTypeBrokenPropEnum []interface{}

func (m *Buggy) validateBrokenEnum(path, location string, value string) error {
	if buggyTypeBrokenPropEnum == nil {
		var res []string
		if err := json.Unmarshal([]byte(`["MIGHT_WORK","MIGHT_FAIL"]`), &res); err != nil {
			return err
		}
		for _, v := range res {
			buggyTypeBrokenPropEnum = append(buggyTypeBrokenPropEnum, v)
		}
	}
	if err := validate.Enum(path, location, value, buggyTypeBrokenPropEnum); err != nil {
		return err
	}
	return nil
}

```

The problem can occur when we have multiple goroutines. Let's assume we have two goroutines: Goroutine 1 and Goroutine 2.

According to the memory model, if Goroutine 1 calls `validateBrokenEnum(...)` runs this code "first", and then "later" Goroutine 2 calls the same function, the following can happen:

- Goroutine 1 calls `validateBrokenEnum(...)`
- Goroutine 1 evaluates `buggyTypeBrokenPropEnum == nil` to true, so it enters the the first `if` block
- Goroutine 1 runs the first iteration of the `for` block, storing something non-nil into the `buggyTypeBrokenPropEnum` var.
[...]

PR #547, opened 2016-05-28
fixes go-swagger/go-swagger#545 Concurrency issue on the validation c…

bfreis:
…ode for string enums


generator/templates/schemavalidator.gotmpl
```diff
@@ -183,32 +183,32 @@ if err := {{.ValueExpression}}.Validate(formats); err != nil {
...
-    if err := json.Unmarshal([]byte(`{{ json .Enum }}`), &res); err != nil {
-      return err
-    }
-    for _, v := range res {
-      {{ camelize .Name }}Enum = append({{ camelize .Name }}Enum, v)
-    }
+func init() {
+  var res []{{ template "dereffedSchemaType" . }}
+  if err := json.Unmarshal([]byte(`{{ json .Enum }}`), &res); err != nil {
+    return err
```
vburenin: it is not possible to return an error here which causes compilation error.

bfreis: True! I probably didn't run my tests after regenerating the bindata.go file :(

Maybe `panic(err)` would be appropriate there? If we can't initialize the validator, it's because the validator code must be wrong, and not something that client code should be able to capture and deal with?

casualjim:
There is still an issue with it where it generates `models/hook_event.go:59: too many arguments to return`  

you can run ./hack/run-canary.sh to reproduce

Issue #476, opened 2016-04-08
lifting things to pointers


casualjim:
go-swagger is very pointer happy it still has an issue for patching objects because with pointers because we don't know if it was actually a value that was set or if it was a value that's nil.

So i think this will have to change one more time to use nullable types where set/unset is a property on the type. 

That keeps GC pressure lower, but I think that might create a need for 2 structs. 
One is the binding/validating struct and the second one is the one the user actually uses because a nullable type would kind of look like

```go
type NullableBool struct {
  value bool
  IsNull bool
  WasSet bool
}

func (n *NullableBool) Value() *bool  {
  if nIsNull {
    return nil
  } 
  return &n.value
}
```

The biggest benefit we derive from this is the ability to do patches on objects

vburenin:
This sounds spooky to me. If I need something like that, I would handle it by my self adding additional properties to the patch object.

sflxn:
I actually like the idea of adding accessor functions to hide the details.  You can only do so much with basic data types.  It's a pointer or it isn't.  Attributes like optional, required, default, etc are semantics that are very difficult to represent in computer languages.  The only way to do it is to make complex data structures to add those semantics.

The only issue I have is that go-swagger has allowed us to access these data types directly up till now, so when the implementation changes, we all have to go through and change our code to use the updated semantics.  Other "IDL" technology uses the approach of accessor functions.  I bring up Thrift as an example.  I cannot remember the exact code, but to find out if a data is optional, that information is only available through a function call.

If a model was something like,

Image {
   ID *string
   name *string
}

An example generation could be

func (i *Image) GetIDValue() (*string, optional, required) {}

I am not advocating that specific signature.  It's just an example.  When Swagger adds more richness to their definition, go-swagger can enrich the data types in the back without affecting users' code.

Yes, it's more tedious to use accessors, but it's the only way to add richness to datatypes in a language (go) that has limited semantics in the datatypes.

sflxn:
Another option for the implementation is to use go's fields and reflection.

casualjim:
I think perhaps a better solution is to add a companion field instead of making new types.
This companion field would describe the value like nullable and so forth and at the same time also contain if the value was set or unset from an unmarshal step.

This would enable the patch use case where you need that level of information, allow for richer knowledge about a type but at the same time not break existing code.

chancez:
I'm not sure how I feel about that.

[...]

PR #557, opened 2016-06-06
[WIP] 476 lift pointers

casualjim:
Starts implementing the items discussed in #476 

For this to work we need a custom serializer for all types, not just the "special" ones. So most of the work is around generating a serializer, to a lesser extent it is about pointer/no pointer decisions at render time.


Yarstions32:
Any updates?

ysomad:
any updates?

Issue #560, opened 2016-06-08
All Go builtin types should be supported


zasran:
## Problem statement

scan/scanner.go function swaggerSchemaForType maps primitive Go types to swagger type/format. It's missing some of Go builtin types.

Error generated when unsupported type is used:

`package reflect, error is: unknown primitive "uintptr"`

## Steps to reproduce

One of the missing types is uintptr which is used in e.g. Go reflect package, error would be triggered by following example.

Example code that triggers the error:

```go
import "reflect"
// User
//
// Use documentation here...
//
// swagger:model
type User struct {
	// reflect type for testing reflect
	//
	// required: false
	Rt reflect.Type `json:"rt"`
	// the id for this user
	//
	// required: true
	Id int64 `json:"id"`
	Name string `json:"name"`
}
```

Swagger invocation:

```bash
$ swagger generate spec -m
package reflect, error is: unknown primitive "uintptr"
```

Seems like a good solution would be to support all Go builtin types as listed at [https://golang.org/pkg/builtin/](https://golang.org/pkg/builtin/) (except for the *Type types which are there for documentation purposes only).

I am not sure what the swagger type/format should be for the missing Go builtin types, seems like some of them are not trivial to handle (e.g. byte, complex).

zasran:
After further investigation we figured out that based on Go builtin doc https://golang.org/pkg/builtin/ the following changes in swaggerSchemaForType might make sense:

- ADD byte: prop.Typed("integer", "uint8")
- ADD complex128, complex64 should cause error, no default JSON marshaller
- CHANGE error
  - OLD: prop.Typed("string", "")
  - NEW: error, it's an interface with no default JSON marshaller
- CHANGE rune
  - OLD: prop.Typed("string", "")
  - NEW: prop.Typed("integer", "int32"), doc says rune is alias for int32, it's encoded as a number
- ADD uintptr: prop.Typed("integer", "uint64")

Note about the error: some implementations of error interface might be JSON encoded as a string however it seems to be dangerous to assume that it's always the case, e.g. errors.New creates error that is encoded as JSON empty object {}.

If anybody wants to play with this https://play.golang.org/p/4l2AKRz4LO demostrates how Go builtin types are encoded to JSON.

Please let me know what you think about these suggestions.

casualjim:
I've started answering here several times. But I keep getting stuck. would it be possible for you to join the slack team?  
[...]

PR #562, opened 2016-06-16
Fix issue #560 Go builtin support

zasran:
* added support for byte, uintptr

* added error for complex64, complex128, there's no default JSON
marshaller

* changed rune from string to int32 according to Go docs

### DIFF
--- a/fixtures/goparsing/classification/models/nomodel.go
+++ b/fixtures/goparsing/classification/models/nomodel.go
@@ -192,7 +192,7 @@ type AllOfModel struct {
 	CreatedAt strfmt.DateTime `json:"createdAt"`
 }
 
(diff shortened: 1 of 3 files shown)


casualjim:
would it be possible to provide a test case analogous to the other tests for these new or redefined types?

zasran:
Makes sense, adding for the valid types that we missed.

Not sure how to add test for the complex* types since it's a negative test, meaning parsing should fail. Would it make sense to add a test to classifier_test.go, something similar to other negative tests like TestDuplicateAnnotations?

zasran:
Updated pull request, added missing positive unit tests (byte, uintptr).

casualjim:
:shipit:

Issue #563, opened 2016-06-16
Embedding anonymous pointer causes error


zasran:
## Problem statement

Following code:

```Go
 type A struct {
        *B
}
```

causes error (regardless of type of B):

```
unable to resolve embedded struct for: &{991 B}
```

Seems like the problem is that the scan/schema.go parseEmbeddedType function does not support StarExpr.

casualjim:
fixed by the PR

PR #564, opened 2016-06-17
Fix issue #563 Added StarExpr in parseEmbeddedType

zasran:
Any suggestions on how to test this? Seems like the test in scan/classifier_test.go are similar, maybe add model with embedded pointer in fixtures/goparsing/classification/models/ ? Didn't find an example that would test whether the JSON model  is correct (e.g. we want the embedded type to be there as optional).

The committed change produces correct  swagger, as far as I can tell, i.e. the members of embedded type are all there and optional.

### DIFF
--- a/fixtures/goparsing/classification/models/nomodel.go
+++ b/fixtures/goparsing/classification/models/nomodel.go
@@ -192,6 +192,17 @@ type AllOfModel struct {
 	CreatedAt strfmt.DateTime `json:"createdAt"`
 }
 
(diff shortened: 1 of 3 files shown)


casualjim:
yeah just add a file that uses it and see if the correct properties are set on the swagger schema.

casualjim:
I merged the other one so this one needs updating :)

casualjim:
lgtm thanks!

PR #569, opened 2016-06-20
make validations optional for "generate model" command

### LINKED ISSUE #568: No way of avoiding generating validations
There seems to be no way of avoiding generating validations even if I run swagger with:

```
swagger generate model --skip-validator -f schema/defs.json -m specs
```

Tried with latest master

casualjim:
closes #568


runcom:
@casualjim ty

PR #589, opened 2016-07-12
Fixed go get path for the source install

myles-mcdonnell-m3:


casualjim:
this would make it so that you can't upgrade the CLI anymore, or get the executable when you run go get.

PR #617, opened 2016-08-09
Support json tags which don't set name

bfirsh:
For example, this should use the name "Foo" rather than the blank string:

    Foo string `json:",omitempty"`


casualjim:
lgtm, thanks!

PR #621, opened 2016-08-10
Refactor JSON tag parsing

bfirsh:
Following on from #617... turns out this was done in 3 places, so I've yanked it out into its own parsing function and added more tests for blank names in JSON tags.

Looks like there's some more duplication and refactoring that can be done here, but I suppose this is a good start.


casualjim:
:+1:  thanks for doing this :)

Issue #608, opened 2016-08-02
Generate spec looks up types in wrong package if there are several of the same name


bfirsh:
The message introduced in #437 is causing pretty sporadic error messages for me. Every time I run`swagger generate spec`, I get different paths for the error:

```
package github.com/coreos/etcd/pkg/types, error is: unknown builtin "Port"
```

```
package github.com/docker/containerd/api/grpc/types, error is: unknown builtin "Port"
```

... and so on. None of those packages have anything to do with the type "Port", nor have any swagger definitions in them.

I'm running off master.

casualjim:
Are the sources for what you're trying to run the generate spec against available? 
I'd like to take a look at the project and be able to reproduce the issue

bfirsh:
https://github.com/bfirsh/docker/tree/generate-swagger-specification
https://github.com/bfirsh/engine-api/tree/add-swagger-annotations

Note that it currently doesn't generate at all for reasons I can't figure out (`Port` is defined as a model, but it's still throwing an error for some reason). I've been running a version of go-swagger which ignores unknown types to try and debug: https://github.com/bfirsh/go-swagger/commit/5acf8482e5cd05a7f0fe4bec9155c7b063f41031

bfirsh:
This is quite hard to debug because the errors seem somewhat non-deterministic, but it seems the object might incorrect in this error too. With that patch applied to ignore unknown types, `Port` _sometimes_ shows up in the output, and _sometimes_ doesn't. If I disable some of the more complex routes (e.g. creating containers), then `Port` consistently shows up in the output.

Perhaps there is an object that is causing the error that isn't actually mentioned in the error? Perhaps files are getting processed in a different order, so `Port` sometimes gets shown in the output, but sometimes doesn't? Not sure... but might be a clue!

casualjim:
I won't be able to dig into this today, but I will for sure over the weekend.

bfirsh:
Fantastic, thanks!

This root problem causing the error is #241, I believe. It also hints at some of the non-deterministic error.

It's quite possible I'm just doing something wrong, but the error message isn't pointing me in the right place so it's very hard to figure out what is actually wrong.

casualjim:
I have an error about github.com/notary/tuf/store when I get the source of that I don't have the store package.
Is there a particular branch I should be looking at?

bfirsh:
[...]

PR #623, opened 2016-08-10
Fix packageForFile picking incorrect package

bfirsh:
If there are multiple packages of the same name, it might pick the wrong one. Pos will absolutely pick the correct file.

This fixes my app in #608, but it unfortunately breaks the test suite because `gofile.Pos()` (`../fixtures/goparsing/classification/models/nomodel.go`) is different to `fil.Pos()` (`/Users/ben/go/src/github.com/go-swagger/go-swagger/fixtures/goparsing/classification/operations/noparams.go`). I'm not sure why one is relative and one isn't, and I copied the code from elsewhere in the file, so I'm not sure why this doesn't work. Any ideas?

Also – not quite sure how to write a test for this, because I think the behaviour is non-deterministic depending on what order files are scanned...

### DIFF
--- a/scan/schema.go
+++ b/scan/schema.go
@@ -754,9 +754,13 @@ func (scp *schemaParser) createParser(nm string, schema, ps *spec.Schema, fld *a
 }
 
 func (scp *schemaParser) packageForFile(gofile *ast.File) (*loader.PackageInfo, error) {
-	for pkg, pkgInfo := range scp.program.AllPackages {
-		if pkg.Name() == gofile.Name.Name {
-			return pkgInfo, nil
+	for _, pkgInfo := range scp.program.AllPackages {
+		if pkgInfo.Importable {
+			for _, fil := range pkgInfo.Files {
+				if fil.Pos() == gofile.Pos() {
+					return pkgInfo, nil
+				}
+			}
 		}
 	}
 	fn := scp.program.Fset.File(gofile.Pos()).Name()


gaplyk:
When this gonna be merged? I faced same issue today and would like to have this asap :(

casualjim:
I'll look at making the tests pass tomorrow

casualjim:
It's not quite this fix, but I think the changes in the branch scan-package fix the same thing.
@gaplyk does that branch fix your issue?

@bfirsh I'm getting a docker build env together because you mention ignoring unkown type in a linked issue. I still want to repro

gaplyk:
Yes this fixed my issues. I can do example how to reproduce it on Monday. It is very simple :)

gaplyk:
And yes looks like this change in scan-package branch will fix same issue but in different way. :)

gaplyk:
https://github.com/gaplyk/go-swagger-issue
here is how to reproduce it

casualjim:
does this mean that the scan-package branch didn't fix your immediate issue?

casualjim:
for your repo I get output that looks correct now.

command:

```bash
swagger generate spec -o issue.json --scan-models
```

output: 

```json
{
  "swagger": "2.0",
  "paths": {},
  "definitions": {
    "Error": {
      "type": "object",
      "properties": {
        "Code": {
          "type": "string"
        }
      },
      "x-go-package": "github.com/gaplyk/go-swagger-issue/errors"
    },
    "ErrorResponse": {
      "x-go-package": "github.com/gaplyk/go-swagger-issue/errors",
      "$ref": "#/definitions/Error"
    }
  },
  "responses": {
    "ErrorResponse": {}
  }
}
```

gaplyk:
I meant scan-package will fix this as well as this PR.

PR #626, opened 2016-08-12
Maximum

nicolaifsf:
should be checking for length of string, rather than arbitrarily (trying) to convert a string to an float64


casualjim:
maximum is a validation for numbers. If you want to specify a max length for a string you should use maxLength as validation instead.
See here for more info: http://json-schema.org/latest/json-schema-validation.html#anchor26

PR #612, opened 2016-08-06
Unified CLI for go swagger. 

vburenin:
It should break all run scripts people used before ;)

Issue #587, opened 2016-07-01
CLI options to define HTTP/HTTPS protocol schema in generated server code.


vburenin:
There is a need to run a server in HTTP mode only behind nginx that adds TLS layer.
Because of that, the provided schema contains HTTP mode only, however server is HTTPS only. That results in the situation that clients like 'bravado' can't not work with this server.

So, It would be good If user could just define which schemas (http or https) to provide in a swagger spec.

ps. I will add all needed changes some time soon.

dhoelle:
@vburenin Anything I can do to help?

vburenin:
I haven't invest any time into this yet. Maybe soon.

vburenin:
Using --force-schema should override whatever default is.

--force-schema=http -> ["http"]
--force-schema=https -> ["https"]
--force-schema=http --force-schema=https -> ["http", "https"]

I am also thinking that server should not really have a different code to run HTTP and HTTPS server, both modes should be optional for any schema definitions.

Let say:
--http - enables HTTP
--https - enables HTTPS requiring TLS key/cert/ca.

@casualjim Do you have any opinion around it?

casualjim:
i would call it scheme and i prefer the --force-scheme over the --http so that we can add transports like grpc, amqp etc

vburenin:
Implemented in https://github.com/go-swagger/go-swagger/pull/612

Panic! It should break all scripts that are running a swagger application.

PR #629, opened 2016-08-16
revert changes to CLI params

casualjim:


nelz9999:
This was related to #587 and #612, right?

I was actually looking forward to those changes. Was there any further discussion about this?

casualjim:
no further discussion yet, but the changes won't be lost. Their intent will be brought back but in a less breaking fashion.
Some signatures changed without there being a real reason for the change and the args also changed too much for their intended use.

So I think the intent is we don't want the spec to decide between https, http or unix sockets. We want that to be decision that is made when we start the process of the server. Would you agree that that is the change you want preserved?

vburenin:
I want to be in the control where and how I listen to client connections as
well as what is in the spec shemes list independently.

nelz9999:
> So I think the intent is we don't want the spec to decide between https, http or unix sockets. We want that to be decision that is made when we start the process of the server. Would you agree that that is the change you want preserved?

Yup, that is exactly what I was looking for!

PR #642, opened 2016-08-25
Tests for issue with maps

gaplyk:
issue with parsing custom types for maps

```go
type MapType map[string]string
type MapTypeArr map[string][]string

type MyStruct struct {
    Field   MapType
    Field2   MapTypeArr
}
```

I have problem with writing test for it fast, just have to spend more time to understand what is going on there and create the test.
 
Looks like there is one issue more, but i didn't find it yet.
example is here: 
https://github.com/gaplyk/go-swagger-issue


casualjim:
Thanks! :+1:

PR #644, opened 2016-08-30
update deps

casualjim:


vendor/github.com/go-openapi/errors/schema.go
```diff
@@ -91,7 +117,7 @@ func (c *CompositeError) Error() string {
 // CompositeValidationError an error to wrap a bunch of other errors
 func CompositeValidationError(errors ...error) *CompositeError {
 	return &CompositeError{
-		code:    422,
+		code:    CompositeErrorCode,
```
vburenin: ValidationErrorCode?

vendor/github.com/go-openapi/errors/schema.go
```diff
@@ -65,6 +65,32 @@ const (
...
+	CompositeErrorCode = 422
+	// InvalidTypeCode is used for any subclass of invalid types
+	InvalidTypeCode = 600 + iota
+	RequiredFailCode
+	TooLongFailCode
+	TooShortFailCode
+	PatternFailCode
+	EnumFailCode
+	MultipleOfFailCode
+	MaxFailCode
```
vburenin: Does it even make sense to have these codes? It feels like an overkill.

vendor/github.com/go-openapi/errors/schema.go
casualjim: see: go-openapi/errors#4
and: go-openapi/errors#5

vendor/github.com/jessevdk/go-flags/parser_test.go
```diff
@@ -398,9 +398,30 @@ func TestOptionAsArgument(t *testing.T) {
 			args: []string{"-o", "-", "-"},
 			rest: []string{"-", "-"},
 		},
+		{
```
vburenin: Does it make sense to include unit tests for the vendored code?

vendor/github.com/go-openapi/errors/schema.go
casualjim: yes codes different from http status codes for errors make a lot of sense.
Not everything is string based, like translated error messages in a ui.

vendor/github.com/jessevdk/go-flags/parser_test.go
casualjim: I removed the tests from vendor

vendor/github.com/go-openapi/errors/schema.go
vburenin: In my opinion it is more interesting to know which parameters caused an error, validation constraints, and the expected type. Specific validation criteria that failed validation doesn't make a lot of sense, since I would just display a specific set of conditions for the user that needs to be taken into account rather than pointing what is exactly wrong. I am not saying it is all wrong, I just pointing it is an overkill for me.

PR #646, opened 2016-08-31
Issue with embedded custom type

gaplyk:
Code which cause error:
```go
// swagger:response SuccessResponse
type successResponseBody struct {
	// In: body
	Body SuccessResponse
}

// swagger:model SuccessResponse
type SuccessResponse map[string]string

type successData struct {
	SuccessResponse
}
```

example here: https://github.com/gaplyk/go-swagger-issue


casualjim:
thanks!

PR #660, opened 2016-09-08
add constants for string enums

rgarcia:
This adds constants for string enum model definitions. E.g., for this swagger model definition:

```
definitions:
  Foo:
    type: string
    enum:
      - bar
      - baz
```

you get:

```go
const (
        FooBar Foo = "bar"
        FooBaz Foo = "baz"
)
```


casualjim:
Thanks!

PR #662, opened 2016-09-11
Example project for a generating a streaming server

nelz9999:
I did this exercise to prove it out for myself... And I thought others might benefit from it too.

Signed-off-by: Nelz <nelz9999@gmail.com>


nelz9999:
Hrm... I see the build failed because it didn't find the generated code... Should I also check in generated resources as well?

casualjim:
yeah for examples the code is checked in so people can browse it. It's been hugely useful in helping people find their way. 
You can add your example to the regenerate script too: https://github.com/go-swagger/go-swagger/blob/master/hack/regen-samples.sh

casualjim:
Thanks a lot! more examples are really helpful 💯

Issue #673, opened 2016-09-30
Improve readability of templates used for code generation


dnephin:
golang templates [support the use of `{{- ... -}}`](https://golang.org/pkg/text/template/#hdr-Text_and_spaces) to remove extra whitespace between tags in templates.

This allows you to write templates which are easily readable by humans, without the extra whitespace changing the output of the rendered template.

I have found that using this style makes it a lot easier to understand the templates.

dnephin:
I've submitted an example of this in #674. If the PR is accepted I can make similar changes to a few other templates.

fredbi:
Most templates have now been refactored: some large ones have been split, the template logic indented, etc.
With the latest round of such refactoring efforts (#2221), we can say that this is now achieved.

PR #674, opened 2016-09-30
Make the template for schematype and structfield  more readable.

dnephin:
Starts to implement #673


casualjim:
this is great, the only reason I didn't do it before is because I wanted to remain compatible with the previous version of go as well. But now the last 2 versions of go accept the formatting, so I'm all for it.

However you will have to run `go generate ./generator` for your changes to make into the binary

dnephin:
Ah, that's what I was missing, thanks!

PR #680, opened 2016-10-05
Bug fixes for generating operation models

### LINKED ISSUE #676: Generating a model from an inline response schema
In the [swagger spec I'm working with](https://github.com/bfirsh/docker-api-reference/blob/master/swagger.yaml) there are lots of inlined response schemas. I'd like to generate a model for them using `swagger generate model ...`. As far as I understand it, currently you can only generate models for the schemas under `definitions`.

I'd like to add a feature to generate a model from these inline response schemas.  The name of the model would be `<operationId><response code>Response`.

Do you think there is value in this, or is there some easier way to work with inline schemas?

dnephin:
Fixes #676

* adds `--skip-validator` to `swagger generate operation` so that models can be generated without validation
* fixes a bug where anonymous response objects were created with a suffix of `BodyBody`. Now only one `Body` as appended
* adds support for setting the name of an anonymous schema from the `title`

If any of this is controversial I'd be happy to split it into a separate PR to get the rest merged.


casualjim:
thanks!

PR #686, opened 2016-10-10
generate constants for string enum properties

rgarcia:
This change generates constants for string enum model properties. E.g., for this model definition

```yaml
definitions:
  Book:
    type: object
    properties:
      genre:
        type: string
        enum:
          - scifi
          - mystery
          - horror
```

you get:

```go
const (
	BookGenreScifi   string = "scifi"
	BookGenreMystery string = "mystery"
	BookGenreHorror  string = "horror"
)
```

similar in spirit to https://github.com/go-swagger/go-swagger/pull/660


casualjim:
Thanks a lot :+1:

PR #689, opened 2016-10-11
Fix issue with empty schema 

gaplyk:
of properties when allOf is exists in the struct

example: 
```go 
type EmbeddedModel struct {
	Field string `json:"field"`
}

type AllOfStruct struct {
    // swagger:allOf
    EmbeddedModel
}
````

the result gonna be:
```json
"AllOfStruct":{
   "allOf":[
      {"type":"object","properties":{"field":{"type":"string","x-go-name":"Field"}}},
      {"type":"object"}
   ]
}
```


casualjim:
thanks again! Do you know of https://slackin.goswagger.io

PR #696, opened 2016-10-13
Use in-line comments for structfield and struct

dnephin:
First commit makes formatting errors fatal. Previously if there was an error here it would print a warning (which was very easy to miss because it's just one line in a lot of text) and not write anything to the file (because `formatted` was the empty string). With this change the unformatted content is written to the file (for debugging purposes) and the generation exits immediately (since the output is invalid).

Second and third commit change from `/* */` style comments to single line `//` comments, which are the go  convention.


casualjim:
thanks a lot :100:

PR #716, opened 2016-10-24
don't use omitempty on array fields

### LINKED ISSUE #707: can't distinguish between an empty array and null array for non-required array fields
## Problem statement

For this spec that contains a model with a non-required array field:

``` yaml
swagger: '2.0'
info:
  title: swagger spec with a model with a non-required array field
  version: 0.1.0
definitions:
  Foo:
    type: object
    properties:
      arrayThing:
        type: array
        items:
          type: string
```

The generated model looks like this:

```go
// Foo foo
// swagger:model Foo
type Foo struct {

	// array thing
	ArrayThing []string `json:"arrayThing,omitempty"`
}
```

The use of `omitempty` here is problematic when clients and servers would like to differentiate between the case when the array is omitted and when the array is empty.

When a client goes to marshal the omitted case, e.g. `Foo{}`, the result (as expected) is an empty JSON object, `{}`.

However, when a client goes to marshal the empty array case, e.g. `Foo{ArrayThing:[]string{}}`, the result is *also an empty JSON object* (`encoding/json` treats length zero slices as empty).

You can see this behavior here: https://play.golang.org/p/KTH36qo8mu

This means the server cannot distinguish between these two cases when processing `Foo`s sent by clients.

I think this is easily fixable by dropping `omitempty` on non-required array fields. This achieves the desired result: https://play.golang.org/p/nRYCj_w9-j

rgarcia:
fixes #707

### DIFF
--- a/generator/bindata.go
+++ b/generator/bindata.go
@@ -110,7 +110,7 @@ func templatesAdditionalpropertiesserializerGotmpl() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "templates/additionalpropertiesserializer.gotmpl", size: 2225, mode: os.FileMode(420), modTime: time.Unix(1472098838, 0)}
+	info := bindataFileInfo{name: "templates/additionalpropertiesserializer.gotmpl", size: 2225, mode: os.FileMode(420), modTime: time.Unix(1476136423, 0)}
 	a := &asset{bytes: bytes, info: info}
 	return a, nil
 }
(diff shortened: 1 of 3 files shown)


casualjim:
looks like you may need to re-run: go generate ./generator :)

https://ci.appveyor.com/project/casualjim/go-swagger/build/0.1.785-gjwxobfn#L122

rgarcia:
@casualjim thanks, just ran that and pushed

rgarcia:
looks like I need to update some tests...

PR #744, opened 2016-11-17
added enum support

morlay:
Just spike for the `enum`

The usage like `swagger: strfmt`, and will be two way to collection the enum values:

* if const be typed, we will use the value of const which defined by the target type
* if `iota` const setted, we will use the typename of const with prefix (snake upper case) equals target type

```go
// swagger:enum
type Level string

const (
	LEVEL_1 Level = "ONE"
	LEVEL_2 Level = "TWO"
	LEVEL_3 Level = "THREE"
)

// swagger:enum
type LevelInt int

const (
        // LEVEL_INT will be type and ONE will be the value
	LEVEL_INT__ONE = iota + 1
	LEVEL_INT__TWO
)

// swagger:model
type Model struct {
	level Level
	levelInt LevelInt;
}
```

will be

```yaml
definitions:
  Model:
     type: "object"
     properties:
        level: 
         type: "string"
         enum: 
           - "ONE"
           - "TWO"
           - "THREE"
       levelInt: 
         type: "string"
         enum: 
           - "ONE"
           - "TWO"
```

### DIFF
--- a/scan/enum.go
+++ b/scan/enum.go
@@ -1,0 +1,81 @@
+package scan
+
+import (
+	"go/ast"
+	"strconv"
+	"unicode"
+	"strings"
+)
+
+func upperSnakeCase(s string) string {
+	in := []rune(s)
+	isLower := func(idx int) bool {
+		return idx >= 0 && idx < len(in) && unicode.IsLower(in[idx])
+	}
+
+	out := make([]rune, 0, len(in) + len(in) / 2)
+
+	for i, r := range in {
+		if unicode.IsUpper(r) {
+			r = unicode.ToLower(r)
+			if i > 0 && in[i - 1] != '_' && (isLower(i - 1) || isLower(i + 1)) {
+				out = append(out, '_')
+			}
+		}
+		out = append(out, r)
+	}
+
+	return strings.ToUpper(string(out))
+}
+
+func getEnumBasicLitValue(basicLit *ast.BasicLit) interface{} {
+	switch(basicLit.Kind.String()){
+	case "INT":
+		if result, err := strconv.ParseInt(basicLit.Value, 10, 64); err == nil {
+			return result
+		}
+	case "FLOAT":
+		if result, err := strconv.ParseFloat(basicLit.Value, 64); err == nil {
+			return result
+		}
+	default:
+		return strings.Trim(basicLit.Value, "\"")
+	}
+	return nil;
+}
+
+func getEnumValues(file *ast.File, typeName string) (list []interface{}) {
+	for _, decl := range file.Decls {
+		genDecl, ok := decl.(*ast.GenDecl)
+
+		if !ok {
+			continue
+		}
+
+		if (genDecl.Tok.String() == "const") {
+			for _, spec := range genDecl.Specs {
+				if valueSpec, ok := spec.(*ast.ValueSpec); ok {
+					switch valueSpec.Type.(type) {
+					case *ast.Ident:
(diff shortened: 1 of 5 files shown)


scan/enum.go
```diff
@@ -0,0 +1,100 @@
...
+					}
+					return strings.TrimSpace(name), true
+				}
+			}
+		}
+	}
+	return "", false
+}
+
+func UpperSnakeCase(s string) string {
```
casualjim: This function should go in the go-openapi/swag repo in the same place as the ToGoName etc

scan/enum.go
```diff
@@ -0,0 +1,100 @@
...
+					}
+					return strings.TrimSpace(name), true
+				}
+			}
+		}
+	}
+	return "", false
+}
+
+func upperSnakeCase(s string) string {
```
casualjim: Could you put this in the go-openapi/swag package and make sure it also works for the commoninitialisms in the golint expected fashion?
https://github.com/go-openapi/swag/blob/master/util.go

casualjim:
Thanks for this, many people will be thankful for this!

scan/enum.go
morlay: Sorry, I am new starter of Golang.
It may be hard for me to manage codes cross deps.
This why I have not write testing codes too. 😿

casualjim: I see, this is great though what you're doing here.
You know one possible improvement could be to just use the value of the enum constant. Swagger allows for enums to be of any type, they can even be complex objects. Although in reality I think they are typically strings, int, uint

morlay: Yes. We can ignore the complex objects, it is not widely used.
And for stings and integer, we could follow the type which is annotated.

```
// swagger:enum
type Level int

const (
	LEVEL_1 Level = 1
	LEVEL_2 Level = 2
	LEVEL_3 Level = 3
)
```
will be
```
type: "integer"
enum:
 - 1
 - 2
 - 3
```

casualjim: yep I like that, you can also set the format according to the size of the integer

casualjim:
hi would it be possible to join the slack channel? https://slackin.goswagger.io/

I have another person in there who wants to work on enums, so it'd be good to coordinate efforts a little

casualjim:
could you rebase this on master?

scan/enum.go
gaplyk: it's better to move value from declaration to comments.

GlenDC:
@morlay are you still working on this? If not, I don't mind taking it over, to finish of this PR. As it's a feature needed by multiple people.

morlay:
@GlenDC Cool. 
I just use this quick version for my project as MVP.
I think @casualjim already complete some work, could you join Slack to catch up with them?

Thanks for taking over this. I am really busy with other work recently and bad network in China (I cannot login Slack agent 😞 )

timbunce:
Any news on this?

evillemez:
Just bumping this... it's a beautiful dream.  Is there still any interest in this?

eduardev:
+1 bump

jon-whit:
Any ETA on this? I really could use this support.

sorenbak:
+1 bump

sam016:
Thanks @morlay 
3 years and counting... Any update or is any help required?

fredbi:
@sam016 thank you for checking this out. To be honest I don't know why this nice piece of work could never be completed. Feel free to resume this line of work, as this is still considered a nice addition.

fredbi:
superseded by #2176

PR #771, opened 2016-12-02
fixed parameters sequence issue

### LINKED ISSUE #768: Maintain parameter sequence in parameters scanner
Currently ```generate spec``` command does not maintain the sequence of the parameters as found in the code. It seems that parameters are being added to the specification file in a random sequence.

## Steps to reproduce

This is the struct:
```go
// swagger:parameters operation
type Example struct {
	// in: query
	ID int64 `json:"id"`

	// in: query
	Score int32 `json:"score"`

	// in: query
	Name string `json:"name"`
}
```

Now if we generate a spec file for the code:
```
swagger generate spec
```

We would expect ```ID``` to be first in the spec file as it is the first parameter in the struct. But this is not the case:
```json
        "parameters": [
          {
            "type": "integer",
            "format": "int32",
            "x-go-name": "Score",
            "name": "score",
            "in": "query"
          },
          {
            "type": "string",
            "x-go-name": "Name",
            "name": "name",
            "in": "query"
          },
          {
            "type": "integer",
            "format": "int64",
            "x-go-name": "ID",
            "name": "id",
            "in": "query"
          }
        ]
```

**NOTE:** This also causes Swagger-UI to display parameters in the incorrect sequence.

## Cause

Since Go-1.4 maps are no longer kept in sequence when iterating over them. The recommended method to maintain the sequence of a map is to create a slice of all the keys in the required sequence, and then to iterate over the slice instead of the map. [reference](https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/)

I will fix this soon.

McSwitch:
Ensure that parameters scanner maintains the sequence of the parameters, as it is in the code.

closes #768

### DIFF
--- a/scan/parameters.go
+++ b/scan/parameters.go
@@ -280,6 +280,9 @@ func (pp *paramStructParser) parseStructType(gofile *ast.File, operation *spec.O
 			}
 		}
 
+		// a slice used to keep track of the sequence of the map keys, as maps does not keep to any specific sequence (since Go-1.4)
+		sequence := []string{}
+
 		for _, fld := range tpe.Fields.List {
 			if len(fld.Names) > 0 && fld.Names[0] != nil && fld.Names[0].IsExported() {
 				gnm := fld.Names[0].Name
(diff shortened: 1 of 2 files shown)


casualjim:
thanks!

PR #772, opened 2016-12-02
244 245 enums defaults for parameters [quick implementation]

McSwitch:


casualjim:
there is some overlap here with #744 which does more extensive enum support. Would it be possible to separate the 2?
If you're planning on contributing more (which would be awesome) we have a slack team. https://slacking.goswagger.io

Issue #738, opened 2016-11-14
Responses should support a description


hlubek:
## Problem statement

The `Responses:` of a `swagger:route` should have support to add a description (which is required by the Swagger specification).

Having a comment on the response type adds a description in the `definitions` but not on the `responses`.

## Swagger specification

See http://swagger.io/specification/#responseObject

## Steps to reproduce

1. Declare a `swagger:route` with `Responses:`
2. No description can be set (just a named response)
3. The `description` property is missing on the `responses` object in the generated JSON

clawconduce:
I was looking at adding this (which would fit right here: https://github.com/go-swagger/go-swagger/blob/master/scan/validators.go#L603) but I changed something that made my responses outputs looks like:

```
          "200": {
            "$ref": "#/responses/pet"
          },
          "500": {
            "$ref": "#/responses/pet"
          }
```

and I believe they were previously outputting the schema as a reference - which I mocked up below: 

```
          "200": {
            "schema": {
              "$ref": "#/responses/pet"
            }
          },
```

Whatever the change was, description is no longer required by the parser at http://editor.swagger.io/#/ .  In fact, when including a description, I get a warning that the field is ignored.

clawconduce:
I also wanted to point out that the reference is url encoded, so spaces turn into `%20`.  I was thinking about parsing value portion of the line based on the first space, so any text after the first space would become the description.  I am not sure what I changed in my comment or build command that stopped outputting schema, but I don't think I understand the problem well enough to code a solution.

clawconduce:
I see what the change was - I had a definition instead of a response and just missed that that changed.

I took a stab at a fix in the PR linked above.

casualjim:
I think this is fixed on master now, please close when confirmed

hlubek:
Hi, thx. for the replies and your work!

I still can't get it working though. Do you have an example how and where the description should go?

casualjim:
in your operation

```
// responses:
//   200: body:someModel and the rest is description
```

GlenDC:
Before closed, the documentation at https://github.com/go-swagger/go-swagger/blob/master/docs/generate/spec/route.md#example should probably be updated to reflect this new feature

hlubek:
Thx, I got it working so the description is added to the responses of the operation when using `body:myNamedResponse`.

[...]

PR #777, opened 2016-12-06
Issue #738: Support descriptions on responses that use definitions

clawconduce:


clawconduce:
I set the default description to the model name so that the field is not empty.

Alternatively, https://github.com/go-openapi/spec/blob/master/response.go#L25 could change description so it's not an `omitempty` field and the output is technically valid since the field is there, just the value is empty.

Issue #779, opened 2016-12-08
Question: Adding reason for response message in swagger:route annotation


Sakibs:
## Problem statement
I have 2 interrelated questions

1)
Is it possible to add the reason for the error within the swagger:route annotation where we define the list of responses?

```
// swagger:route GET /profile profile
//
// Get a user profile for specified user token.
//
//
//     Consumes:
//     - application/json
//
//     Produces:
//     - application/json
//     - text/plain
//
//     Schemes: http, https
//
//     Responses:
//       200: profile
//       401:
//       500:

```
for 401 I want to say Unauthorized without creating a swagger response annotation and struct. maybe something like (I tried the below option didnt work)

`// 401: description: Unauthorized
`

or even 
`// 401: reason: Unauthorized
`

Right now I have declared an empty struct that describes what I want the reason to be in the annotation. I feel it is cumbersome and not the best way to do it.

```
// Unauthorized
//
// swagger:response Unauthorized
type Unauthorized struct {
}

```

2)
Is it possible to define required headers inside the route itself?

so maybe have something like 

```
// swagger:route GET /profile profile
//
// Get a user profile for specified user token.
//
//
//     Consumes:
//     - application/json
//
//     Produces:
//     - application/json
//     - text/plain
//
//    Headers:
//        Authorization: The authorization token
//          - required
//
//     Schemes: http, https
//
//     Responses:
//       200: profile
//       401:
//       500:

```

Instead of again making a struct for the parameters that then connects to the route. Right now this is what I have implemented
```
// swagger:parameters profile
type Authorization struct {
	// Authorization header for a users bearer token. 
	// in: header
	// required: true
	Authorization string
}
```

## Swagger specification
swagger 2.0 json

casualjim:
I can see why you want a description, but at that point the swagger spec is nothing more than a documentation tool. So I don't know what you would gain by that over just using or a markdown document.

When you define headers like you've described in your proposal, we lose type information. The point of making you define structs is so that you get the type checking from the compiler, and overall it leads to a better defined code base and API.

As for your second point, are you saying you want to reuse a partial parameter struct to also serve as list of headers for a response.  Because I think that might be a good addition. The problem arises when there are things defined on the parameter struct that aren't headers or a body, do you have a solution for that?

Sakibs:
I want the description so I wouldnt have to make a bunch of empty structs for status codes that just return the status code text in the body.

[...]

PR #788, opened 2016-12-15
Support description tags in responses and support description as the …

clawconduce:
…only tag (#779)


clawconduce:
I feel much less certain about this PR than my original.  Feel free to tear it apart, comments welcome!

PR #792, opened 2016-12-18
fix #767: field-name-not-showing-if-using-ref

### LINKED ISSUE #767: Definition field name not showing when string pattern validation failed, if use $ref to reference another one in defintions
## Problem statement
When a pattern validation fails on a field in request body, which is a type definition referencing to another definition, the returned error message won't show the validation failed field name, causing trouble to read what field failed at validation. 

## Swagger specification
'''
swagger: "2.0"
info:
  description: The product of a tutorial on goswagger.io
  title: A To Do list application
  version: 1.0.0
consumes:
- application/io.goswagger.examples.todo-list.v1+json
produces:
- application/io.goswagger.examples.todo-list.v1+json
schemes:
- http
paths:
  /:
    get:
      tags:
        - todos
      operationId: findTodos
      parameters:
        - name: since
          in: query
          type: integer
          format: int64
        - name: limit
          in: query
          type: integer
          format: int32
          default: 20
      responses:
        200:
          description: list the todo operations
          schema:
            type: array
            items:
              $ref: "#/definitions/item"
        default:
          description: generic error response
          schema:
            $ref: "#/definitions/error"
    post:
      tags:
        - todos
      operationId: addOne
      parameters:
        - name: body
          in: body
          schema:
            $ref: "#/definitions/item"
      responses:
        201:
          description: Created
          schema:
            $ref: "#/definitions/item"
        default:
          description: error
          schema:
            $ref: "#/definitions/error"
  /{id}:
    parameters:
      - type: integer
        format: int64
        name: id
        in: path
        required: true
    put:
      tags:
        - todos
      operationId: updateOne
      parameters:
        - name: body
          in: body
          schema:
            $ref: "#/definitions/item"
      responses:
        200:
          description: OK
          schema:
            $ref: "#/definitions/item"
        default:
          description: error
          schema:
            $ref: "#/definitions/error"
    delete:
      tags:
        - todos
      operationId: destroyOne
      responses:
        204:
          description: Deleted
        default:
          description: error
          schema:
            $ref: "#/definitions/error"
definitions:
  IdentifierType:
    type: string
    format: identifier
    pattern: ^[A-Za-z][-A-Za-z0-9_]*$
  item:
    type: object
    required:
      - description
    properties:
      id:
        type: integer
        format: int64
        readOnly: true
      description:
        $ref: '#/definitions/IdentifierType'        #<------ use another definition above, this causes the issue
#<----- if replacing the $ref reference to its direct definition, the issue is gone.
        minLength: 1
      completed:
        type: boolean
  error:
    type: object
    required:
      - message
    properties:
      code:
        type: integer
        format: int64
      message:
        type: string
'''

## Steps to reproduce
a git repo can be provided if needed.

Simon-Li:
fix the problem: 
when a pattern validation fails on a field in request body, which is a type definition referencing to another definition, the returned error message won't show the validation failed field name, causing trouble to read what field failed at validation.

### DIFF
--- a/generator/templates/schemavalidator.gotmpl
+++ b/generator/templates/schemavalidator.gotmpl
@@ -157,7 +157,7 @@ if err := validate.Required({{ if .Path }}{{ .Path }}{{else}}""{{end}}, {{ print
 {{ end }}
 {{ if .IsNullable }}if {{ .ValueExpression }} != nil {
 {{ end }}
-if err := {{.ValueExpression}}.Validate(formats); err != nil {
+if err := {{.ValueExpression}}.Validate(formats, "{{pascalize .Name}}"); err != nil {
   return err
 }
 {{ if .IsNullable }}}{{ end }}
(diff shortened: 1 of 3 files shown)


generator/templates/schemavalidator.gotmpl
```diff
@@ -247,6 +247,18 @@ func ({{ .ReceiverName }} *{{ if .IsExported }}{{ pascalize .Name}}{{ else }}{{
 // Validate validates this {{ humanize .Name }}
 func ({{.ReceiverName}} {{ if or .IsTuple .IsComplexObject .IsAdditionalProperties }}*{{ end }}{{ if .IsExported }}{{ pascalize .Name}}{{ else }}{{ .Name }}{{ end }}) Validate(formats strfmt.Registry) error {
   var res []error
+
+  {{ if .IsPrimitive }}
+    var propName string
```
casualjim: wouldn't you be able to set the prop name just to {{ pascalize .Name }} and not have a runtime component to this?

Simon-Li: The use case of this particular issue is I'm using a common modal component (e.g. `type IdentifierType string`), while this common modal (`IdentifierType`) are referenced by many different places of path/modal/property definition through `$ref`. When the `schemavalidator` template is rendered, the `IdentifierType` modal validator (methods) are not able to know what components it's validating for, actually the `IdentifierType` modal is just a common modal for other components. While, in runtime, who's calling the `IdentifierType` is determined and so being able to parsed out. This's mostly based on testing and may also be constrained by the implementation, please update if any concerns.

casualjim:
I had to revert this once I saw it introduces a breaking change in the generated code.
I think this needs an extra method: ValidateNamed which can have the name as extra argument.
Then an aliased type should be validated with the ValidateNamed method while other things can continue to use the Validate method.

Simon-Li:
@casualjim This makes most sense. I'll create another story to implement.

PR #816, opened 2016-12-25
improve defaults API constructor, fixes #810

### LINKED ISSUE #810: Improve Defaults of Server (API) Configuration
## Problem statement

When developing a service using go-swagger, a server can be automatically generated. On top of that a server cli application can be created, but one can be manually created as well. Before one can call `server.Serve()` the api needs to be properly configured AND the server has to have the handler set. Currently that process is in my opinion quite error prone.

## Steps to reproduce

Generate a server based on any valid OpenAPI spec and then create a server cli app more or less as follows:

```go
package main

// imports...

func main() {
	swaggerSpec, err := loads.Analyzed(restapi.SwaggerJSON, "")
	if err != nil {
		log.Fatalln(err)
	}

	api := foo.NewFooAPI(swaggerSpec)

	server := restapi.NewServer(api)
	defer server.Shutdown()

	// calling any route will make the server app panic (and recover)
	// the reason being that server.ConfigureAPI() isn't called
	// instead the user choose to set the handler herself,
	// but forgot to set the producers and consumers among other things.

	server.SetHandler(api.Serve(nil))
	if err := server.Serve(); err != nil {
		log.Fatalln(err)
	}
}
``` 

When you use the generated server cli instead you'll have it calling `server.ConfigureAPI()`, making it not panic and instead return you the expected NotImplemented errors, when calling any route.

## Proposed solution

While this work greats, it's not a good solution I think. The generated configureAPI function is pretty useless, which makes sense I suppose, knowing that it seems to be a file that is meant to be edited and thus being made useful.

Personally however I think we should rely on better defaults and I think I have a solution that doesn't break any code for people using older versions.

> Make the API constructor have reasonable defaults.

For this we have to change a constructor such as this:

```go
func NewFooAPI(spec *loads.Document) *FooAPI {
	return &FooAPI{
		handlers:        make(map[string]map[string]http.Handler),
		formats:         strfmt.Default,
		defaultConsumes: "application/json",
		defaultProduces: "application/json",
		ServerShutdown:  func() {},
		spec:            spec,
	}
}
```

to:

```go
func NewFooAPI(spec *loads.Document) * FooAPI {
	return & FooAPI{
		handlers:        make(map[string]map[string]http.Handler),
		formats:         strfmt.Default,
		defaultConsumes: "application/json",
		defaultProduces: "application/json",
		ServerShutdown:  func() {},
		// basically giving the `ServerError`, `producers`
		// and `consumers` a decent default
		ServeError:      errors.ServerError,
		JSONConsumer:    runtime.JSONConsumer(),
		JSONProducer:    runtime.JSONProducer(),
		TxtProducer:     runtime.TextProducer(),
		spec:            spec,
	}
}
```

Doing this makes it so that the API as it is just works. On top of that we could also make all the handlers by default _NotImplementedHandlers_, such that the simple example given above just works, without any weird unexpected panics.

Another idea would be that the `configureAPI` function wouldn't just stupidly override all handlers, and instead first check if the handlers are still nil. And only override a handler if it is indeed still nil.

Lastly, one could even forget to set the handler (e.g. via `server.SetHandler(api.Serve(nil))`) and not get a panic or fatal error at start up to warn about this. Instead no routes will be available and everything will return a `404`. These errors might tip of the user, but perhaps explicitly shouting to the user about this problem at start-up of the cmd app might be more helpful and beginner-friendly?

GlenDC:
Before this PR all defaults were done in configureAPI, right before calling serve, this made that function pretty useless without modifying. With this PR that function because optional and is only to be used if that's how you want to customize a server. I would however advice against it, as it mixes generated code and manal code. I think it's a much better idea to do the customization similar to how it's done in the go-openapi/kvstore example.

With the modifications in the PR, creating a server cmd can be as simple as following:

```go
package main

// imports...

func main() {
	swaggerSpec, err := loads.Analyzed(restapi.SwaggerJSON, "")
	if err != nil {
		log.Fatalln(err)
	}

	api := foo.NewFooAPI(swaggerSpec)

	server := restapi.NewServer(api)
	defer server.Shutdown()

	api.FooHandler = myNewFooHandler()

	if err := server.Serve(); err != nil {
		log.Fatalln(err)
	}
}
```

And this will actually work as expected. In this example the producers, consumers and server handler are all set by default, and can run as it is. This doesn't prevent any of the customization as was possible before, but it does make things a lot easier for when the defaults (based on the given swagger spec) are good as they are.


GlenDC:
The changes made in this PR are based on the suggestions made in the enhancement issue #810.

PR #829, opened 2016-12-29
828-Support-default-value-for-model-property

Simon-Li:
If there is another place able to pull the defaults without introducing a new `Default` field in `GenSchema` (and feeding into the templates), the following would be revised.

### DIFF
--- a/generator/templates/schema.gotmpl
+++ b/generator/templates/schema.gotmpl
@@ -94,7 +94,7 @@ func unmarshal{{ pascalize .Name }}(data []byte, consumer runtime.Consumer) ({{ 
 func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalJSON(b []byte) error {
   type {{ pascalize .Name }}Alias {{ pascalize .Name }}
   var t {{ pascalize .Name }}Alias
-  if err := json.Unmarshal([]byte(`{{json .Default}}`), &t); err != nil {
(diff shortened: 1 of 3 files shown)


generator/templates/schema.gotmpl
```diff
@@ -90,6 +90,19 @@ func unmarshal{{ pascalize .Name }}(data []byte, consumer runtime.Consumer) ({{
 
 }
 {{ else }}{{ if or .IsComplexObject .IsTuple .IsAdditionalProperties }}{{ if .Name }}type {{ if not .IsExported }}{{ .Name }}{{ else }}{{ pascalize .Name }}{{ end }} {{ end }}{{ template "schemaBody" . }}
+{{if .Default}}
+func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalJSON(b []byte) error {
+  type {{ pascalize .Name }}Alias {{ pascalize .Name }}
+  var t {{ pascalize .Name }}Alias
+  if err := json.Unmarshal([]byte(`{{json .Default}}`), &t); err != nil {
```
casualjim: It's safer to use: `[]byte({{ printf "%q" (json .Default) }})`
Otherwise if you have a string property in the json that contains a backtick it won't compile

Simon-Li: Thanks!

casualjim:
Thanks!

Issue #837, opened 2017-01-03
Implement ValidateNamed method to support validation errors with proper names


Simon-Li:
Per @casualjim advice, a new method (ValidateNamed) should be considered to add in order to support validation errors with proper names. This is especially useful for the use case of aliased property types, in which a model can be referenced commonly (via $ref) by another property. For the sake of cleanness for the existing `Validate` implementation, the `ValidateNamed` method is therefore taken into this implementation.

casualjim:
I've been thinking a bit more about this and because the thing being validated would always be called from another method., that method could just rewrite the errors by adding the field name info in there.
Less invasive

Simon-Li:
@casualjim Woo... great hints! will verify it.

PR #840, opened 2017-01-04
837-support-validation-error-with-proper-name

Simon-Li:
The implementation is hinted by #837 .


generator/templates/schemavalidator.gotmpl
```diff
@@ -158,7 +158,7 @@ if err := validate.Required({{ if .Path }}{{ .Path }}{{else}}""{{end}}, {{ print
 {{ if .IsNullable }}if {{ .ValueExpression }} != nil {
 {{ end }}
 if err := {{.ValueExpression}}.Validate(formats); err != nil {
-  return err
+  return err.(*errors.Validation).ValidateName({{ if .Path }}{{ .Path }}{{else}}""{{end}})
```
casualjim: we may want to make this a safer cast.

```go
if ve, ok := err.(*errors.Validation); ok {
  ve.ValidateName(...
}
```

I think it will otherwise panic if it turns out to be something other than a *errors.Validation

vendor/github.com/go-openapi/errors/headers.go
```diff
@@ -74,3 +74,12 @@ func InvalidResponseFormat(value string, allowed []string) *Validation {
 		message: fmt.Sprintf(responseFormatFail, allowed),
 	}
 }
+
+// Validate error message name for aliased property
+func (e *Validation) ValidateName(name string) *Validation {
```
casualjim: can you make these changes in: https://github.com/go-openapi/errors/blob/master/headers.go
They my get lost otherwise because this is in a vendored package, which get overwritten when I update the vendored deps

generator/templates/schemavalidator.gotmpl
Simon-Li: Right!

vendor/github.com/go-openapi/errors/headers.go
Simon-Li: Sure!

Simon-Li: I think the above change in `vendor/github.com/go-openapi/errors/headers.go` still need to be kept, right?

casualjim: if you install this tool: https://github.com/FiloSottile/gvt
you can run `gvt update github.com/go-openapi/errors`
and push that change

casualjim:
so sorry I didn't see it the first time around.
you still need to run `go generate ./generator` because you changed the template, and without it the changes won't show up in bindata.go

Simon-Li:
Should I run `go generate ./generator` and push the result if a template is changed? Previously I changed templates but did not do this step though.

casualjim:
yes if templates are changed then bindata needs to get changed too, in the binary that's how we distribute the templates

Simon-Li:
Got it, thank-you! @casualjim

PR #843, opened 2017-01-05
expose transport config, fixes #838

### LINKED ISSUE #838: Allow customized client constructor while respecting defaults
## Problem statement

Currently there are only 2 provided ways to create a generated API client:

1. Create a client with default `basePath`, `host`, `schemes` and `httpTransport`;
2. Create a client with a custom `httpTransport`, hardcoding the `basePath`, `host` and `schemes`;

Ideally it should _also_ be possible to:

- Create a client with all defaults, except 1 customized one (e.g. a custom `host`, as is often required for public API's);
- Create a client with a custom `httpTransport`, without having to hardcode the `basePath`, `host` and `schemes`;

GlenDC:
Fixes issue #838.

## Requests from issue

> Create a client with a custom httpTransport, without having to hardcode the basePath, host and schemes;

This is now possible by using the constants provided by the client package: `DefaultHost`, `DefaultBasePath` and `DefaultSchemes`.

> Create a client with all defaults, except 1 customized one (e.g. a custom host, as is often required for public API's);

This is now possible by calling `NewHTTPClientWithConfig` instead of `NewHTTPClient`:

```go
NewHTTPClientWithConfig(nil, DefaultTransportConfig().WithHost("http://localhost:9000"))
```

Logic from `NewHTTPClient` has been moved to `NewHTTPClientWithConfig`. `NewHTTPClient` calls `NewHTTPClientWithConfig` now, with a nil value for the config value, making it default via `DefaultTransportConfig`.

### DIFF
--- a/generator/bindata.go
+++ b/generator/bindata.go
@@ -111,7 +111,7 @@ func templatesAdditionalpropertiesserializerGotmpl() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "templates/additionalpropertiesserializer.gotmpl", size: 2225, mode: os.FileMode(420), modTime: time.Unix(1482416923, 0)}
+	info := bindataFileInfo{name: "templates/additionalpropertiesserializer.gotmpl", size: 2225, mode: os.FileMode(420), modTime: time.Unix(1476889131, 0)}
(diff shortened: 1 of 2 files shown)


GlenDC:
Can't review it @casualjim, but it does pass the tests.

Issue #844, opened 2017-01-05
Wrong field selectors are generated when using allOf in response schema definition


rjeczalik:
## Problem statement

When allOf is used in response schema definition, generated code does no build due to erroneous field selectors.

## Swagger specification

```json
{
  "swagger": "2.0",
  "info": {
    "title": "anyOf bug",
    "version": "0.0.1",
    "description": "repro",
    "license": {
      "name": "Apache 2.0",
      "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
    }
  },
  "definitions": {
    "foo": {
      "type": "object"
    },
    "bar": {
      "type": "object"
    }
  },
  "paths": {
    "/": {
      "get": {
        "responses": {
          "200": {
            "description": "OK",
            "schema": {
                "allOf": [
                    {
                        "$ref": "#/definitions/foo"
                    },
                    {
                        "$ref": "#/definitions/bar"
                    }
                ]
            }
          }
        }
      }
    }
  }
}
```

## Steps to reproduce

```
$ swagger generate client -f swagger.json
$ cat client/operations/get_responses.go | grep -A4 -E 'type\ GetOKBody|o\.models\.'
type GetOKBody struct {
        models.Foo

        models.Bar
}
--
        o.models.Foo = getOKBodyAO0

        var getOKBodyAO1 models.Bar
        if err := swag.ReadJSON(raw, &getOKBodyAO1); err != nil {
                return err
--
        o.models.Bar = getOKBodyAO1

        return nil
}

--
        getOKBodyAO0, err := swag.WriteJSON(o.models.Foo)
        if err != nil {
                return nil, err
        }
        _parts = append(_parts, getOKBodyAO0)
--
        getOKBodyAO1, err := swag.WriteJSON(o.models.Bar)
        if err != nil {
                return nil, err
        }
        _parts = append(_parts, getOKBodyAO1)
```

gregmarr:
@rjeczalik You say anyOf in the text, but allOf in the swagger file.

rjeczalik:
@gregmarr Fixed title - was thinking about something else, thanks for pointing out.

GlenDC:
@rjeczalik can you update your original issue description as well, as it still says `anyOf` instead of `allOf`.

PR #845, opened 2017-01-05
generator: strip package name from type name

rjeczalik:
Quick fix for #844.


casualjim:
can you run `go generate ./generator` so that the template changes are included in the binary?

rjeczalik:
@casualjim Done.

What about adding actual stale file check to your ci? It'd be something like:

```diff
$ git diff HEAD~1
diff --git a/appveyor.yml b/appveyor.yml
index c072bad..6a03635 100644
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -44,6 +44,10 @@ install:
   - go get -u gopkg.in/matm/v1/gocov-html
   - go get -u github.com/cee-dub/go-junit-report
 
+  # check for stale files
+  - go generate -v ./generator
+  - git diff --exit-code ./generator
+
 build_script:
   - cd %APPVEYOR_BUILD_FOLDER%
   - echo %PATH%
diff --git a/generator/shared.go b/generator/shared.go
index 81df7c3..38564ef 100644
--- a/generator/shared.go
+++ b/generator/shared.go
@@ -34,7 +34,7 @@ import (
        "golang.org/x/tools/imports"
 )
 
-//go:generate go-bindata -pkg=generator -ignore=.*\.sw? ./templates/...
+//go:generate go-bindata -mode 420 -modtime 1482416923 -pkg=generator -ignore=.*\.sw? ./templates/...
 
 // LanguageOpts to describe a language to the code generator
 type LanguageOpts struct {
```

casualjim:
sure if you want to add by all means :)

rjeczalik:
@casualjim Please don't merge yet, testing if the check works.

rjeczalik:
@casualjim Ok, looks like .git directory is removed from go-swagger before install stage:

```
git diff --exit-code ./generator
Not a git repository
To compare two paths outside a working tree:
usage: git diff [--no-index] <path> <path>
Command exited with code 129
```

It'd require more work, for a separate PR - let's merge the fix first.

rjeczalik:
Previous fix was wrong, I've stripped package from wrong variable - it should be GoType instead.

How to best add a test for building fixtures/codegen/issue844.json?

GlenDC:
@rjeczalik you can do so by adding a (unit) Test function named `TestIssue844` in https://github.com/go-swagger/go-swagger/blob/master/generator/model_test.go. There you can load the spec file, which it should be able to load without any errors. After that you could, if applicable, also test specifics. In the `model_test.go` file you'll see other examples that could give you inspiration if needed.

GlenDC:
Once you added the unit test @rjeczalik, you'll have to also fetch & rebase the latest master branch into your personal feature branch. You can resolve the conflict of `generator/bindata.go`, simply by generating a new one yourself, which contains both the latest templates (which you've just fetched & rebased) and your changes already committed in this PR.

Once that's done this PR is good to go :)

rjeczalik:
I've added a test that generates a client and then tries to build it. The test currently fails due to a different problem:

```
--- FAIL: TestGenerateAndBuild (0.28s)
    --- FAIL: TestGenerateAndBuild/issue_844 (0.28s)
        build_test.go:40: go get ../fixtures/bugs/844/generated398758314/...: exit status 2
                # github.com/go-swagger/go-swagger/fixtures/bugs/844/generated398758314/client/operations
                ../fixtures/bugs/844/generated398758314/client/operations/get_responses.go:117: o.Foo.Validate undefined (type models.Foo is interface with no methods)
                ../fixtures/bugs/844/generated398758314/client/operations/get_responses.go:121: o.Bar.Validate undefined (type models.Bar is interface with no methods)
```

Indeed the generated models lack a Validate method. Not quite sure it's a regression, as every other test passes:

```
$ go test -run -TestGenerateAndBuild
PASS
ok      github.com/go-swagger/go-swagger/generator      0.197s
```

Probably we'd want to have this problem also fixed before merging, so we do not have a broken master.

rjeczalik:
I've added a test that generates a client and then tries to build it. The test currently fails due to a different problem:

```
--- FAIL: TestGenerateAndBuild (0.28s)
    --- FAIL: TestGenerateAndBuild/issue_844 (0.28s)
        build_test.go:40: go get ../fixtures/bugs/844/generated398758314/...: exit status 2
                # github.com/go-swagger/go-swagger/fixtures/bugs/844/generated398758314/client/operations
                ../fixtures/bugs/844/generated398758314/client/operations/get_responses.go:117: o.Foo.Validate undefined (type models.Foo is interface with no methods)
                ../fixtures/bugs/844/generated398758314/client/operations/get_responses.go:121: o.Bar.Validate undefined (type models.Bar is interface with no methods)
```

Indeed the generated models lack a Validate method. Not quite sure it's a regression, as every other test passes:

```
$ go test -run -TestGenerateAndBuild
PASS
ok      github.com/go-swagger/go-swagger/generator      0.197s
```

Probably we'd want to have this problem also fixed before merging, so we do not have a broken master.

GlenDC:
@rjeczalik I think your test file triggered a bug in the current generator code already on master.

The template responsible for your non-compiable code can be found here: https://github.com/go-swagger/go-swagger/blob/master/generator/templates/schemavalidator.gotmpl#L251

The appropriate fix would probably be to ensure that we don't try to validate in case it's an interface. So within that range you're probably missing an if/conditional check to ensure it's indeed not an interface or I suppose it's called discriminator in Swagger/OpenAPI terms.

Would be cool if you could fix this alongside your other commits within this PR.

Also don't forget to rebase origin/master into your branch.

rjeczalik:
> Also don't forget to rebase origin/master into your branch.

Already rebased.

> Would be cool if you could fix this alongside your other commits within this PR.

Sure, will take a look.

rjeczalik:
@GlenDC Fixed, please take a look.

GlenDC:
Thank you for the extra effort @rjeczalik, very much appreciated.

rjeczalik:
@GlenDC You're welcome :-)

PR #849, opened 2017-01-07
add HTTPClient parameter functions

GlenDC:


generator/templates/client/parameter.gotmpl
```diff
@@ -104,6 +117,16 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) SetContext(ctx context.C
   {{ .ReceiverName }}.Context = ctx
 }
 
+// WithHTTPClient adds the HTTPClient to the {{ humanize .Name }} params
+func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) WithHTTPClient(client *http.Client) *{{ pascalize .Name }}Params {
```
casualjim: also needs a SetXxx without a return type so that it can be used as interface

generator/templates/client/parameter.gotmpl
```diff
@@ -65,6 +65,19 @@ func New{{ pascalize .Name }}ParamsWithContext(ctx context.Context) *{{ pascaliz
   }
 }
 
+// New{{ pascalize .Name }}ParamsWithHTTPClient creates a new {{ pascalize .Name }}Params object
+// with the default values initialized, and the ability to set a custom HTTPClient for a request
+func New{{ pascalize .Name }}ParamsWithHTTPClient(client *http.Client) *{{ pascalize .Name }}Params {
```
casualjim: Also needs a constructor with the context if you need a constructor like this.

generator/templates/client/parameter.gotmpl
GlenDC: @casualjim isn't that this one: https://github.com/go-swagger/go-swagger/pull/849/commits/e50ccb700cd77d1a0fdbc7a7ce6d5e29a946a4fc#diff-643df0026bfd7f1638796085caeb52f5R127?

generator/templates/client/parameter.gotmpl
GlenDC: @casualjim isn't that why you have the WithContext function, for situations like that? Otherwise one would have to provide N constructors, where N is the amount of combinations possible.

generator/templates/client/parameter.gotmpl
casualjim: o right, didn't see it

Issue #833, opened 2017-01-02
Validate client parameters before sending


GlenDC:
Currently (generated) client parameters aren't validated before being sent.

Relevant _TODO_: https://github.com/go-swagger/go-swagger/blob/master/generator/templates/client/client.gotmpl#L42

Thanks to the swagger there is all kind of information that is exposed about what we expect from the different parameters, if any. Yet, currently none of that information seems to be used:

+ if _required_, validate if present;
+ if _format_ is defined, see if given value respects this format;
+ if _ allowEmptyValue_ is set to `false` (default), make sure value isn't empty (Query/formData only);
+ Validate _minimum_, _maximum_ and relevant info, if defined and the value is a numeric type;
+ Validate _pattern_ if defined;
+ Validate _minItems_, _maxItems_, _uniqueItems_ if the value is an array/slice and those are defined;

PR #850, opened 2017-01-08
WIP: validate client params

GlenDC:
fixes issue #833


examples/task-tracker/client/tasks/add_comment_to_task_parameters.go
```diff
@@ -112,6 +133,16 @@ func (o *AddCommentToTaskParams) SetID(id int64) {
 	o.ID = id
 }
 
+// Validate these params
+func (o *AddCommentToTaskParams) Validate(formats strfmt.Registry) error {
+
+	if err := o.Body.Validate(formats); err != nil {
```
GlenDC: @casualjim this fails to compile, as that body model has not defined the Validate functions. This is however defined for the Body called here: https://github.com/go-swagger/go-swagger/pull/850/files#diff-ab37f2908b4671ebce8d3bc51539d622R123 ; I'm a bit stuck on this one, could you point me to the correct template location on how I could ensure that body model has the validation defined? Or is it because this model is automatically appended in the generator code (via the name+Body stuff)? If i'm doing crazy stuff in this commit you can also point it out though

examples/task-tracker/client/tasks/update_task_parameters.go
```diff
@@ -49,19 +49,30 @@ func NewUpdateTaskParamsWithContext(ctx context.Context) *UpdateTaskParams {
...
+	  In: body
 	*/
 	Body *models.Task
 	/*ID
 	  The id of the item
-
+	  Required: true
+	  In: path
 	*/
 	ID int64
```
GlenDC: This ID parameter is required, but for now Required isn't validated in my added Validate Client Parameter code. Any idea how I would be able to add validation for required parameters @casualjim ?

examples/task-tracker/client/tasks/add_comment_to_task_parameters.go
casualjim: That struct should actually have a Validate method, it has 2 required fields, so there is a reason it's not rendering that validate method

GlenDC: so are you saying, @casualjim, that the reason it is not rendering is an unreported bug?

casualjim: yeah that's what it looks like because the same issue exists on the server side

GlenDC: Ok weird, I'll try to track that issue later today. Can you please also check if my conditionals added in the template are reasonable, as you might have a better idea whether or not those will hold up for all permutations.

examples/task-tracker/client/tasks/update_task_parameters.go
GlenDC: The biggest question I have with this one, how to check a non-nil value. As I tried some string query parameter as required in a local test API, and it seemed to render as an ordinary string.

casualjim: I think there is an issue here: https://github.com/go-swagger/go-swagger/blob/master/generator/operation.go#L903

GlenDC:
Things left to do in order for the code in this PR to work and thus become merge-able:

- [ ] make sure Validate method is defined for all models/bodies structs;
- [ ] make sure the Required validation always is checked;
- [ ] rebase latest master and regenerate bindata;

PR #851, opened 2017-01-09
fix #847

### LINKED ISSUE #847: Type using allOf in generate server has broken MarshalJSON and UnmarshalJSON in 0.8.0
## Problem statement

A type, using allOf, generated by 0.7.4, doesn't implement MarshalJSON and UnmarshalJSON and works properly.  This same type, generated by 0.8.0, includes these functions, and they're broken.  
## Swagger specification

```yaml
swagger: '2.0'

info:
  title: ""

paths:
  /execute:
    post:
      responses:
        200:
          schema:
            type: "array"
            items:
              $ref: "#/definitions/ExecuteParametersWorkspace"

definitions:
  ExecuteParameters:
    type: "object"
    properties:
      WantGeometry:
        type: "boolean"

  ExecuteParametersWorkspace:
    allOf:
    - $ref: "#/definitions/ExecuteParameters"
    - required:
      - "Workspace"
      properties:
        Workspace:
          type: "string"
```

## Steps to reproduce

`swagger generate server`

Broken functions in 0.8.0:

```go
// UnmarshalJSON unmarshals this object from a JSON structure
func (m *ExecuteParametersWorkspace) UnmarshalJSON(raw []byte) error {

	var aO0 ExecuteParameters
	if err := swag.ReadJSON(raw, &aO0); err != nil {
		return err
	}
	m.ExecuteParameters = aO0

	return nil
}

// MarshalJSON marshals this object to a JSON structure
func (m ExecuteParametersWorkspace) MarshalJSON() ([]byte, error) {
	var _parts [][]byte

	aO0, err := swag.WriteJSON(m.ExecuteParameters)
	if err != nil {
		return nil, err
	}
	_parts = append(_parts, aO0)

	return swag.ConcatJSON(_parts...), nil
}
```

Simon-Li:
This case happens if AllOf contains an inline property definition which is missed by (un)marshal rendering. Proposed below tweaks for the fix. 
If there is no further change, will push binddata update.


generator/templates/tupleserializer.gotmpl
```diff
@@ -208,7 +208,15 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalJSON(raw []byte) error
      return err
    }
    {{ .ReceiverName }}.{{ .GoType }} = {{ varname .Name }}
-   {{ end }}{{ end }}
+   {{ end }}
+   {{ range .Properties }}
```
GlenDC: This logic is already defined here: https://github.com/Simon-Li/go-swagger/blob/d61294f01b4d2aac01024432a775db4bb5a22ff7/generator/templates/tupleserializer.gotmpl#L187-L203 the fix for this issue should figure out why that code isn't doing the job.

generator/templates/tupleserializer.gotmpl
```diff
@@ -241,7 +249,15 @@ func ({{.ReceiverName}} {{ pascalize .Name }}) MarshalJSON() ([]byte, error) {
      return nil, err
    }
    _parts = append(_parts, {{ varname .Name }})
-   {{ end }}{{ end }}
+   {{ end }}
+   {{ range .Properties }}
```
GlenDC: This logic is already defined here: https://github.com/Simon-Li/go-swagger/blob/d61294f01b4d2aac01024432a775db4bb5a22ff7/generator/templates/tupleserializer.gotmpl#L226-L243 the fix for this issue should figure out why that code isn't doing the job.

PR #852, opened 2017-01-09
validate spec prior to gen. client/server

### LINKED ISSUE #825: validate spec before generating, with opt-out option
There are many invalid specs submitted in bug reports. Perhaps it would be a good idea to always validate a spec before generating with a `--skip-validation` flag to allow restoring current behavior.

GlenDC:
this behaviour is opt-out, and can be skipped via the skip-validation flag

fixes #825


GlenDC:
One second, wait for the next commit, about to push a commit that aims at re-using the `loads.Document` loaded during validating, such that we don't have to load it twice.

GlenDC:
OK @casualjim if all CI tests pass, it should be ready for a final review by you!

GlenDC:
ok @casualjim sorry it took a bit longer, all should be fine now on my side.

GlenDC:
@vburenin rebased my work on top of latest master. Didn't change anything to the actual commits of this PR, just had to force push it to make the rebasing work cleanly. So if the checks still all pass, I think this is good to be merged in

PR #856, opened 2017-01-09
Generate code that keeps swagger spec as a string instead of a []byte

vburenin:
embeded_spec.go file contains swagger spec as an array of bytes that is not readable and not really diffable in case of the original swagger spec change. This PR addresses this by representing swagger spec as a string.


sflxn:
lgtm

vburenin:
@casualjim Are you around?

PR #878, opened 2017-01-21
Fix some grammar

PaulCapestany:


vendor/github.com/go-openapi/strfmt/time.go
```diff
@@ -81,7 +81,7 @@ func ParseDateTime(data string) (DateTime, error) {
 
 // DateTime is a time but it serializes to ISO8601 format with millis
 // It knows how to read 3 different variations of a RFC3339 date time.
-// Most API's we encounter want either millisecond or second precision times. This just tries to make it worry-free.
+// Most APIs we encounter want either millisecond or second precision times. This just tries to make it worry-free.
```
GlenDC: Could you please update this commit and reverting your change to the vendor folder, this is vendored code, and thus should not be modified manually. In this case you could fix that by opening a PR in the github.com/go-openapi/strfmt repository, after which we can update the vendored code and have your fix in.

Please simply amend that change to this commit and force push.

PaulCapestany: Oops, didn't notice this one, my bad!

PR #882, opened 2017-01-21
Remove unnecessary ToUpper from builder template

posener:


casualjim:
Thanks, can you run: `go generate ./generator`
Otherwise the template changes won't be picked up in the binary

posener:
@casualjim wow that was fast, Thanks!

PR #885, opened 2017-01-22
fix sliceserverheaderbuilder's broken ValueExpression selector

### LINKED ISSUE #881: Generating array in response-header fails
The following declaration validates fine but generates error while generating code:

```

              "responses": {
                    "200": {
                        "description": "ok",
                        "headers": {
                            "status": {
                                "type": "array",
                                "items": {
                                    "type": "string",
                                    "enum": [
                                        "NEW",
                                        "PUBLISHED",
                                        "OPEN",
                                        "FINISHED"
                                    ]
                                }
                            }
                        }
                    }
```

Error while genrating server:

```

template: serverResponses:31:48: executing "sliceserverheaderbuilder" at <.ValueExpression>: can't evaluate field ValueExpression in type generator.GenHeader
```

Client generates without errors.

GlenDC:
fixes #881


casualjim:
Could you add a test, and verify the generated code will do the right thing for items that look like:

```yaml
headers:
  status:
    type: array
    items:
      type: array
      items:
        type: array
        items:
          type: string
          enum:
          - NEW
          - PUBLISHED
          - OPEN
          - FINISHED
```

GlenDC:
@casualjim I added the test, and for now it seems to generate invalid code. I suppose you knew that would happen, hence the question.

Will look into a fix for this issue tomorrow and push it together with the updated test.

GlenDC:
@casualjim I replaced the original fix with the actual fix.

`ValueExpression` is now added to the `GenHeader`. Also added your test, and checked the generated code also manually. Seems to be be all fine AFAIK.

PR #897, opened 2017-01-24
Fixes #896 - Added *ast.StarExpr to parsers

### LINKED ISSUE #896: Embedded pointers cause errors in spec generation
## Problem statement

Similar to #563, but the fix in #564 only applied to `swagger:model`. This fixes the same problem with embedded pointers for `swagger:parameters` and `swagger:response`.

```
type params struct {
  Param1 string `json:"param1"`
}

type RealParams struct {
  *params
  Extra int `json:"extra"`
}
```

Results in:

```
unable to resolve embedded struct for: &{7807473 params}
```

## Environment
swagger version: 0.8.0
go version: 1.8.0rc2
OS: OSX 10.11.6

bjyoungblood:
Fixes #896

### DIFF
--- a/fixtures/goparsing/classification/operations/noparams.go
+++ b/fixtures/goparsing/classification/operations/noparams.go
@@ -33,6 +33,17 @@ type MyFileParams struct {
 	MyFormFile *bytes.Buffer `json:"myFormFile"`
 }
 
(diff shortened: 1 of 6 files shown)


GlenDC:
Thank you for your contribution @bjyoungblood, looks good! :)

PR #900, opened 2017-01-26
add custom server tutorial

### LINKED ISSUE #889: tutorial needed on how to generate a service using kvstore as an example
Currently it is not always clear to newcomers how one could generate a server/client, which contains just of the generated server, client and models code. That is to say, without generating the main file automatically, and keeping all the generated code under 1 directory. This approach means, you have full control over writing the actual CLI applications and handler code, where you want, without losing focus on what matters to you.

The need for this tutorial became clear of #888

GlenDC:
Fixes #889 

Note that I in the end opted for building a super simple server, as the kvstore had in my opinion way to much details that would distract the reader of this tutorial. The actual goal of this tutorial is to teach users new to go-swagger how to build a server, with a custom defined cli and hand-written handlers.

The tutorial does reference the kvstore example twice. Especially in the end we want to make clear that the server we created in this tutorial is very simple, and that the user might want to check out the kvstore example next to see a more complex example, built using the exact same techniques as showcases in this tutorial.

For now I've stored the end result of the tutorial as https://github.com/GlenDC/custom-swagger-server but in case you want it to be under the `go-swagger` namespace I don't mind transferring ownership.

### DIFF
--- a/docs/tutorial/custom-server.md
+++ b/docs/tutorial/custom-server.md
@@ -1,0 +1,314 @@
+# Custom Server Tutorial
+
+In this tutorial we'll be building up a custom server.
+The core will be generated using a manually written and maintained OpenAPI 2.0 spec.
+The cli code will be a thin layer around that, and will simply setup the API and server,
+using the parsed configurations and our hand written handlers.
+
+<!--more-->
+
+The server we'll building will be very simple.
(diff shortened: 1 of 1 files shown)


casualjim:
perhaps it would be good to just put it in the examples folder in this repo

PR #913, opened 2017-02-02
fix(generator): always use runtime/flagext in server generation

mitch000001:
We experienced a malformed server generation while using goswagger with gb.
There was an error because the code expected the go-swagger runtime flagext
package to be imported, which was not.

I regenerated the packaged templates using ./generator/gen-debug.sh.

To reproduce the error you can just use any swagger file like the petstore
example and use `gb` as build tool. After fetching all dependencies gb complains
about not being able to compile.

If there is still something missing or if you have any questions about
reproducing the error feel free to get back to me.


casualjim:
can you regenerate the embedded templates with `go generate ./generator` 
the gen-debug makes them refer to a path you our file system ` /Users/401208/src`

mitch000001:
Ah, yeah, sorry, will do.

PR #921, opened 2017-02-06
Remove no-op basePath assignment.

snargleplax:
The body of this `if` statement only reassigns the same value that was already assigned from; the value of `basePath` will still be the empty string afterwards, in all cases where the `if`-block actually executes.


GlenDC:
Thank you.

PR #926, opened 2017-02-09
Set the read and write timeout values for the http scheme

jbowes:
These configuration values were never set for vanilla http serving; only
their TLS counterparts were actually getting used.


jbowes:
> Looks good, can you please also regenerate the ./generator/bindata.go file?

Done :)

generator/templates/server/server.gotmpl
```diff
@@ -294,6 +294,8 @@ func (s *Server) Serve() (err error) {
 	if s.hasScheme(schemeHTTP) {
 		httpServer := &graceful.Server{Server: new(http.Server)}
 		httpServer.MaxHeaderBytes = int(s.MaxHeaderSize)
+		httpsServer.ReadTimeout = s.ReadTimeout
```
casualjim: this should be: `httpServer.ReadTimeout` and `httpServer.WriteTimeout`

jbowes: :( fixed

PR #933, opened 2017-02-16
Use json.RawMessage to decode polymorphic types

kars7e:
[][]byte worked fine until Go 1.7.4. In Go 1.7.5, decoding fails 
with following error:
"json: cannot unmarshal object into Go value of type []uint8"
Using json.RawMessage works correctly in this case.


casualjim:
can you run `go generate ./generator` so that the change gets picked up in the compiled binary?
it updates the ./generator/bindata.go file

kars7e:
Done!

PR #939, opened 2017-02-21
Added custom tag functionality

johnnadratowski:
⚠️ I probably butchered this and missed a bunch of places ⚠️ 

So, I discovered swagger last night, and this project today.  I am starting a new project, and I'd like to try to do something with this type of declarative API structure swagger gives you.  This project is above and beyond any other go integration, and seems to make some solid choices on the code generation 💯 

One of the things I like to do for my personal projects is make them as quick to get something done as possible, so I usually end up using an ORM to streamline the workflow.  The snag I hit was with the generated models: Most of the ORMs utilize custom struct tags for hints on what to do.  I did not see any way in the documentation or code to add custom tags to a struct field.

So, I built it.  Tested it.  It works in my local dev env for the simple use case I've tried.

🔍  Didn't add it to additional properties, I'm not sure if that should be done or what.
🔍  Didn't add it to the other `json:` tag fields in the other templates.  Maybe it should be there in some cases too?

⚠️ The current implementation does actually allow the user to create an invalid tag, as it just templates the entire thing as a string.  I could have used a different suffix for the extension name to prevent that. For example:

```
Model:
   properties:
      foo:
         type: string
         x-go-custom-tag-one: x,y,z
         x-go-custom-tag-two: a,b,c
```

would generate:

```
type Model struct {
   foo string `json:foo one:"x,y,z" two:"a,b,c"`
}
```

🤔 Maybe I misunderstood this all along and I'm not using these models the appropriate way.  It was still fun diving into the codebase. 👍


casualjim:
Excellent contribution :100:

johnnadratowski:
👍 NICE!!! I was not expecting this to be merged so quickly. Wonderful, I can continue with my project!

PR #942, opened 2017-02-22
Updates schema scanner to print warning

robertacosta:
While annotating my project, I discovered that a func type alias in our project was causing the swagger.json generation to quit execution.

Here is an example of the type alias in question:
`type action func(int) error`

The generator would print the following error:
`this is a bug, missing parser for a *ast.FuncType: &{Func:778790 Params:0xc4206a3440 Results:0xc4206a3500}`

At first I was quite confused what was causing this in my code and after much digging discovered the type alias above.

Perhaps rather than stopping execution all together when a model can't be parsed, go-swagger could print the model name that was skipped and continue execution.

Just for further context, I only ran into this when using the `-m` flag.


scan/schema.go
```diff
@@ -298,7 +298,8 @@ func (scp *schemaParser) parseDecl(definitions map[string]spec.Schema, decl *sch
 			}
 		}
 	default:
-		return fmt.Errorf("this is a bug, missing parser for a %T: %+v", tpe, tpe)
+		fmt.Printf("WARNING: Missing parser for a %T, skipping model: %s\n", tpe, decl.Name)
```
casualjim: can you make this use `log.Printf` or `fmt.Sprintf(os.Stderr, ...)`
the other things that get printed are also using log.

robertacosta:
Thank you for reviewing this so promptly.  I would love to improve the code coverage and was trying to add a test but was running into `unsupported type`.  Let me know if you are concerned.

PR #969, opened 2017-03-22
Bugfix: Check error of SetTimeout

kenjones-cisco:
The interface `github.com/go-openapi/runtime.ClientRequest` indicates
that `SetTimeout` can return an error. Adds a check in the client
parameter template for the error, and returns if not nil.

Identified by linter [errcheck](https://github.com/kisielk/errcheck) of generated code.

Signed-off-by: Kenny Jones <kenjones@cisco.com>


casualjim:
can you also run `go generate ./generator` and commit the generator/bindata.go file
Otherwise the changes won't be picked up by the binary

kenjones-cisco:
Looks like the tag `1.7.3` is no longer available at DockerHub, should I change it to `1.7.5` so that the circleci test will pass?

casualjim:
yes please

PR #1004, opened 2017-04-13
Task: Add Makefile and scripts

kenjones-cisco:
Adds Makefile and scripts for the common development flow.
After adding use of gometalinter, several linting issues were found
and resolved as part of the fix.


kenjones-cisco:
I'm not 100% happy with the Makefile and scripts as some parts do not work as I would have expected.
But with the number of linting items that I noticed and resolved after trying to determine how gotmetalinter was being used on the project, I figured it was best to submit the PR for feedback.

Worst case, I can revert any items.

casualjim:
this looks great, seen as I work for vmware i'll have to step up and also add vmware fusion/workstation based builds in the scripts :)

I've updated the circle config file on master so that it pulls in an image with an ssh client so that should make your build proceed

PR #1023, opened 2017-04-29
Add the list of supported consumers and producers

bobvanluijt:
Allows people to find the actually supported formats.

Updated after: https://github.com/go-swagger/go-swagger.github.io/pull/8#issuecomment-298176500


casualjim:
Thanks!

Issue #1005, opened 2017-04-14
How to serve 2 or more swagger specs from one server


msample:
## Problem statement

I want a go-swagger generated server to serve 2 swagger specs that have no overlap on paths.  One is a generic metadata api that is implemented by many services and the other is a service-specific api.   The built-in server.go template Server struct, by having exactly one API & handler, appears to prevent this.

Is there a good way to do this without re-templating or major modifications to go-swagger while staying compatible with the go-openapi runtime?

gregmarr:
Could you create a top-level swagger file that just includes the two lower-level ones?

casualjim:
you can also make your own main function and use the code from the generations of both with --skip-main

This allows for customization like using a different middleware stack, which in turn gives you the ability to serve 2 swagger specs at different paths.
And generally compose things a bit differently

msample:
Thanks.  Using a top level swagger file to merge two other files seems to require explicitly $ref'ing each path and data type entry?  ie I can't just generically union the paths: $ref:file1.json/paths $ref:file2.json/paths  and definitions: $ref:file1/definitions $ref:file2.json/definitions ?

There appears to be tooling to merge swagger files into a single file with only internal $refs. For example https://github.com/BigstickCarpet/swagger-parser has a 'bundle' function.

casualjim:
in this tool that is called `flatten`
it does exactly this and also moves anonymous objects from responses and request bodies to refs

msample:
Good point. I guess I'm looking for a flatten that takes 2 or more swagger files and merges them by ref or value. I'm trying to avoid writing the root one that ref's everything together.

casualjim:
I've written a go script that merges swagger docs from different locations in the past. I haven't spent any time thinking about how to make this a more general solution, so ideas and/or PR's are more than welcome :)

```go
package main

import (
	"encoding/json"
	"io/ioutil"
	"log"

	"github.com/go-openapi/loads"
	"github.com/go-openapi/loads/fmts"
)

// +build ignore

//NO TESTS

func main() {
	loads.AddLoader(fmts.YAMLMatcher, fmts.YAMLDoc)
	// CLI
	public, err := loads.Spec("../swagger/external.yml")
	if err != nil {
		log.Fatalln(err)
	}

	internal, err := loads.Spec("../swagger/internal.yml")
	if err != nil {
		log.Fatalln(err)
	}

	internalAPI := public.Pristine().Spec()
[...]

PR #1024, opened 2017-04-29
Add mixin feature that can merge swagger specs. Issue #1005

msample:
Also bumped vendored version of go-openapi/analysis with 'gvt update' for access to the new Mixin func.

### DIFF
--- a/cmd/swagger/commands/mixin.go
+++ b/cmd/swagger/commands/mixin.go
@@ -1,0 +1,97 @@
+package commands
+
+import (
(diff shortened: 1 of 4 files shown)


msample:
Thanks, Ivan. Do you need a PR for go-swagger.github.io to document the mixin command or should we let it 'settle' for a while?

casualjim:
the docs are also in this repo in the docs folder, yes please add those It helps people find out about the new functionality

PR #1030, opened 2017-05-05
add binary marshaler interface

casualjim:


jcoyne:
I'm curious why you implemented this interface.  Is it used somewhere?

casualjim:
yeah it's used when the handlers are returning errors and the content type is application/octet-stream

PR #1033, opened 2017-05-06
Add ability to use pre-generated model in server and client generation

nickrmc83:
Add ability to use pre-generated model in server and client generation. This is useful when, for example, you wish to share some helper code between clients and servers that validate and construct models. The new command-line flag provides the user the ability to specify an external go package that contains the required models. Setting "--existing-models" implies "--skip-models".

Usage:
swagger generate server --existing-models="github.com/\<repo\>\/models"
swagger generate client --existing-models="github.com/\<repo\>/models"


casualjim:
there are some conflicts here that need to be resolved. changes look good though

nickrmc83:
Thanks Ivan. I'll get the conflicts resolved early next week.

PR #1035, opened 2017-05-08
Adding ability to use pre-generated models.

nickrmc83:
Both a client and server can be generated and reference pre-generated
models using the new "existing-models" command-line switch e.g.:

swagger generate server --existing-models="github.com/<repo>/models"
swagger generate client --existing-models="github.com/<repo>/models"

This is a pr that replaces #1033 removing conflicts.


casualjim:
there is still a conflict for some reason in: cmd/swagger/commands/generate/server.go

casualjim:
thanks!

PR #1045, opened 2017-05-18
Fix 1040: make generation of api key authencator customizable.

dreamlover:
Signed-off-by: lfree xh <xh.dreamlover@gmail.com>


casualjim:
Thanks, would it be possible to do the same for BasicAuth and BearerAuth so that it's a consistent pattern?

dreamlover:
OK, BasicAuth and BearerAuth added.

casualjim:
The build is failing because of name collisions, those authenticators should either figure out unique names, but I think a reasonable work around is to call the properties: ApiKeyAuthenticator, BasicAuthenticator and BearerAuthenticator

casualjim:
thanks!

PR #1046, opened 2017-05-18
helper function for extending a default repositiry of templates

seifer:
Hi, I want to add some template to default registry but didn't find how to do it. This PR have to solve this issue. Is it a correct solution or we have some different way for this?


seifer:
@casualjim Up, please!!

PR #1048, opened 2017-05-20
Add easyjson:json annotation for generated types

bozaro:
Currect swagger code is support easyjson serialization, but don't mark generated types as `easyjson:json`.

This issue can partically workarounded by `easyjson -all`, but unfortunatelly this command ignore array types like:
```
// swagger:model Posts
type Posts []*Post
```

As result you can't use easyjson serialization with swagger model like:
```
swagger: '2.0'
paths:
  /posts/{thread_id}/profile:
    get:
      parameters:
      - name: thread_id
        in: path
        required: true
        type: number
      responses:
        200:
          schema:
            $ref: '#/definitions/Posts'
definitions:
  Post:
    type: object
    ...
  Posts:
    type: array
    items:
      $ref: '#/definitions/Post'
```

### DIFF
--- a/generator/templates/model.gotmpl
+++ b/generator/templates/model.gotmpl
@@ -7,6 +7,7 @@
       // {{ pascalize .Name }} {{ template "docstring" . }}
       {{- if not .IsBaseType }}
 // swagger:model {{ .Name }}
+// easyjson:json
       {{- else }}
 // swagger:discriminator {{ .Name }} {{ .DiscriminatorField }}
       {{- end }}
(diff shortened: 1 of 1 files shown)


casualjim:
HI thanks for this but the problem is more complicated 
Take a look at this PR: https://github.com/go-swagger/go-swagger/pull/557/files

It has the template changes required for easyjson support, unfortunately it has gone stale so rebasing that one is complicated. But it would be great if just the serialization changes (not the patch support and nullable/unset field stuff) could turn into a PR with just changes to serialization. It would fix about a dozen open bugs.

casualjim:
I've brought the changes from PR #557 up to date with current master in the easyjson branch
https://github.com/go-swagger/go-swagger/tree/easyjson

I won't be able to work on this for a while to come still. The AllOf functionality is still missing

casualjim:
stale

PR #1076, opened 2017-06-27
fix a shadow in server.gotmpl

arielt:
go vet reports shadowing in generated server.go
variable name changed in server.gotmpl

Signed-off-by: Ariel Tubaltsev <tubaltzev@gmail.com>


casualjim:
can you run `go generate ./generator` and push those changes otherwise the binary won't pick them up

arielt:
looks like circleci resources problem. let me recreate the request

PR #1081, opened 2017-06-29
Discriminator should also contain the base struct

alexmontecucco:
According to the swagger spec 2.0 (http://swagger.io/specification/#itemsObject), the discriminator "When used, the value MUST be the name of this schema or any schema that inherits it."

Also add safety check in template: buf2 should be created only if the Discriminates value is not empty + regenerate bindata.go. It is not mandatory now that by default the base struct is added, but it makes it safer.

A simple definition like the following would have generated not compiling code:
```
definitions:
  SomeStruct:
    type: object
    discriminator: SomeDiscField
    required:
      - SomeDiscField
    properties:
      SomeDiscField:
        type: string
```


alexmontecucco:
Hi, I just noticed there is a mistake in my proposal: I forgot that the base type is implemented as an interface.
Thus unmarshalling cannot work as such on the base type. Below is the result of my current proposal:
`type BaseType interface {
	runtime.Validatable

	// the base type name
	// Required: true
	Type() string
	SetType(string)
}

func unmarshalBaseType(data []byte, consumer runtime.Consumer) (BaseType, error) {
	buf := bytes.NewBuffer(data)
	buf2 := bytes.NewBuffer(data)

	// the first time this is read is to fetch the value of the type property.
	var getType struct {
		Type string `json:"type"`
	}
	if err := consumer.Consume(buf, &getType); err != nil {
		return nil, err
	}

	if err := validate.RequiredString("type", "body", getType.Type); err != nil {
		return nil, err
	}

	// The value of type is used to determine which type to create and unmarshal the data into
	switch getType.Type {
	case "BaseType":
		var result Geometry
		if err := consumer.Consume(buf2, &result); err != nil {
			return nil, err
		}
		return &result, nil

	case "SomeSubType":
		var result MultiPolygon
		if err := consumer.Consume(buf2, &result); err != nil {
			return nil, err
		}
		return &result, nil

	}
	return nil, errors.New(422, "invalid type value: %q", getType.Type)
}
`

These two lines won't work:
`var result Geometry
if err := consumer.Consume(buf2, &result); err != nil {`

I see two possibilities:

1. Define a base interface, containing a Get and Set Discriminator. The unmarshalBaseType function would return an instance of this interface.
2. Define a private struct and a public constructor used to unmarshall and create only. Thus, for the base type we would have to use the interface, to access the fields.

The problem is that we cannot really create both a public interface and a public struct because the current implementation (if I am correct) is already limiting when have struct with different cases.
Having for instance a struct named 'BaseStruct' and an interface named 'IBaseStruct' is not possible because the 'IBaseStruct' name could already be defined in the swagger file.

The first solution looks cleaner imo.

What do you think ?

casualjim:
I'm confused about this PR and problem statement.  The interface already expresses an is-a relationship.
Now if I repeat the requirement you have in my own words I think it summarizes as this:
I want a concrete implementation of the base type which can be instantiated in addition to the derived types.

So wouldn't an unexported struct be enough? because it would be able to capture the fields with getters and setters just like it does for the sub types but it wouldn't have any extra properties anyway so there is no need to expose the struct directly.

alexmontecucco:
Yes, I agree. I was just concerned about usage consistency between the different types. Base type an interface, others through an interface.
I can update the diff in this direction if you are ok ?

casualjim:
I'm fine with the change.

you can run the tests that are failing with ./hack/run-canary.sh

alexmontecucco:
Hi, 
I will for now push a simple tiny diff to avoid the compilation error because the diff approach with the private struct makes the template even more complex to understand. Mostly: 
- it adds a new condition in a few templates to have camelized names
- because the interface uses the exact name of the fields, the validate schema becomes "bloated" with both camelized and field name changes just for this PR's case.
- I tested this simple case where the object with the discriminator is itself a subtype, and it was not handled properly in the existing codebase.
So I feel, the full PR would have respected more the swagger spec but would introduce with the private struct approach more complexity and I cannot test all the cases.

nickrmc83:
This change introduces a non-compiling break where by the base type is now included as a switch case in the unmarshalXXX function and there's an attempt to return a pointer to the base interface. The below model and generation demonstrates the issue:
---
swagger: '2.0'
info:
  title: Test
  description: |
    Test
  version: "1.0.0"
definitions:
  vehicle:
    type: object
    discriminator: vehicle
    properties:
      id:
        description: Unique identifier of the vehicle
        type: string
      vehicle:
        description: The type of vehicle
        type: string
    required:
    - id
    - vehicle
  car:
    description: A car
    allOf:
    - $ref: '#/definitions/vehicle'
    - type: object
      properties:
        color:
          description: The color
          type: string
  bike:
    description: A bike
    allOf:
    - $ref: '#/definitions/vehicle'
    - type: object
      properties:
        brakes:
          description: Type of brakes
          type: string
---
swagger generate model -f <the-yaml-file-from-above>

PR #1087, opened 2017-07-02
Generate spec follows alias types

### LINKED ISSUE #649: swagger generate spec: type aliases in query parameters
I have code like this:

```
type Weekday int

type Parameters struct {
    //in:query
    Day Weekday
}
```

generate spec generates this json for parameter:

```
{
  "$ref": "#/definitions/Weekday",
  "x-go-name": "Day",
  "name": "Day",
  "in": "query"
},
```

what is invalid, because query parameter can have one of predefined datatypes according to spec.

Is it possible somehow to force go-swagger to use custom parameter type for spec generation?

Like:

```
type Parameters struct {
  //in:query
  //type:int
  //minimum:1
  //maximum:7
  Day Weekday
}
```

smacker:
Aliases in swagger.json present as native types.
Therefore all validation comments work as expected for them.

```go
type Int64 int64

type MyModel struct {
    // min: 10
    IntField Int64 `json:"intField"`
}
```

Also fixes #649

I use go-swagger only for json generation, please let me know if it breaks anything in code generator or anywhere else.


smacker:
oops. master merged

casualjim:
Perhaps something like this should get a marker annotation and be an additional feature instead of replacing what is there now for schemas. I don't know who's depending on the existing behavior.

For query/path/form params this is different though; there it would actually be the expected behavior.

smacker:
Do I understand right that it should be like:
```go
type Int64 int64

type MyModel struct {
    // swagger:alias
    // min: 10
    IntField Int64 `json:"intField"`
}
```
for any struct and behavior by default for structs annotated with `swagger:params`?

casualjim:
Yes I think so, that gives people a way to choose

smacker:
good. thanks for the clarification. I'll try to implement it later.

smacker:
Thank you sir!

casualjim:
yeah sorry I lost track of it. I try to follow a merge quickly policy

PR #1119, opened 2017-08-02
Correctly validate each element in array parameter

benpye:
The template was generating code that checked the members of the array, before actually assigning the array from the body of the request. This should now be validating the members of the array from the body, and then assigning.


casualjim:
can you run `go generate ./generator` otherwise the changes to the template won't be picked up

benpye:
Apologies, should be fixed now.

generator/templates/server/parameter.gotmpl
```diff
@@ -191,7 +191,7 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) BindRequest(r *http.Requ
     }
     {{ end }}
   {{ end }}} else {
-    {{ if .IsArray }}{{ if .Child }}{{ if (and (not .Schema.IsInterface) (or .Child.IsAliased .Child.IsComplexObject)) }}for _, {{ .IndexVar }}{{ .ReceiverName }} := range {{ .ReceiverName }}.{{ pascalize .Name }} {
+    {{ if .IsArray }}{{ if .Child }}{{ if (and (not .Schema.IsInterface) (or .Child.IsAliased .Child.IsComplexObject)) }}for _, {{ .IndexVar }}{{ .ReceiverName }} := range {{ if and (not .Schema.IsBaseType) .IsNullable }}&{{ end }}body {
```
casualjim: I think that just body would be enough, it won't hurt because when IsArray returns true IsNullable will always be false

so
```
{{ if and (not .Schema.IsBaseType) .IsNullable }}&{{ end }}
```

is not needed

benpye: Should be resolved now.

casualjim:
thanks! :100:

PR #1123, opened 2017-08-04
add support for strfmt.ObjectId

### LINKED ISSUE #884: Question: Defining bson ID response from mongoDB
Struggling with go-swagger and mongoDB:
*_id* returned from mongoDB using mgo driver is of type *bson.ObjectId*.

Example model:
```yaml
Message:
  type: object
  required:
    - text
  properties:
    id:
      type: string
      readOnly: true
    text:
      type: string
```

When I get *Messages* from DB by
```go
...
messages := []*models.Message{}
if err := c.Find(nil).All(&messages); err != nil {
  log.Print("Error on getting messages: ", err)
}
return message.NewGetMessageOK().WithPayload(messages)
```
The API response omitting the *id* property. Actually it's even stored as an empty string in mongoDB.
```
HTTP/1.1 200 OK
Content-Length: 774
Content-Type: application/json
Date: Sat, 21 Jan 2017 20:53:08 GMT

[
  {
    "text": "string"
  }
]
```
Is there any way to cast the *bson.ObjectId* to something that the generated server can work with or do I have to use the opposite approach by writing the server and generate the swagger spec afterwards?
And by the way, what effect do readOnly have since it's possible to POST that parameter against the API?

ultimateboy:
mgo compatibility was added to strfmt in https://github.com/go-openapi/strfmt/pull/12

This adds the`strfmt.ObjectId` type to the type resolver.

Signed-off-by: Matt Tucker <matt.tucker@microsoft.com>

Fixes #884

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,3 +267,4 @@ For a V1 I want to have this feature set completed:
   - [x] duration
   - [x] password
   - [x] custom string formats
+  - [x] BSON ObjectId
--- a/generator/typeresolver_test.go
+++ b/generator/typeresolver_test.go
@@ -59,6 +59,7 @@ var schTypeVals = []struct{ Type, Format, Expected string }{
 	{"string", "hexcolor", "strfmt.HexColor"},
 	{"string", "rgbcolor", "strfmt.RGBColor"},
 	{"string", "duration", "strfmt.Duration"},
+	{"string", "ObjectId", "strfmt.ObjectId"},
 	{"string", "password", "strfmt.Password"},
 	{"file", "", "runtime.File"},
 }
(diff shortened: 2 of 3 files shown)


ultimateboy:
This seems like the appropriate way to add support for mongodb ObjectId's. I've tested with a swagger spec that includes:
```yaml
properties:
  _id:
    type: string
    format: ObjectId
```

This generates a struct with:
```
type Test struct {
	ID strfmt.ObjectId `json:"_id,omitempty"`
}
```

casualjim:
yes the PR looks good, waiting for CI to agree

Issue #961, opened 2017-03-17
Validation for extended string format


RomanBuzuk:
## Problem statement

We can define that some property has extended string format ( for example ipv4). Then go-swagger generate this property with special type ( for ipv4 it is *strfmt.IPv4). But generated code don`t do validation for this type.
## Example:
swagger.yaml 
```
definitions:
  ethernet:
    type: object
    required:
      - ip
    properties:
      id:
        type: integer
        format: int32
        readOnly: true
      ip:
        type: string
        format: ipv4
```
We get file ethernet.go with code 
```
...
type Ethernet struct {
	// id
	// Read Only: true
	ID int32 `json:"id,omitempty"`

	// ip
	// Required: true
	IP *strfmt.IPv4 `json:"ip"`
}
...
func (m *Ethernet) validateIP(formats strfmt.Registry) error {

	if err := validate.Required("ip", "body", m.IP); err != nil {
		return err
	}
	return nil
}
```
But why  it is not generated like this : 
```
func (m *Ethernet) validateIP(formats strfmt.Registry) error {

	if err := validate.Required("ip", "body", m.IP); err != nil {
		return err
	}
// check that we get string in right format
	if err := validate.FormatOf("ip", "body", "ipv4", m.IP.String(), formats); err != nil {
		return err
	}
	return nil
}
```

## Environment
swagger version: 0.8.0

benpye:
Did you ever work out what was going wrong here?

casualjim:
haven't been able to look at it for cause of the issue. I assume it's missing an opportunity to include the format validator

fredbi:
Duplicate of #909

PR #1124, opened 2017-08-04
Ensure extend string formats are validated

benpye:
This should fix issue #961 , it allows for the custom format validator to be emitted even for types backed by a primitive type (ie. strfmt.IPv4)


benpye:
Rebased on master and fixed a typo in the commit message.

PR #1128, opened 2017-08-11
Fixing Un/MarshalJSON calls for polymorphic types

nickrmc83:
Errors were visible when defining polymorphic types that had one or more
fields of polymorphic arrays.

All tests pass.

This should be carefully code reviewed as I've made some changes to how marshaling code works. Additionally it would be good to validate whether marking an array of polymorphic types as a base type and non-exportable is the correct approach or whether setting something like HasDiscriminator is better.


casualjim:
I forgot to review this over the weekend. Would it be possible to update? I'll try to do it tonight

nickrmc83:
I've got some more changes as I found a few errors in my submission so
don't worry about review ingress quite yet.

nickrmc83:
@casualjim this is now ready to be reviewed. All unit tests pass and I've done some validation with my more complex internal models.

Thanks

casualjim:
Could you update to latest master. I verified it works but I don't want to lose attribution

nickrmc83:
Done! Let me know if there's anything else for me to do or that I can help with.

casualjim:
Thanks, this was a non-trivial piece of work :100:

PR #1148, opened 2017-08-22
Adds a Copyright comment header to auto generated code files.

### LINKED ISSUE #307: Flag to add a licence/copyright header on generated files
It would be useful to have a flag in the CLI which accepts a copyright or license header file. This file could then be appended as a header to all the generated files.

EleanorRigby:
- Adds -r option to specify copyright file from where copyright comments are copied by go-swagger in plain text format.

- The  copyright text is added to a /* */ block in each auto-generated go file.

- Copyright must not contain any comment markers for golang (like */). It will create erroneous code file of form 
`/*
Copyright Comment
*/ 
*/`

- The copyright -r option is optional. 

fixes #307


generator/templates/swagger_json_embed.gotmpl
```diff
@@ -1,5 +1,9 @@
 // Code generated by go-swagger; DO NOT EDIT.
 
+
+{{ if .Copyright }}/*{{"\n"}}{{.Copyright}}*/{{ end }}
```
casualjim: you can do this:

```
{{- if .Copyright }}{{ comment .Copyright }}{ end -}}
```

EleanorRigby: I am not sure if comment action works. it is giving me this error when I add comment action

`2017/08/23 11:19:22 Failed to load template serverMain: template: serverMain:127: unexpected EOF`

EleanorRigby:
#307

PR #1164, opened 2017-09-08
Test model generator

joshq00:
Adds test for model generation via command


joshq00:
Created these (passing) tests as a preface to solving an issue -- will submit that in a separate PR

Issue #1172, opened 2017-09-15
Composite Model Generated Structs have changed


kkarabian:
This just started happening with 0.12.0 I believe.

## Problem statement

Generated models structure has changed for composites from:
```
type ReportScore struct {
    ScoreData

    LastUpdateTime *string `json:"lastUpdateTime"`
}
```
to
```
type ReportScore struct {
    ScoreData

    ReportScoreAllOf1
}
```
for the following specification:

## Swagger specification
```
  ReportScore:
    allOf:
      - $ref: '#/definitions/ScoreData'
      - type: object
        properties:
          lastUpdateTime:
            type: string
        required: [lastUpdateTime]
```
## Steps to reproduce
swagger generate client -t swaggen -f apispec.yaml

## Environment
swagger version: 0.12.0 
go version: 1.8.3
OS:   Mac OS 10.10.5

casualjim:
if you build the swagger command from master you can generate your client with a `--skip-flatten` flag and that should restore the previous behavior

PR #1173, opened 2017-09-15
Add skip-flatten flag to generate server, client and operation

EleanorRigby:
Adds `--skip-flatten` flag to `swagger   ##generate` command for skipping the flattening of spec before generation

#1172


EleanorRigby:
@casualjim : Please check if I referenced correct issue here before merging.

PR #1175, opened 2017-09-17
Adds Text Marshal/Unmarshal to model objects. Add String() interface

EleanorRigby:


generator/templates/schema.gotmpl
```diff
@@ -185,5 +185,40 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalBinary(b []byte) error
...
+  fmt.Printf("%s", string(text))
+  *{{ .ReceiverName }} = res
+  return nil;
+}
+
+
+// String interface implementation
+func ({{.ReceiverName}} *{{ pascalize .Name }}) String() string {
+  if {{ .ReceiverName }} == nil {
+    fmt.Errorf("Nil calling object")
```
casualjim: this is not used so it should be removed.

generator/templates/schema.gotmpl
```diff
@@ -185,5 +185,40 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalBinary(b []byte) error
...
+    return ""
+  }
+  b, err := swag.WriteJSON({{ .ReceiverName }})
+  if err != nil {
+    fmt.Errorf("%v", err)
```
casualjim: this is not used so it should be removed

generator/templates/schema.gotmpl
```diff
@@ -185,5 +185,40 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalBinary(b []byte) error
...
+// MarshalText interface implementation
+func ({{.ReceiverName}} *{{ pascalize .Name }}) MarshalText() ([]byte, error) {
+  if {{ .ReceiverName }} == nil {
+    return nil, nil
+  }
+  b, err := swag.WriteJSON({{ .ReceiverName }})
+  if err != nil {
+    return nil, err
+  }
+  return []byte(fmt.Sprintf("%v", b)), nil
```
casualjim: b is already a byte slice no need to fmt.Sprintf
This can just be return swag.WriteJSON

generator/templates/schema.gotmpl
```diff
@@ -185,5 +185,40 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalBinary(b []byte) error
...
+}
+
+// UnmarshalText interface implementation
+func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalText(text []byte) error {
+  var res {{ pascalize .Name }}
+  fmt.Printf("%s", string(text))
```
casualjim: this is not used, should use swag.ReadJSON

As a general rule, because I think you are trying to print to the log stream, fmt.Print  should not be used.

generator/templates/schema.gotmpl
EleanorRigby: As far as I recall , you asked me to implement string() function.

So, I remove both UnmarshalText and string()

generator/templates/schema.gotmpl
EleanorRigby: fixed

generator/templates/schema.gotmpl
EleanorRigby: fixed

generator/templates/schema.gotmpl
```diff
@@ -185,5 +185,32 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalBinary(b []byte) error
...
+    return nil, nil
+  }
+  return swag.WriteJSON({{ .ReceiverName }})
+
+}
+
+// UnmarshalText interface implementation
+func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalText(text []byte) error {
+  var res {{ pascalize .Name }}
+  *{{ .ReceiverName }} = res
```
casualjim: this needs to use swag.ReadJSON so it's symmetrical with the MarshalText

generator/templates/schema.gotmpl
```diff
@@ -185,5 +185,32 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}) UnmarshalBinary(b []byte) error
...
+  return nil;
+}
+
+
+// String interface implementation
+func ({{.ReceiverName}} *{{ pascalize .Name }}) String() string {
+  b, err := swag.WriteJSON({{ .ReceiverName }})
+  if err != nil {
+    fmt.Errorf("%v", err)
```
casualjim: this fmt.Errorf needs to be removed

PR #1189, opened 2017-10-03
Added x-omitempty extension to fix #774

### LINKED ISSUE #774: x-nullable for array values doesn't have any effect
## Problem statement

If an array/slice value is `nil` and the aim is to exclude it from the encoded JSON there doesn't seem to be a way to define that. Is this intended behaviour or should `x-nullable` for a property containing an array or `nil` also result in an `omitempty` annotation within the model spec? If it is intentional, is there any other way to exclude null array values from the encoded response?

In previous version null arrays were excluded but to allow `nil` and empty arrays to be distinguished, improvements got introduced in #707.

In the example given below, the `books` array of a store object is always included in the JSON.

## Swagger specification
```yaml
swagger: "2.0"
info:
  title: Playground
  description: Swagger Playground
  version: 1.0.0
basePath: /v1
schemes:
- http
consumes:
- application/vdn.playground.v1+json
produces:
- application/vdn.playground.v1+json

paths:
  /stores:
    get:
      summary: Get a list of stores
      tags:
        - store
      responses:
        200:
          description: List of stores
          schema:
            type: array
            items:
              $ref: "#/definitions/Store"
        default:
          description: generic error response
          schema:
            $ref: "#/definitions/Error"

definitions:
  # Store
  Store:
    type: object
    properties:
      id:
        type: integer
        format: int64
        example: 123
      name:
        type: string
        example: Daunt Books
      books:
        type: array
        items:
          $ref: "#/definitions/Book"
        x-nullable: true

  # Book
  Book:
    type: object
    properties:
      id:
        type: integer
        format: int64
        example: 123
      title:
        type: string
        example: Winnie the Pooh

  # Error
  Error:
    type: object
    readOnly: true
    properties:
      code:
        type: integer
        format: int64
        example: 400
      message:
        type: string
        example: Unexpected error
    required:
      - message

```

## Steps to reproduce

```go
api.StoreGetStoresHandler = store.GetStoresHandlerFunc(func(params store.GetStoresParams) middleware.Responder {
  res := []*models.Store{&models.Store{
    ID:    1,
    Name:  "My Book Store",
  }} // note the Books attribute is not set
  return store.NewGetStoresOK().WithPayload(res)
})
```

## Expected Result
```json
[{
  "id":1,
  "name":"My Book Store"
}]
```

## Actual Result
```json
[{
  "id":1,
  "name":"My Book Store",
  "books":null
}]
```

> Note: Version used for this 0.7.4

akhobov:
The new vendor extension works only with array types and controls
presence of the 'omitempty' struct tag that is used by JSON Marshaler


casualjim:
could you regenerate bindata and push again? I have a bindata conflict

PR #1194, opened 2017-10-05
Add StatusCode() and GetPayload() methods to response

imikushin:
We need an easy way to get the response status code and payload.


casualjim:
discussed offline

PR #1200, opened 2017-10-10
Only write body to request when the body is not nil

### LINKED ISSUE #1199: Body is written to request even when it's empty
## Problem statement

Any swagger definition that happen to have an array payload as a request will end up writing a body that contains the field with the value `null` (in case of a complex type such as `[]*MyType`).

Golang Struct:
```go
// MoveClustersRequest Configuration for moving clusters off an allocator
// swagger:model MoveClustersRequest
type MoveClustersRequest struct {

	// Optional list of Elasticsearch clusters to move off the allocator. If not specified, then all Elasticsearch clusters on the allocator are moved.
	ElasticsearchClusters []*MoveElasticsearchClusterConfiguration `json:"elasticsearch_clusters"`

	// Optional list of Kibana clusters to move off the allocator. If not specified, then all Kibana clusters on the allocator are moved.
	KibanaClusters []*MoveKibanaClusterConfiguration `json:"kibana_clusters"`
}
```

Because the fields are optional and a list they are not marked with the `omitempty` struct tag, which makes the request payload to be not empty:

```json
{
    "elasticsearch_clusters": null,
    "kibana_clusters":null
}
```

The problem is that the autogenerated Go code that results for the `<action>_params.go` writes a nullable type to the request body even when the body is clearly empty. I find this a bit inconsistent of how you would expect this to work, Here's the autogenerated Go code:

```go
// WriteToRequest writes these params to a swagger request
func (o *MoveClustersParams) WriteToRequest(r runtime.ClientRequest, reg strfmt.Registry) error {

	if o.Body == nil {
		o.Body = new(v1models.MoveClustersRequest)
	}

	if err := r.SetBodyParam(o.Body); err != nil {
		return err
	}
}
```

## Swagger specification

```json
{
    "swagger": "2.0",
    "host": "localhost",
    "info": {
        "description": "",
        "version": "",
        "title": "",
        "termsOfService": ""
    },
    "basePath": "/api/v1/",
    "paths": {
        "/platform/infrastructure/allocators/{allocator_id}/clusters/_move": {
            "post": {
                "summary": "Move clusters",
                "description": "Moves clusters.",
                "operationId": "move-clusters",
                "parameters": [
                    {
                        "in": "body",
                        "name": "body",
                        "description": "Overrides defaults for the move of each cluster",
                        "required": false,
                        "schema": {
                            "$ref": "#/definitions/MoveClustersRequest"
                        }
                    },
                    {
                        "name": "allocator_id",
                        "in": "path",
                        "description": "The identifier of the allocator of which to move clusters.",
                        "required": true,
                        "type": "string"
                    }
                ],
                "responses": {
                    "202": {
                        "description": "response"
                    }
                }
            }
        }
    },
    "definitions": {
        "MoveElasticsearchClusterConfiguration": {
            "type": "object",
            "required": [
                "cluster_ids"
            ],
            "properties": {
                "cluster_ids": {
                    "type": "array",
                    "description": "Identifiers for the Elasticsearch clusters.",
                    "items": {
                        "type": "string"
                    }
                }
            },
            "description": "Configuration object for moving Elasticsearch clusters."
        },
        "MoveKibanaClusterConfiguration": {
            "type": "object",
            "required": [
                "cluster_ids"
            ],
            "properties": {
                "cluster_ids": {
                    "type": "array",
                    "description": "Identifiers for the Kibana clusters.",
                    "items": {
                        "type": "string"
                    }
                }
            },
            "description": "Configuration object for moving Kibana clusters."
        },
        "MoveClustersRequest": {
            "type": "object",
            "properties": {
                "elasticsearch_clusters": {
                    "type": "array",
                    "description": "Optional list of Elasticsearch clusters to move off the allocator. If not specified, then all Elasticsearch clusters on the allocator are moved.",
                    "items": {
                        "$ref": "#/definitions/MoveElasticsearchClusterConfiguration"
                    }
                },
                "kibana_clusters": {
                    "type": "array",
                    "description": "Optional list of Kibana clusters to move off the allocator. If not specified, then all Kibana clusters on the allocator are moved.",
                    "items": {
                        "$ref": "#/definitions/MoveKibanaClusterConfiguration"
                    }
                }
            },
            "description": "Configuration for moving clusters off an allocator"
        }
    }
}
```

## Steps to reproduce

1. Copy the swagger spec and save it as `nonEmptyBody.json`.
2. Issue `swagger generate client -f nonEmptyBody.json` to generate the client code.
3. open `client/operations/move_clusters_parameters.go` and check `WriteToRequest` method.

You should see the problematic snippet in the `WriteToRequest` method

```go
if o.Body == nil {
	o.Body = new(models.MoveClustersRequest)
}

if err := r.SetBodyParam(o.Body); err != nil {
	return err
}
``` 

## Environment
swagger version: 2.0.0 
go version: 1.9.0
OS:  macOS Sierra

marclop:
## Description

The body won't be written to the request if it's empty.

Closes #1199


casualjim:
sorry for the long wait, I'm in the middle of moving homes.

PR #1219, opened 2017-10-18
oauth2 authentication example

huikang:
- use google's oauth2 server as example

Signed-off-by: Hui Kang <kangh@us.ibm.com>


huikang:
Not sure what @houndci-bot does not like the format? I see other example has similar format.

casualjim:
it's the blank lines

PR #1223, opened 2017-10-20
handle model property conflicts when format is "date" or "datetime"

### LINKED ISSUE #1220: Model file cannot compile when a string field has both date format and maxLength
## Problem statement

When the swagger file has a string field with both date format and maxLength, the generated model file does not compile.

In this example (adapted from [petstore](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/json/petstore-simple.json), the field "birth" is string, has date format, and a max length of 10.

We should call m.Birth.String() rather than string(m.Birth) in new_pet.go.

## Swagger specification

```json
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.0",
    "title": "Swagger Petstore"
  },
  "host": "petstore.swagger.io",
  "basePath": "/api",
  "schemes": [
    "http"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/pets": {
      "post": {
        "operationId": "addPet",
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "pet",
            "in": "body",
            "description": "Pet to add to the store",
            "required": true,
            "schema": {
              "$ref": "#/definitions/NewPet"
            }
          }
        ],
        "responses": {
          "200": {
            "description": "pet response",
            "schema": {
            }
          },
          "default": {
            "description": "unexpected error",
            "schema": {
            }
          }
        }
      }
    }
  },
  "definitions": {
    "NewPet": {
      "type": "object",
      "required": [
        "name"
      ],
      "properties": {
        "name": {
          "type": "string"
        },
        "birth": {
          "type": "string",
          "format": "date",
          "maxLength": 10
        }
      }
    }
  }
}
```

## Steps to reproduce

```
$ swagger generate client --spec apis/test.json
$ go build client/swagger_petstore_client.go 
# github.com/neighborly/experiments/pershing-demo/models
models/new_pet.go:60:54: cannot convert m.Birth (type strfmt.Date) to type string
```

## Environment
swagger version: dev
go version: 1.9.1
OS: OSX

kreamyx:
fixes #1220

### DIFF
--- a/generator/model.go
+++ b/generator/model.go
@@ -546,9 +546,10 @@ func hasValidations(model *spec.Schema, isRequired bool) (needsValidation bool, 
 // handleFormatConflicts handles all conflicting model properties when a format is set
 func handleFormatConflicts(model *spec.Schema) {
 	switch model.Format {
-	case "date", "datetime", "uuid":
+	case "date", "datetime", "uuid", "bsonobjectid", "base64", "duration":
 		model.MinLength = nil
 		model.MaxLength = nil
+		model.Pattern = ""
 		// more cases should be inserted here if they arise
 	}
 }
--- a/generator/model.go
+++ b/generator/model.go
@@ -546,10 +546,7 @@ func hasValidations(model *spec.Schema, isRequired bool) (needsValidation bool, 
 // handleFormatConflicts handles all conflicting model properties when a format is set
 func handleFormatConflicts(model *spec.Schema) {
 	switch model.Format {
-	case "date":
-		model.MinLength = nil
-		model.MaxLength = nil
-	case "datetime":
+	case "date", "datetime", "uuid":
 		model.MinLength = nil
 		model.MaxLength = nil
 		// more cases should be inserted here if they arise
@@ -558,7 +555,7 @@ func handleFormatConflicts(model *spec.Schema) {
 
 func (sg *schemaGenContext) schemaValidations() sharedValidations {
 	model := sg.Schema
-	// resolve any conflicting properties if the model has a Fromat
+	// resolve any conflicting properties if the model has a format
 	handleFormatConflicts(&model)
 
 	isRequired := sg.Required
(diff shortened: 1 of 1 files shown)


generator/model.go
```diff
@@ -543,8 +543,23 @@ func hasValidations(model *spec.Schema, isRequired bool) (needsValidation bool,
...
+	case "datetime":
+		model.MinLength = nil
+		model.MaxLength = nil
+		// more cases should be inserted here if they arise
+	}
+}
+
 func (sg *schemaGenContext) schemaValidations() sharedValidations {
 	model := sg.Schema
+	// resolve any conflicting properties if the model has a Fromat
```
gregmarr: typo: Fromat

generator/model.go
```diff
@@ -543,8 +543,23 @@ func hasValidations(model *spec.Schema, isRequired bool) (needsValidation bool,
 	return
 }
 
+// handleFormatConflicts handles all conflicting model properties when a format is set
+func handleFormatConflicts(model *spec.Schema) {
+	switch model.Format {
+	case "date":
+		model.MinLength = nil
+		model.MaxLength = nil
+	case "datetime":
```
gregmarr: Since "date" and "datetime" have the same conflicts, you can combine the cases:

    case "date", "datetime":

gregmarr: Probably "uuid" as well.

generator/model.go
```diff
@@ -543,8 +543,20 @@ func hasValidations(model *spec.Schema, isRequired bool) (needsValidation bool,
 	return
 }
 
+// handleFormatConflicts handles all conflicting model properties when a format is set
+func handleFormatConflicts(model *spec.Schema) {
+	switch model.Format {
+	case "date", "datetime", "uuid":
```
casualjim: also: bsonobjectid, base64 and duration

none of those are actually strings so the validations don't apply to them

PR #1249, opened 2017-11-06
Remove Content-Type for empty responses

databus23:
This is an attempt at fixing https://github.com/go-openapi/runtime/issues/64.

Some clients trip over responses with empty bodies and `Content-Type: application/json` as they feed the body directly to a JSON parser. As a zero length string is not a valid JSON this causes a parse error.
As a prominent client jQuery barfs on this intentionally since 1.9: https://jquery.com/upgrade-guide/1.9/#jquery-ajax-returning-a-json-result-of-an-empty-string :(

Note that at the moment this change will cause empty responses to have `Content-Type: text/plain` set because its what go's http server does by default: https://github.com/golang/go/issues/20784.
Hopefully this will change in later go version.
Nonetheless this fixes pedantic clients as an empty string is valid `text/plain`


databus23:
Just adding that https://github.com/golang/go/issues/20784 was closed so the spurious `text/plain` content-type should be gone with go 1.10

pheepi:
Unfortunately, this change causes unexpected behavior when the content type is previously set by `WithContentType` or `SetContentType` methods. The generated `WriteResponse` method might look like this:

```
// WriteResponse to the client
func (o *HeadOK) WriteResponse(rw http.ResponseWriter, producer runtime.Producer) {

	// response header Content-Type

	contentType := o.ContentType
	if contentType != "" {
		rw.Header().Set("Content-Type", contentType)
	}

	rw.Header().Del(runtime.HeaderContentType) //Remove Content-Type on empty responses

	rw.WriteHeader(200)
}
```

The content type is deleted even if it is wanted. I do not agree that empty response should not carry Content-type header from https://github.com/go-openapi/runtime/issues/64, at least in case of HEAD requests, because it is one of often function.

vincentvanderweele:
Agree with @pheepi. I'm running in the issue now that the `Content-Type` header is stripped from my HEAD responses and there is no way to work around this.

As far as I know, `WriteResponse` is scoped to a model, not an operation, so it cannot decide if this header should be there or not.

Should this change be reverted and the burden moved to the application developer to just not set the `Content-Type` header if it is undesired?

PR #1257, opened 2017-11-10
check for child isinterface

e-zhang:
Here's a minimal example of a parameter that takes in array of objects
```
swagger: '2.0'
info:
  title: Test
  version: '1'
paths:
  /test:
    post:
      parameters:
        - name: test
          in: body
          schema:
            type: array
            items:
              type: object
      responses:
        200:
          description: success
```

This generates the following segment in the restapi/operations dir
```
var body []interface{}
...
for _, io := range body {
   if err := io.Validate(route.Formats); err != nil {
      res = append(res, err)
      break
  }
}

if len(res) == 0 {
  o.Test = body
}
```

This causes a compile error because `io.Validate` is not valid on `interface{}`.

I *think* we should be checking for `.Child.IsInterface` instead of `.Schema.IsInterface`. Given that `.IsArray` is already `true`, I'm not sure if `.Schema.IsInterface` is the right check. 
Please correct me if I misinterpreted this section. 
My expectation is that in the case of `Items []interface{}` the `for` loop should not be generated.


casualjim:
can you run `go generate ./generator` and push the changes. Otherwise your change won't be picked up

e-zhang:
Will do. Checked in the generated bindata.go

Issue #1263, opened 2017-11-12
Command line option to specify copyright does not work in long mode


fredbi:
## Problem statement
swagger -r : ok
swagger --copyright-definition-file: ko

Culprit: ./cmd/swagger/commands/generate/shared.go

Also: option is not documented/ I propose --copyright-file={file}

Preparing P.R to fix

## Swagger specification
N/A
## Steps to reproduce
Self explanatory
## Environment
swagger version: 0.13.0
go version: 1.9
OS:   ubuntu 14.04

fredbi:
Merge tested OK.

PR #1264, opened 2017-11-12
Fix issue#1263

fredbi:
Small issues, small fixes... Moving forward and learning.


fredbi:
I checked it out from master. #1263 now closed.

PR #1273, opened 2017-11-17
fix #1272 - Improve error handling in generator

### LINKED ISSUE #1272: improve error handling in generator
## Problem statement
Playing around with template generation, I found out that error messages are often unclear.
Besides, errors in templates are hard to figure out since formatting errors do not dump any generated code.

I submit a PR with the following (small) improvements:
- errors are returned with some enrichment and more explicit errors
- errors in formatting after a successful template execution are reported, the faulty generated source is dumped for inspection and the process stops here
- explicit empty return values on errors
- panic on errors in functions which do not return an error part
- panic recovery in main, with clean exit

Hope this helps.

## Swagger specification
N/A

## Steps to reproduce
swagger generate server ... --template-dir=~/myowntemplates
...

## Environment
swagger version: 0.13.0-1.gb015bda
go version: 1.9
OS:   linux ubtuntu 14.04

fredbi:
#1272 

Improve error handling while generating templates.

### DIFF
--- a/cmd/swagger/swagger.go
+++ b/cmd/swagger/swagger.go
@@ -22,7 +22,6 @@ import (
 	"github.com/go-openapi/loads/fmts"
 	"github.com/go-swagger/go-swagger/cmd/swagger/commands"
 	"github.com/jessevdk/go-flags"
-	//"runtime/debug"
 )
 
 func init() {
@@ -42,6 +41,7 @@ func main() {
 	// TODO: reactivate 'defer catch all' once product is stable
 	// Recovering from internal panics
 	// Stack may be printed in Debug mode
+	// Need import "runtime/debug".
 	//defer func() {
 	//	r := recover()
 	//	if r != nil {
--- a/cmd/swagger/swagger.go
+++ b/cmd/swagger/swagger.go
@@ -22,25 +22,36 @@ import (
 	"github.com/go-openapi/loads/fmts"
 	"github.com/go-swagger/go-swagger/cmd/swagger/commands"
 	"github.com/jessevdk/go-flags"
+	//"runtime/debug"
 )
 
 func init() {
 	loads.AddLoader(fmts.YAMLMatcher, fmts.YAMLDoc)
 }
 
+var (
+	// Debug is true when the SWAGGER_DEBUG env var is not empty
+	Debug = os.Getenv("SWAGGER_DEBUG") != ""
+)
+
 var opts struct {
 	// Version bool `long:"version" short:"v" description:"print the version of the command"`
 }
 
 func main() {
+	// TODO: reactivate 'defer catch all' once product is stable
 	// Recovering from internal panics
-	defer func() {
-		r := recover()
-		if r != nil {
-			log.Fatal("Fatal error:", r)
-			os.Exit(1)
-		}
-	}()
+	// Stack may be printed in Debug mode
+	//defer func() {
+	//	r := recover()
+	//	if r != nil {
+	//		log.Printf("Fatal error:", r)
+	//		if Debug {
+	//			debug.PrintStack()
+	//		}
+	//		os.Exit(1)
+	//	}
+	//}()
 
 	parser := flags.NewParser(&opts, flags.Default)
 	parser.ShortDescription = "helps you keep your API well described"
--- a/cmd/swagger/swagger.go
+++ b/cmd/swagger/swagger.go
@@ -33,6 +33,15 @@ var opts struct {
 }
 
 func main() {
+	// Recovering from internal panics
+	defer func() {
+		r := recover()
+		if r != nil {
+			log.Fatal("Fatal error:", r)
+			os.Exit(1)
+		}
+	}()
+
 	parser := flags.NewParser(&opts, flags.Default)
 	parser.ShortDescription = "helps you keep your API well described"
 	parser.LongDescription = `
--- a/generator/model.go
+++ b/generator/model.go
@@ -74,7 +74,7 @@ func GenerateDefinition(modelNames []string, opts *GenOpts) error {
 		// lookup schema
 		model, ok := specDoc.Spec().Definitions[modelName]
 		if !ok {
-			return fmt.Errorf("model %q not found in definitions in %s", modelName, specPath)
+			return fmt.Errorf("model %q not found in definitions given by %q", modelName, specPath)
 		}
 
 		// generate files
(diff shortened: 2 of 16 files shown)


generator/model.go
```diff
@@ -106,7 +106,7 @@ func (m *definitionGenerator) Generate() error {
 
 	mod, err := makeGenDefinition(m.Name, m.Target, m.Model, m.SpecDoc, m.opts)
 	if err != nil {
-		return err
+		return fmt.Errorf("could not generate definitions for model %s on target %s: %s", m.Name, m.Target, err.Error())
```
casualjim: instead of %s, err.Error() you can use %v, err

generator/operation_test.go
```diff
@@ -581,7 +581,7 @@ func TestGenServerIssue890_ValidationTrueFlatteningTrue(t *testing.T) {
 func TestGenClientIssue890_ValidationTrueFlatteningTrue(t *testing.T) {
 	defer func() {
 		dr, _ := os.Getwd()
-		os.RemoveAll(dr+"/restapi/")
+		os.RemoveAll(dr + "/restapi/")
```
casualjim: we should probably use `os.RemoveAll(filepath.Join(filepath.FromSlash(dr), "restapi"))` here

generator/shared.go
```diff
@@ -339,39 +339,47 @@ type GenOpts struct {
 	Copyright         string
 }
 
-// TargetPath returns the target path relative to the server package
+// TargetPath returns the target generation path relative to the server package
+// May be used in templates. Errors are not fatal: an empty string is returned instead
 func (g *GenOpts) TargetPath() string {
```
casualjim: Why are errors not fatal?

generator/shared.go
```diff
@@ -339,39 +339,47 @@ type GenOpts struct {
...
-		log.Fatalln(err)
+		log.Printf("Target path \"%s\" and server path \"%s\" are not related. You shouldn't specify an absolute path in --server-package: %s", g.Target, g.ServerPackage, err.Error())
+		return ""
 	}
 	return tgtRel
 }
 
 // SpecPath returns the path to the spec relative to the server package
+// May be used in templates. Errors are not fatal: an empty string is returned instead
 func (g *GenOpts) SpecPath() string {
```
casualjim: why are errors not fatal, this seems like a fairly critical piece of info

generator/support.go
```diff
@@ -593,6 +595,9 @@ func (a *appGenerator) makeCodegenApp() (GenApp, error) {
 
 	log.Println("planning definitions")
 	for mn, m := range a.Models {
+		if Debug {
+			fmt.Printf("makeGenDefinition(\"%s\")\n", mn)
```
casualjim: I think you mean `log.Printf` here

cmd/swagger/swagger.go
```diff
@@ -33,6 +33,15 @@ var opts struct {
 }
 
 func main() {
+	// Recovering from internal panics
+	defer func() {
```
casualjim: this changes makes it so there is no more stacktrace, i don't think that is what we want

generator/shared.go
fredbi: Actually I didn't find where this method could be used in the code. The only use I could find was through templates: hence forcing empty string

generator/operation_test.go
fredbi: Yes. My change just came from go fmt... But I'll do it fine.

cmd/swagger/swagger.go
fredbi: The point is to catch errs from methods which do not return error.
My concern was essentially the key "baseImport()" method.
I can let the panic go if you prefer. I suggest to leave it as a comment: best to catch the stack hen the product is more stable.

generator/model.go
fredbi: Yes. Cute.

generator/shared.go
fredbi: Same as TargetPath() method.

generator/support.go
fredbi: Oops. Sure I did

fredbi:
@casualjim , so let's recap your review:

  1.  comment out the defer() in main, since we expect stacks on panic().
  I hope in the future, panic() would be better handled.
  Possible alternative: baseImport() returns error (more changes)
  2. more idiomatic go with error printing (e.g; "%v",err instead of %s,err.Error()
  3. replace fmt.Print with log.Printf 
  4. operation_test.go : more idiomatic string joining (I didn't write this code, but ok to fix it)

 5. Your opinion on the criticality of SpecPath() and TargetPath(): in my view, it's okay (and not much changing with master to return blank string on errors - just more explicit)

Okay with that?

Cheers,

Fred

generator/support.go
```diff
@@ -170,15 +170,16 @@ func checkPrefixAndFetchRelativePath(childpath string, parentpath string) (bool,
 
 }
 
+// TODO: handle error with a returned error rather than panic()
 func baseImport(tgt string) string {
 	tgtAbsPath, err := filepath.Abs(tgt)
 	if err != nil {
-		log.Fatalln(err)
+		log.Fatalln("could not evaluate base import path with target \"%s\". Target directory must be created beforehand: %v", tgt, err)
```
casualjim: you probably mean log.Fatalf here
ping me in slack so we can walk through getting an env where the editor you use tells you about these stylistic errors

fredbi: Yes, you're probably right I haven't looked hard enough into vim add-ons.
Don't merge the PR yet: a number of unit test is coming up.
I let the CI run to figure out the coverage.

fredbi:
There is still annoying behaviors on Windows. I installed everything on Windos and trying to fix things up on this os.

fredbi:
Should be fine on windows now. Unit test suite (culprit: TestBaseImport) did block when not run as admin. Fixed that by skipping test if OS does not enable symlink for the current user.

casualjim:
💯 great work

fredbi:
Thanks Ivan for the cheer up. 

Difficult to get around codecov/patch since this PR specialized on error handling and that few unit tests did dig into all error causes.

Mostly codecov/patch complains about "untested" deltas resulting from mere gofmt or log message. Untested they were, untested they remain (for now).

Anyhow, I tried hard to improve test coverage (btw codecov result is disappointing: I got a better impression on my go tool cover stats...).

_En passant_, fixed a small bug: loading custom template files (not assets, new ones) from TargetTemplate and not ".".

Next contribution will be I think a small tutorial about playing with templates...

casualjim:
yeah don't worry about the codecov patch stuff. Code coverage is a poor measure of quality anyway.

PR #1300, opened 2017-12-05
1299-fix-unit-test-verbosity

fredbi:
Minor change in order to better get the essentials when running CI checks.


fredbi:
Grr still too chatty...

fredbi:
Now I may read Appveyor's output.

PR #1302, opened 2017-12-05
Allow adding or overriding consumers and producers

### LINKED ISSUE #1301: Allow adding/overriding Consumers and Producers
## Problem statement

go-swagger knows how to map certain mime types to the relevant `runtime.Producer` and `runtime.Consumer`. If you want to use a mime type that isn't supported (such as `application/pkcs10`), there currently isn't a way to do that without editing generated files.

I will shortly submit a partially completed PR for feedback. This introduces new methods into the `restapi/operations/<service>_api.go` to allow callers to configure the service in the `restapi/configure_<service>.go` file and add consumers and producers.

If there is interest in this change, I'll finish the job (docs, tests etc.) and update the PR.

dmjones500:
Fixes #1301.

Note: PR is not complete (not doc changes, testing). Just wanted to get some feedback on the idea before finalising the work.


dmjones500:
Hmmm, seems I've checked in something dependent on my own system. I'll try and figure out where that came from.

casualjim:
how would that custom consumer/producer be picked up at runtime?

dmjones500:
In the `restapi/configure_foo.go`, you would register the consumer/producer against an application type. For example:

````go
api.RegisterConsumer("application/pkcs10", customConsumer)
````

In my testing, that was all I needed to do to support reading PKCS#10 input. Because  my schema defined the type as string, my custom consumer expects to unmarshall into a string pointer.

**Comment by casualjim** [comment]
yes you're right, I verified after asking the question.. in this case: carry on :)

**Comment by dmjones500** [comment]
@casualjim How would you normally test this sort of change? I can add some pattern matching tests to `/go-swagger/generator/server_test.go` to confirm the methods have been added. Is there an alternative approach in which you execute the generated code and verify the methods work as expected? (I'm still groking the code base, so some pointers would be great).

**Comment by fredbi** [comment]
A few doc links glitches remain... I'll patch them soon

**Comment by dmjones500** [comment]
@casualjim  This is potentially ready to go, although I'm not sure how to resolve the CircleCI issues. It's failing for odd reasons:

https://circleci.com/gh/thales-e-security/go-swagger/5

```
#!/bin/bash -eo pipefail
.circleci/build-docker.sh
?   	github.com/go-swagger/go-swagger	[no test files]
?   	github.com/go-swagger/go-swagger/cmd/swagger	[no test files]
PASS
testing: open /go/src/github.com/go-swagger/go-swagger/github.com/go-swagger/go-swagger/cmd/swagger/commands/profile.tmp: no such file or directory
FAIL	github.com/go-swagger/go-swagger/cmd/swagger/commands	9.401s
Exited with code 1
```

fredbi:
It is the hack script to build the coverage report that fails

dmjones500:
@fredbi Do I need to worry about that? I could rebase this PR to contain one commit, which I assume would trigger a new build.

fredbi:
It might be that you have an outdated version of this script in your branch. The "rm -f" deletion of the temp file has been added a few months ago.

dmjones500:
> It might be that you have an outdated version of this script in your branch

I don't think so. My branch only alters 5 files vs master, none of them is related to circle CI.

fredbi:
I know it doesn't.

fredbi:
But still, the CI runs your branch, before merging

dmjones500:
How do you suggest we move forward?

dmjones500:
It must have been my own CI configuration messing things up. I rebased my commits to trigger another build and this seems more hopeful.

fredbi:
Before all compare the ./.circleci/build-docker.sh in your branch with the master

dmjones500:
AFAICT, the code coverage drop is simply because some of the generated data is longer with this change.

casualjim:
thanks!

dmjones500:
@casualjim When are you next planning a release? I will shortly be publishing software that relies on this feature, so it would be great to point people at a tagged release rather than master.

Issue #1171, opened 2017-09-15
Go-swagger v0.12.0 - invalid memory address or nil pointer dereference


bobvanluijt:
## Problem statement

On Mac, go-swagger v0.12.0, the command: `swagger generate server --name=weaviate --spec=https://raw.githubusercontent.com/weaviate/weaviate-swagger/develop/weaviate-swagger.json --default-scheme=https` fails with:

```
2017/09/15 12:07:07 building a plan for generation
2017/09/15 12:07:07 planning definitions
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xd pc=0x156bc09]

goroutine 1 [running]:
github.com/go-swagger/go-swagger/generator.(*schemaGenContext).makeGenSchema(0xc42079a000, 0xc421fbcf34, 0x6)
	/go/src/github.com/go-swagger/go-swagger/generator/model.go:1390 +0x5f9
github.com/go-swagger/go-swagger/generator.(*schemaGenContext).buildProperties(0xc42079b800, 0xc42079b8a8, 0x100)
	/go/src/github.com/go-swagger/go-swagger/generator/model.go:660 +0x304
github.com/go-swagger/go-swagger/generator.(*schemaGenContext).makeGenSchema(0xc42079b800, 0x0, 0xc421355740)
	/go/src/github.com/go-swagger/go-swagger/generator/model.go:1433 +0x9f3
github.com/go-swagger/go-swagger/generator.makeGenDefinitionHierarchy(0xc421fbd4b0, 0xf, 0xc421296c88, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/go-swagger/go-swagger/generator/model.go:166 +0x32c
github.com/go-swagger/go-swagger/generator.makeGenDefinition(0xc421fbd4b0, 0xf, 0xc421296c88, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/go-swagger/go-swagger/generator/model.go:133 +0xbc
github.com/go-swagger/go-swagger/generator.(*appGenerator).makeCodegenApp(0xc42064a0f0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/go-swagger/go-swagger/generator/support.go:608 +0x5d9
[...]

PR #1305, opened 2017-12-05
1171. Add fixture

fredbi:
Issue#1171 has been much embroiled...
There are basically 2 cases mentionned:
- object with an items property : covered with fixture for issue#1238 (more or less a duplicate of #1171, only clearer)
- items missing in array. This case specifically covered by the present fixture.

Together with issue#1238, both cases discussed are covered by UT (new exerciser for cmd/swagger/validate).


fredbi:
Of course, merge issue...

fredbi:
Thanks Ivan.
Now I believe we can close #1171 and move on further.

Issue #1309, opened 2017-12-05
Generated client parameter does not set request body for arbitrary object.


e-zhang:
## Problem statement
For our given spec we have an endpoint that takes in any arbitrary JSON object. This is picked up by go-swagger as an anonymous interface in go (interface{}). However in the corresponding WriteToRequest method of the relevant parameter, there is no SetBodyParam call on the request, even though the payload member is set and defined. 

## Swagger specification
```
swagger: '2.0'
info:
  title: Test
  version: '1'
paths:
  /test:
    post:
      parameters:
        - name: test
          in: body
          schema:
            type: object
      responses:
        200:
          description: success
```

The output function is
```
/*PostTestParams contains all the parameters to send to the API endpoint
for the post test operation typically these are written to a http.Request
*/
type PostTestParams struct {

        /*Test*/
        Test interface{}

        timeout    time.Duration
        Context    context.Context
        HTTPClient *http.Client
}

// WriteToRequest writes these params to a swagger request
func (o *PostTestParams) WriteToRequest(r runtime.ClientRequest, reg strfmt.Registry) error {

        if err := r.SetTimeout(o.timeout); err != nil {
                return err
        }
        var res []error

        if len(res) > 0 {
                return errors.CompositeValidationError(res...)
        }
        return nil
}
```

It looks like the codegen checks for IsNullable 
https://github.com/go-swagger/go-swagger/blob/master/generator/templates/client/parameter.gotmpl#L238

In the type resolver, objects with no properties are defaulted to `IsNullable = false`
https://github.com/go-swagger/go-swagger/blob/master/generator/types.go#L540

I was hoping to get some guidance for my scenario in this case where the API needs to send/accept arbitrary JSON objects (no pre-defined properties). Is this the expected behavior for go-swagger? If it is, are there any recommendations for the spec to handle this?

casualjim:
this is a bug it should really submit that body

e-zhang:
Ok thanks for the quick response. 

Do you know what the proper fix would be? I can submit a pull request, but I wasn't quite sure where the fix would be. Would I be breaking anything if I removed the `IsNullable` check from the parameter.gotmpl?

casualjim:
I think make line 238 read: `{{ if or .Schema.IsInterface (and .Schema.IsNullable (not .HasDiscriminator)) }}`

PR #1313, opened 2017-12-07
client request parameter will set body param for array of interface{}

e-zhang:
Sorry I missed this previously, but same as https://github.com/go-swagger/go-swagger/pull/1310 / https://github.com/go-swagger/go-swagger/issues/1309, but for []interface{}

```
swagger: '2.0'
info:
  title: Test
  version: '1'
paths:
  /test:
    post:
      parameters:
        - name: test
          in: body
          schema:
            type: array
            items:
              type: object
      responses:
        200:
          description: success
```

Also doesn't write body on `WriteToRequest`


generator/templates/client/parameter.gotmpl
```diff
@@ -235,7 +235,7 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}Params) WriteToRequest(r runtime
   {{ end }}
 
   {{if .IsBodyParam}}
-  {{ if or .Schema.IsInterface (and .Schema.IsNullable (not .HasDiscriminator)) }}if {{ .ValueExpression }} != nil {
+  {{ if or .Schema.IsInterface (and .Schema.IsArray .Child .Child.IsInterface) (and .Schema.IsNullable (not .HasDiscriminator)) }}if {{ .ValueExpression }} != nil {
```
casualjim: I think that may need to be `.Schema.Child`

e-zhang: There are other references in the file to `.Child`. I get an error with `.Schema.Child`
`at <.Schema.Child>: can't evaluate field Child in type *generator.GenSchema`

Issue #1331, opened 2017-12-19
Temporary files are not deleted on program termination


sebastien-rosset:
## Problem statement

The swagger utility creates temporary files and do not delete these files before program termination.

In particular, "swagger validate" calls the ExpandSchema() method from the dependent go-openapi/spec to validate swagger spec files. Every invocation of "swagger validate" may end up creating multiple temporary files in the temp folder (e.g. /tmp for Linux).

The ExpandSchema() method calls ioutil.TempFile(), which creates a temporary file, and the file is not deleted before program termination. As specified in the golang documentation, it is the caller's responsibility to delete temporary files created by ioutil.TempFile()

https://github.com/go-openapi/spec/blob/aa90e7cfc0a78736511f9eeab8f7586d98c8f53a/expander.go#L534

## Steps to reproduce
1) Invoke swagger validate xyz.json
2) After program execution, notice the /tmp directory (in Linux) contains multiple rootXXXX files.

## Environment
swagger version: x.x.x  
go version: 1.9.2
OS:  Linux

## Proposed fix
The tracking issue in go-openapi/spec is https://github.com/go-openapi/spec/issues/47.
Pull Request: https://github.com/go-openapi/spec/pull/48

The proposed fix for go-swagger is to update Gopkg.lock and use https://github.com/go-openapi/spec/pull/48/commits/

fredbi:
@sebastien-rosset 
Do you confirm the issue may be closed now that your PR has been merged?

sebastien-rosset:
Yes, I confirm the issue can be closed. thanks.

PR #1333, opened 2017-12-20
1331 - Delete temporary files

sebastien-rosset:
See issue https://github.com/go-swagger/go-swagger/issues/1331
A commit would also be required in the vendor sub-directory; I'm not sure how you manage the commits for the vendor sub-directory.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -102,7 +102,7 @@
   branch = "master"
   name = "github.com/go-openapi/spec"
   packages = ["."]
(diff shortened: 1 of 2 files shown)


casualjim:
seems odd that the only change is a revision in gopkg and not any files in vendor

sebastien-rosset:
Do you use a specific procedure to update the files under the vendor directory?

casualjim:
```
go get -u github.com/golang/dep/cmd/dep
dep ensure -update
```

PR #1345, opened 2017-12-28
case sensitive import path on Windows

koron:
Generated codes on Windows by go-swagger can't be compiled when the package path has case sensitive name.

For example, the name of pacakge is `github.com/FooBar/awesome`, go-swagger generate codes which have `"github.com/foobar/awesome/xxx"` or so in import paths on Windows.  It causes compilation errors because golang compiler checks import paths by case sensitive.

This PR modifies `checkPrefixAndFetchRelativePath()` function to compare two paths by case insensitive on Windows, but create relative path by case sensitive.

### DIFF
--- a/generator/shared.go
+++ b/generator/shared.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
-	goruntime "runtime"
 	"sort"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


casualjim:
thanks, sorry I missed this PR earlier

koron:
thank you 🙇

PR #1364, opened 2018-01-08
Adding fix for issue #1341

### LINKED ISSUE #1341: Improper UnmarshalJSON of polymorphic object contained in another object
## Problem statement

Code from April

    // UnmarshalJSON unmarshals this object with a polymorphic type from a JSON structure
    func (m *ExecuteValues) UnmarshalJSON(raw []byte) error {
	var data struct {
		Array []*ExecuteValues `json:"Array,omitempty"`
		Value json.RawMessage `json:"Value,omitempty"`
	}

	buf := bytes.NewBuffer(raw)
	dec := json.NewDecoder(buf)
	dec.UseNumber()

	if err := dec.Decode(&data); err != nil {
		return err
	}

	value, err := UnmarshalExecuteValue(bytes.NewBuffer(data.Value), runtime.JSONConsumer())
	if err != nil && err != io.EOF {
		return err
	}

Current code

    func (m *ExecuteValues) UnmarshalJSON(raw []byte) error {
	var data struct {
		Array []*ExecuteValues `json:"Array,omitempty"`
	}

	buf := bytes.NewBuffer(raw)
	dec := json.NewDecoder(buf)
	dec.UseNumber()

	if err := dec.Decode(&data); err != nil {
		return err
	}

	value, err := UnmarshalExecuteValue(bytes.NewBuffer(raw), runtime.JSONConsumer())
	if err != nil && err != io.EOF {
		return err
	}

Notice the difference in what is passed to `UnmarshalExecuteValue()`.  It's being passed the data for the entire `ExecuteValues` object, rather than the `ExecuteValue` sub-object.

## Swagger specification

    ExecuteValues:
      type: object
      properties:
        Value:
          $ref: "#/definitions/ExecuteValue"
        Array:
          type: array
          items:
            $ref: "#/definitions/ExecuteValues"

    ExecuteValue:
      type: object
      discriminator: ValueType
      required:
      - ValueType
      properties:
        ValueType:
          type: string

## Environment
swagger version: master 
go version: 1.9.2

nickrmc83:
Properties Unmarshal<type> was not consistent with how allOf properties
are unmarshalled. These changes brings the two into line and fixes #1341.

Fixed test for bug 423.


gregmarr:
@nickrmc83 fixtures/bugs/1341/swagger.yaml is missing

gregmarr:
:shipit:

PR #1372, opened 2018-01-09
[WIP] Enable simple param format validation

### LINKED ISSUE #863: Non-compiling generated code when parameter is array type with enum validation and underscore in name
## Problem statement
If there is a query parameter that meets the following criteria:
1. The name has an underscore, e.g., some_parameter
2. The type is an array (of strings for the case I found)
3. The items of that array undergo enum validation

## Swagger specification
```
get:
   ...
   ...
   parameters:
     - name: some_parameter
        in: query
        ...
        type: array
        items:
          type: string
          enum:
            - enum1
            - enum2
            - enum3
```

The result of running swagger produces generated code that does not compile at the following sha: e4738db

In the generated code for getting the parameters, there's code like this:
```
someParameterI := someParameterIV

if err := validate.Enum(fmt.Sprintf("%s.%v", "some_parameter", i), "query", some_parameterI, []interface{}{"enum1", "enum2", "enum3"}); err != nil {
	return err
}
```

which obviously does not compile.

### LINKED ISSUE #909: UUIDs do not appear to be validated
## Problem statement

In a server: When attempting to validate a UUID in client-supplied data, the generated code under `restapi/operations` will make a call like

    value, err := formats.Parse("uuid", raw)

This will cause the `Parse()` function in `go-openapi/strfmt/format.go` to be called, which will attempt to validate the format of the client-supplied data by applying the `strfmt.UUID.UnmarshalText()` function. 

However, the `strfmt.UUID.UnmarshalText()` function (defined in `go-openapi/strfmt/default.go`) actually accepts any input string and not just those formatted as UUIDs. (There is a comment associated with that function: "validation is performed later on".)

Should validation of UUIDs should actually be performed using `strfmt.IsUUID()`?

## Environment

swagger version: 0.8.0 (commit: d1a0b6c1007402b2ccde73702eb0e8e43a5e7498)
go version: 1.7.4 linux/amd64 
OS: Fedora 25

### LINKED ISSUE #961: Validation for extended string format

### LINKED ISSUE #1373: Generated server with form-data required=false evaluates file to nil
## Problem statement
When formData parameter is not required (required: false OR omitted), in the generated server io.ReadCloser always evaluates  to nil, even when file is supplied by the client.

When parameter is required it works fine

## Swagger specification
```yaml
paths:
  /jobs:
    post:
      operationId: JobCreator
      consumes:
        - multipart/form-data
      parameters:
        - in: formData
          name: yamlfile
          type: file
          description: Job description yaml file
      produces:
        - application/json
      responses:
        '201':
          description: Created
          schema:
            $ref: '#/definitions/JobID'
        '400':
          description: Bad request
          schema:
            $ref: '#/definitions/ErrResponse'
        '415':
          description: Unsupported media type (bad syntax)
          schema:
            $ref: '#/definitions/ErrResponse'
        '422':
          description: Invalid input (bad semantics)
          schema:
            $ref: '#/definitions/ErrResponse'
```
## Steps to reproduce
1. Generate server with a POST path and formData parameter that is NOT required
2. evaluate io.ReadCloser member of the parameters struct
3. make a request and check the io.ReadCloser member of the params struct
## Environment
swagger version: dev (pulled 10/JAN/2018 15:20 UTC)
go version: 1.9.2 
OS:   ubuntu 16.04

### LINKED ISSUE #1378: Default values for simple arrays generate code that does not compile
## Problem statement
When declaring default values for arrays, eg:
```yaml
default:
- a
- b
- c
```

the generated code for server does not produce a correct syntax.

Moreover, generated code attempts to parse the default value using the CollectionFormat, which is inconsistent with the model the value complies with.

## Swagger specification
@todo

## Environment
swagger version: [current master]
go version: 1.9.1
OS:   linux ubuntu

### LINKED ISSUE #1384: Generated configure file has confusing directions.
## Problem statement

Line 1 says "don't edit": https://github.com/go-swagger/go-swagger/blob/18337a2ecf1022fa5e6b3b01b3ac4033e1e4955b/generator/templates/server/configureapi.gotmpl#L1

But line 28 says "you should edit this file": https://github.com/go-swagger/go-swagger/blob/18337a2ecf1022fa5e6b3b01b3ac4033e1e4955b/generator/templates/server/configureapi.gotmpl#L28

These messages should be made consistent.

fredbi:
**This PR has been closed and replaces by PR #1394**

- fixes #909 
- fixes #961 (duplicate)
- fixes #1378 
- fixes #1384 
- fixes #863
- fixes #1373 

- fixed similar issues with nested arrays and default values in response headers
- this PR does **NOT** address known related issues with client generation
- however, urlbuilder now supports collections and nested collection for path params

## On the ci/testing side
- [x] Reviewed CircleCI config to split the original script in several more tractable and visible steps
- [x] Reviewed the run-canary.sh test vs valid/invalid specs. Added some "fixed" specs to move on further with generation testing ["fixed" means that spec validation issues where corrected to generate properly]
- [x] Introduced new code generation build test on ./fixtures/codegen + some other fixtures. The objective is to generate a server and build it


gregmarr:
I saw a few places where "literal" had an extra "t": "litteral".

fixtures/codegen/todolist.simplepath.yml
```diff
@@ -33,6 +33,7 @@ paths:
           in: path
           type: integer
           description: an integer  property
+          required: true
```
casualjim: required: true is implied by in: path   it doesn't have to be provided per spec

fixtures/canary/ms-cog-sci-fixed/swagger.json
```diff
@@ -0,0 +1,649 @@
+{
```
casualjim: These are specs taken from actual services. What was fixed?

fixtures/codegen/todolist.simple-fixed.yml
```diff
@@ -0,0 +1,187 @@
+swagger: '2.0'
```
casualjim: what does the `-fixed` mean? What was fixed?

generator/model.go
```diff
@@ -1363,7 +1363,7 @@ func (sg *schemaGenContext) makeGenSchema() error {
 		return nil
 	}
 	if Debug {
-		log.Println("after shortcuit named ref")
```
casualjim: this is a misspelling of short circuit, not shortcut

generator/templates/server/responses.gotmpl
```diff
@@ -28,10 +28,11 @@ if {{ .ValueExpression }}S != "" {
 {{ end }}
 {{define "sliceserverheaderbuilder" }}
 var {{ varname .Child.ValueExpression }}R []string
-for _, {{ varname .Child.ValueExpression }} := range {{ .ValueExpression }} {
-  {{- if not .Child.IsArray }}{{ template "sliceitemserverheaderbuilder" .Child }}{{ else }}{{ template "sliceserverheaderbuilder" .Child }}{{ end -}}
+if len({{ .ValueExpression }}) > 0 {
```
casualjim: doesn't a for loop imply `if len > 0` ?

generator/template_repo.go
```diff
@@ -207,6 +208,15 @@ func asPrettyJSON(data interface{}) (string, error) {
 	return string(b), nil
 }
 
+func asGoInitializer(data interface{}) (string, error) {
+	// Constructs a Go literal initializer from interface{}
+	b, err := json.Marshal(data)
+	if err != nil {
+		return "", err
+	}
+	return strings.Replace(strings.Replace(strings.Replace(string(b), "}", ",}", -1), "[", "{", -1), "]", ",}", -1), nil
```
casualjim: bytes has the same set of methods, so you don't strictly need to convert to string to do this.
How does this deal with property names?

fixtures/canary/ms-cog-sci-fixed/swagger.json
fredbi: Theses specs are somehow invalid for go-swagger (docker, ms-cog, bitbucket).
I kept the original version so we can continue to measure our success against this yardstick.

However, for cases of legitimate spec invalidation by `swagger validate`, I produced a "fixed" version which is valid and allow us to check our full generation.

Spec "fixes" are as follows:
- docker: made OperationId unique
- ms-cog: duplicate property in allOf construct (which is valid JSON but refused by our validator, which is debatable, I don't know the expected behavior from Swagger 2.0 in this case)

We have other problems for valid canary specs.
These ones are not fixable at the spec level: there are simply excluded from the test run:
- bitbucket: wrong model generation
- kubernetes: use of options in mime type is not supported (produces duplicates in switch constructs) (kubernetes spec produces a valid client, though)

fixtures/codegen/todolist.simple-fixed.yml
fredbi: You have been using invalid spec for codegen unit testing for quite a while.
It doesn't hurt since you are using low level templating UT, but it's not workable for full code generation.

"-fixed" means that this is a valid spec.

I tried to change as little as possible because of the code pattern matching in UT.
Hence, in cases where I wished to non-regress your fixtures, I add to make a copy and make it a valid spec.

Changes: 
- todolist.simple-fixed.yml : default value must validate the data type, path params are required
- azure-analysis: same as ms-cog i canary (duplicate property in allOf)

I introduced reuse codegen specs in a new CI test run (./hack/codegen-nonreg.sh) to extend our testing capabilities. However many fixtures fail for full generation. All this is explained as comments in the test script.

fixtures/codegen/todolist.simplepath.yml
fredbi: Perhaps it should, but it is not working like this currently.
You've just opened a new issue in the go-openapi/spec package... 💯 

required: true is not implied and (currently) must be provided explicitly. This check is carried on by validate.

generator/model.go
fredbi: OK

generator/template_repo.go
fredbi: This part is under full rework as it is not satisfactory and did not pass my internal round of testing.

Initialization of default values for parameters is actually quite a bit more complex than expected,
when considering formats (which do not expose a consistent initialization interface), arrays and nested arrays.

This method was originally intended to initialize slices.

We'll have another round of discussion on this with forthcoming commit

generator/templates/server/responses.gotmpl
fredbi: Sure, when it is nested.
But the template is self-reproduced and the first instance and we need to guard against possible ranging into the next level.
Here is an example of code I am currently generating with tests:
Please mark that array outermost validation still go to the validate private func.
For nested arrays, we are compelled to produce some less elegant code.

For param spec:
```yaml
/mixed/deepnested:
    get:
      operationId: getInDepth
      parameters:
      - name: deepNested1
        in: query
        type: array
        collectionFormat: pipes
        required: true
        maxItems: 2
        items:
          type: array
          collectionFormat: csv
          minItems: 2
          items:
            type: array
            collectionFormat: ssv
            uniqueItems: true
            items:
              type: string
              pattern: '^[a-z]$'
```

I got:
```golang
func (o *GetInDepthParams) bindDeepNested1(rawData []string, hasKey bool, formats strfmt.Registry) error {
	if !hasKey {
		return errors.Required("deepNested1", "query")
	}

	var qvDeepNested1 string
	if len(rawData) > 0 {
		qvDeepNested1 = rawData[len(rawData)-1]
	}

	deepNested1IC := swag.SplitByFormat(qvDeepNested1, "pipes")

	if len(deepNested1IC) == 0 {
		return errors.Required("deepNested1", "query")
	}

	var deepNested1IR [][][]string
	for i, deepNested1IV := range deepNested1IC {
		deepNested1IIC := swag.SplitByFormat(deepNested1IV, "csv")
		if len(deepNested1IIC) > 0 {

			var deepNested1IIR [][]string
			for ii, deepNested1IIV := range deepNested1IIC {
				deepNested1IIIC := swag.SplitByFormat(deepNested1IIV, "ssv")
				if len(deepNested1IIIC) > 0 {

					var deepNested1IIIR []string
					for iii, deepNested1IIIV := range deepNested1IIIC {
						deepNested1III := deepNested1IIIV

						if err := validate.Pattern(fmt.Sprintf("%s.%v", fmt.Sprintf("%s.%v", fmt.Sprintf("%s.%v", "deepNested1", i), ii), iii), "query", deepNested1III, `^[a-z]$`); err != nil {
							return err
						}

						deepNested1IIIR = append(deepNested1IIIR, deepNested1III)
					}

					deepNested1IIR = append(deepNested1IIR, deepNested1IIIR)
				}

				// uniqueItems: true
				if err := validate.UniqueItems(fmt.Sprintf("%s.%v", fmt.Sprintf("%s.%v", "deepNested1", i), ii), "query", deepNested1IIIC); err != nil {
					return err
				}

			}

			deepNested1IR = append(deepNested1IR, deepNested1IIR)
		}

		deepNested1ISize := int64(len(deepNested1IIC))

		// minItems: 2
		if err := validate.MinItems(fmt.Sprintf("%s.%v", "deepNested1", i), "query", deepNested1ISize, 2); err != nil {
			return err
		}

	}

	o.DeepNested1 = deepNested1IR
	if err := o.validateDeepNested1(formats); err != nil {
		return err
	}

	return nil
}

func (o *GetInDepthParams) validateDeepNested1(formats strfmt.Registry) error {

	deepNested1Size := int64(len(o.DeepNested1))

	// maxItems: 2
	if err := validate.MaxItems("deepNested1", "query", deepNested1Size, 2); err != nil {
		return err
	}

	return nil
}
```

fredbi:
@casualjim I am well aware that this PR started with a very small change, and is gradually growing.
This was unexpected.

**So your review at this intermediate stage is welcome.**

Inspection and thorough testing the generated code for BindRequest() for parameters showed out many shortcomings and generation bugs (e.g. arrays, nested arrays, default values...).

**This is the main reason why I had to muscle up the CI.**

So far, we simply checked generated code with pattern. Now, we should be able to check that it compiles.

On my side, I have a more advanced testing suite which unit-tests BindRequest() with fixtures and check that validations are well processed. I am planning to contribute this as an additional template (with a `--gen-test=operations` option for instance) but it needs a lot of refactoring on my side to be generic (currently it is very much targeted to my own usage, since the unit test integrates an Hyperledge Fabric mock inside and checks the API response....).

Sample fixture used to test this P.R. Obviously, I haven't been able to generate such fixtures automatically...

```json
[
   {
        "expectInvalid": false,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c,a b c|d e f, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c,a b c|d e f, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c,a b c|d e f, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c,a b c|d e f, g h i|x y z, u v w"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c,a b c|d e f, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c,a b c|d e f, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c,a b c|d e f, g h i|x y z, u v w"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c|d e f, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c,a b c|d e f, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c, a b c|d e f, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c|d e f, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c, a b b|d e f, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c, a b c|d e f, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c, a b c|d e f, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c, a b c|d e e, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c, a b c|d e f, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a 9 c, a b c|d e f, g h i"
            }
        ]
   },
   {
        "expectInvalid": true,
        "params": [
            {
                "in": "query",
                "name": "deepNested1",
                "value": "a b c, a b c|d e 9, g h i"
            },
            {
                "in": "header",
                "name": "deepNested2",
                "value": "a b c, a b c|d e f, g h i"
            }
        ]
   }
]
```

fixtures/codegen/todolist.simplepath.yml
gregmarr: No, it is not implied by path.  It is required and must be true.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md

>Determines whether this parameter is mandatory. If the parameter is in "path", this property is **required** and its value MUST be true. Otherwise, the property MAY be included and its default value is false.

generator/templates/server/responses.gotmpl
gregmarr: The outer if is unnecessary, as the only thing inside the if is a for loop over the thing that len is testing.  These two blocks have exactly the same meaning.

    for _, {{ varname .Child.ValueExpression }} := range {{ .ValueExpression }} {
    }

and 

    if len({{ .ValueExpression }}) > 0 {
        for _, {{ varname .Child.ValueExpression }} := range {{ .ValueExpression }} {
        }
    }

fredbi: I don't remember how exactly, but I had a case of panic() when testing and added the guard in a hurry. I've to look at it more closely the

fredbi: Ah got it now (it has been a few days...): the if does not only guard the loop but the append() after the loop too. So it's not completely useless.

gregmarr: I don't see it.  Do the nested templates do something funky like closing the loop and opening a new one?  I don't see it there either.

    if len({{ .ValueExpression }}) > 0 {
        for _, {{ varname .Child.ValueExpression }} := range {{ .ValueExpression }} {
            {{- if not .Child.IsArray }}
                {{ template "sliceitemserverheaderbuilder" .Child }}
            {{ else }}
                {{ template "sliceserverheaderbuilder" .Child }}
            {{ end -}}
        }
    }

fredbi: BTW hello Greg. And thank you for your comments.

You are looking at the wrong place (this part is to generate headers not parameters - yes, I had to change it too to support arrays, this story shall also be told...).

Look at the sample generated code I provided above. I believe it's more understandable than the template, which is well, ahem, funky, yes.

gregmarr: I think you're looking at a different place than we are.  We're looking at a simple `for` loop that was wrapped in an `if` in this PR.  There is no such structure in your generated code above.

https://github.com/fredbi/go-swagger/blob/e77845a4df1c90949bd7a9091ae2344f953d8f2f/generator/templates/server/responses.gotmpl#L31-L35

gregmarr: I think you're looking at this, which does contain the `append`:  

https://github.com/fredbi/go-swagger/blob/e77845a4df1c90949bd7a9091ae2344f953d8f2f/generator/templates/server/parameter.gotmpl#L45-L48

fredbi: yeah i am focused on parameters right now. that's the most complex. i have to look again for the response - this one was a quick fix - thanks

generator/template_repo.go
fredbi: I am not sure that `string(bytes.Replace(.... Replace(b,[]byte("}"), [}byte(",}",-1), []byte("["), []byte("{"....)`
would be more readable... 
I am adding tests to demonstrate use with properties, even if the initial intent is especially to deal with slices.

PR #1394, opened 2018-01-31
Fix issues on simple params

### LINKED ISSUE #863: Non-compiling generated code when parameter is array type with enum validation and underscore in name

### LINKED ISSUE #909: UUIDs do not appear to be validated

### LINKED ISSUE #961: Validation for extended string format

### LINKED ISSUE #1373: Generated server with form-data required=false evaluates file to nil

### LINKED ISSUE #1378: Default values for simple arrays generate code that does not compile

### LINKED ISSUE #1384: Generated configure file has confusing directions.

fredbi:
Fix issues on simple params: collections, nested collections and default values. Applied same rules for headers in responses.

_Supersedes PR #1372_

## Fixed issues
- fixes #909 
- fixes #961 (duplicate)
- fixes #1378 
- fixes #1384 
- fixes #863
- fixes #1373 

- fixed similar issues with nested arrays and default values in response headers
- this PR does **NOT** address known related issues with client generation
- however, urlbuilder now supports collections and nested collection for path params

## On the ci/testing side
- [x] Reviewed CircleCI config to split the original script in several more tractable and visible steps
- [x] Reviewed the run-canary.sh test vs valid/invalid specs. Added some "fixed" specs to move on further with generation testing ["fixed" means that spec validation issues where corrected to generate properly]
- [x] Introduced new code generation build test on ./fixtures/codegen + some other fixtures. The objective is to generate a server and build it


fredbi:
Found a bug with duration type. Patching up.
For the longer term, I believe we would need a less hacky way of figuring out how to initialize custom types.

generator/template_repo.go
```diff
@@ -207,6 +213,58 @@ func asPrettyJSON(data interface{}) (string, error) {
...
+	// NOTE: this is currently used to construct simple slice intializers for default values.
+	// This allows for nicer slice initializers for slices of primitive types and avoid systematic use for json.Unmarshal().
+	b, err := json.Marshal(data)
+	if err != nil {
+		return "", err
+	}
+	return strings.Replace(strings.Replace(strings.Replace(string(b), "}", ",}", -1), "[", "{", -1), "]", ",}", -1), nil
+}
+
+func hasDefaultParam(parameters GenParameters) bool {
```
casualjim: it seems odd that this has to be done at render time. Why isn't this just set on the parameter groups etc?

generator/template_repo.go
```diff
@@ -207,6 +213,58 @@ func asPrettyJSON(data interface{}) (string, error) {
...
+	// NOTE: this is currently used by templates to avoid empty constructs
+	for _, param := range parameters {
+		if param.HasDefault {
+			return true
+		}
+	}
+	return false
+}
+
+func hasDefaultHeader(headers GenHeaders) bool {
```
casualjim: same remark as the paramters

generator/template_repo.go
```diff
@@ -207,6 +213,58 @@ func asPrettyJSON(data interface{}) (string, error) {
...
+	// NOTE: this is currently used by templates to avoid empty constructs
+	for _, header := range headers {
+		if header.HasDefault {
+			return true
+		}
+	}
+	return false
+}
+
+func asItemsLocation(items interface{}) string {
```
casualjim: this is again not something that needs to be done in the templates but rather it's context so should go on the GenXxx models

generator/operation.go
```diff
@@ -950,25 +977,37 @@ func (b *codeGenOpBuilder) MakeParameter(receiver string, resolver *typeResolver
 			Enum:             param.Enum,
 		}
 
+		res.ZeroValue = res.resolvedType.Zero()
+
 		if param.Items != nil {
+			if res.CollectionFormat == "" {
+				// When not defined,CollectionFormat defaults to csv
+				res.CollectionFormat = "csv"
```
casualjim: I'm not a fan of making it explicit like this because it risks the chance of breaking a default and sprinkles this handling all over the code. Whereas handling an empty value as default in the things that deal with strfmt seems wiser

generator/operation.go
```diff
@@ -784,25 +796,34 @@ func (b *codeGenOpBuilder) MakeParameterItem(receiver, paramName, indexVar, path
 	res.Name = paramName
 	res.Path = path
 	res.Location = location
-	res.ValueExpression = swag.ToVarName(valueExpression)
+	res.ValueExpression = golang.MangleVarName(swag.ToGoName(valueExpression))
```
casualjim: why is there a need to change this?

generator/operation.go
```diff
@@ -784,25 +796,34 @@ func (b *codeGenOpBuilder) MakeParameterItem(receiver, paramName, indexVar, path
...
 	res.CollectionFormat = items.CollectionFormat
+	if res.CollectionFormat == "" && items.Items != nil {
+		// When not defined,CollectionFormat defaults to csv
+		res.CollectionFormat = "csv"
```
casualjim: not a fan of this providing a value

generator/operation.go
```diff
@@ -739,8 +743,12 @@ func (b *codeGenOpBuilder) MakeHeaderItem(receiver, paramName, indexVar, path, v
 	res.Name = paramName
 	res.Path = path
 	res.Location = "header"
-	res.ValueExpression = valueExpression
+	res.ValueExpression = golang.MangleVarName(swag.ToGoName(valueExpression))
```
casualjim: why does this need to change, it was working afaik

generator/operation.go
```diff
@@ -739,8 +743,12 @@ func (b *codeGenOpBuilder) MakeHeaderItem(receiver, paramName, indexVar, path, v
...
 	res.CollectionFormat = items.CollectionFormat
+	if res.CollectionFormat == "" && items.Items != nil {
+		// When not defined,CollectionFormat defaults to csv
+		res.CollectionFormat = "csv"
```
casualjim: not a fan of this because it now sprinkles this across the codebase and people have to remember to do this. Better to postpone this until the last responsible moment

generator/operation.go
```diff
@@ -709,6 +709,10 @@ func (b *codeGenOpBuilder) MakeHeader(receiver, name string, hdr spec.Header) (G
 	}
 
 	if hdr.Items != nil {
+		if res.CollectionFormat == "" {
+			// When not defined,CollectionFormat defaults to csv
+			res.CollectionFormat = "csv"
```
casualjim: not a fan of this because it now sprinkles this across the codebase and people have to remember to do this. Better to postpone this until the last responsible moment

generator/operation.go
fredbi: got naming problems with headers when starting generating collections. I try to gather evidence for the change (please wait)

generator/operation.go
fredbi: Need some guidance as I am not following you on this 
This was supposed to be done by spec (<= the first possible moment) and it is not.

Last possible moment is render time. Don't you think we haven't already loaded the templates enough with logics like this?

generator/template_repo.go
fredbi: Was just easier to experiment with. I am replacing it with a struct member then.

generator/template_repo.go
fredbi: ok. It jast just easier to avoid spreading too much changes everywhere. If now you agree with the change, I can make it to the Gen struct.

Another similar info not mentioned in your comments (because it's stealthier) is the choice of initialization method for default values (i.e. determining if a native initializer is used, a type conversion TYPE() construct, or calling to MarshalText())

This one should be a member of the GenXXX structs as well.

generator/operation.go
fredbi: I understand, this means handling the default in templates, everywhere we use Split/JoinByFormat
Is it what you want? Even more logics in templates?

generator/operation.go
casualjim: "" and "csv" should be functionally equivalent. The runtime, strfmt and swag functions that deal with this should equate "" and "csv".

generator/operation.go
casualjim: no Splt and JoinByFormat already deal with the empty string case as csv

fredbi: OK. Will test this, though.

generator/template_repo.go
fredbi: OK. Will do. Will add a HasSomeDefaults bool in the GenXXX structure for Params and Headers.
Will see later on if relevant to generalize this to models (there are some defaulting issues with models too... this story shall also be told)

generator/template_repo.go
fredbi: OK

generator/template_repo.go
fredbi: OK. 
BTW, this is just to generate a comment in source to more easily relate generated code to spec in nested loops. 
Do you think I should follow a specific guideline to allow some day the "generate spec" feature to benefit from this as well?

generator/formats.go
```diff
@@ -117,6 +117,7 @@ var typeMapping = map[string]string{
 	// plus the following aliases:
 	//  - "datetime" alias for the more official "date-time"
 	//  - "objectid" and "ObjectId" aliases for "bsonobjectid"
+	// TODO: aliasing does not work, and was not really intended. Best to suppress it altogether.
```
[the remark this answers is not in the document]
fredbi: Format aliasing is more complex than it seemed at first and interacts with strfmt format normalizer.
Something is wrong somewhere, I don't know exactly where, but I didn't fix it yet. Hence the TODO. 
I feel that all the fuss started with the introduction of bson support, but I am not sure really.

I remove the TODO. Anyhow, I plan to ditch this whole structure and hand it over to strfmt someday (as a first step toward real custom formats, but before plugin support)

casualjim:
yeah it's fine, it's not a critical optimization

generator/operation.go
fredbi: good catch: I left an intermediate experiment here. This is actually equivalent to the simple: swag.ToVarName().
Why this transform? Actually the idea came from issue #863 
Example for header: x-isAnOptionalHeader
With nested arrays constructs, I have an expressionValue at some point like 'x-IsAnOptionalHeader3I'
VarNaming it gives the more correct xIsAnOptionalHeader3I.
What is most important is that names are consistent with childs.

generator/operation.go
fredbi: good catch. The ToVarName() is sufficient and issue #863 fix with items remains valid.

generator/operation.go
fredbi: OK

generator/operation.go
fredbi: ok.

generator/template_repo.go
fredbi: Ok tested on home-made unit test suite

generator/template_repo.go
fredbi: Done everything here but a "clean" method to select the proper initializer for default values, which is still a shaky hack in templates. Postponed.

casualjim:
can you rebase this so I can squash the commits?

PR #1399, opened 2018-02-01
fixes writing client stream body

### LINKED ISSUE #533: Code generation error for binary stream producer, consumer
We've found a scenario on our project where the code generated response correctly assumes the payload is an io.reader but tries to reflate the response as a consumer (from client side) and incorrectly assumes an io.writer.  @casualjim and I have realized that is incorrect.

If you specify a binary stream is your response like the following example,

``` yaml
'200':
  schema:
    format: binary
```

It generates client code that correctly creates an io.Reader payload.  However, the reflation code eventually dives into ByteStream.go, and the following code tries to interpret it as an io.Writer:

``` go
func ByteStreamConsumer() Consumer {
  return ConsumerFunc(func(r io.Reader, v interface{}) error {
    wrtr, ok := v.(io.Writer)
    if !ok {
      return errors.New("ByteStreamConsumer can only deal with io.Writer")
    }

    _, err := io.Copy(wrtr, r)
    return err
  })
}
```

casualjim:
fixes #533


fredbi:
Hello @casualjim. Does this one fixe #533 ?

casualjim:
yeah it should fix that one

Issue #1404, opened 2018-02-03
New feature to allow Parameters section


jucardi:
## Problem statement

I would like to add a new feature where the `Parameters` section is allowed in `swagger:route`.

#### Background to this feature

Working on a company project, we have a custom version of the Swagger API which allow us to do calls to our microservices by reading the swagger documentation generated from our builds. (like a Postman that updates the contracts automatically by using the generated swagger).

For our Go Microservices we use `gin`  and unfortunately we were unable to tie the Request models to the router definitions since the gin methods just receive a context, and then we deserialize the gin request object into our Request Struct, which would be our `swagger:model`

The swagger documentation gets generated perfectly except for the request body model is missing.

#### Details of the feature

Add a section `Parameters` when reading a `swagger:route` that will allow to define multiple parameters using the parameters information. Each parameter declaration will start with a `+` at the first line, like declaring arrays in a YAML file with `-`

Example:
```
//  Parameters:
//  + name: request
//     in: body
//     schema: modelName
//  + name: id
//     in: path
```

casualjim:
you can look into `swagger:operation` 
alternatively you can define the parameter structs, look at them as documentation but the types are verified by the compiler.

jucardi:
Thank you @casualjim, I was not aware of `swagger:operation`. Reading the spec though, it seems that the Schemas have to be added manually with the ref, type and description? This was the example I found in the documentation

```
    // responses:
    //   '200':
    //     description: pet response
    //     schema:
    //       type: array
    //       items:
    //         "$ref": "#/definitions/pet"
    //   default:
    //     description: unexpected error
    //     schema:
    //       "$ref": "#/definitions/errorModel"
```

I was hoping to just provided a `swagger:model` name and obtain all this information directly from there.

I submitted a PR that would allow adding the parameters through `swagger:route`

https://github.com/go-swagger/go-swagger/pull/1405

I think it would be a nice addition to `swagger:route`. I'd love to hear your thoughts, hopefully you'll find it useful, if not I'll start using `swagger:operation` instead

fredbi:
@jucardi do you think this issue could be closed now?

DrummerKH:
This feature is missed in the documentation - https://goswagger.io/use/spec/route.html

deepak7737:
does anyone have idea how to write query parameters in go-swagger

PR #1405, opened 2018-02-03
Added Parameters section to swagger:route

jucardi:
Working on a company project, we have a custom version of the Swagger API which allow us to do calls to our microservices by reading the swagger documentation generated from our builds.

We use Gin for our API endpoints in our Go Microservices and unfortunately we were unable to tie the Request models to the router definitions.

I added a section in the `swagger:route` parser, so `Parameters` can be included with their properties, to easily allow adding a reference to a `swagger:model` for body params. Multiple parameters can be added under `Parameters` by simply using the `+` prefix when a new parameter starts. Example:

```
//  Create a pet based on the parameters.
//
//  Consumes:
//  - application/json
//  - application/x-protobuf
//  
//  Produces:
//  - application/json
//  - application/x-protobuf
//
//  Schemes: http, https, ws, wss
//
//  Parameters:
//  + name: request
//    in: body
//    schema: modelName
//  + name: id
//    in: path
//
//  Responses:
//    default: body:genericError
//    200: body:someResponse
//    422: body:validationError
//
//  Security:
//    api_key:
//    oauth: read, write
```


jucardi:
I'll increase the coverage in a bit

generator/bindata.go
```diff
@@ -729,34 +729,34 @@ func AssetNames() []string {
 // _bindata is a table, holding each asset generator, mapped to its name.
 var _bindata = map[string]func() (*asset, error){
```
jucardi: The changes in this file are just the result of a `gofmt -s -w -l .`, no changes in logic. If needed they could be rolled back before merging

casualjim:
have you seen swagger:operation it allows you to specify routes but with the full swagger yaml syntax. 

This PR somewhat duplicates this, but with bespoke syntax. I wonder where the type goes and some of the other validations etc that can be defined on parameters.
It's great that you figured out how to extend this :100:

casualjim:
O I hadn't seen your response on #1404 
I see you have "bigger" plans, I'm good with your remarks there would you be up for making this a bit more complete so that it covers most of the possible use cases?

jucardi:
Hi @casualjim will do, I'll cover more validations, such as only allow the use of schemas when `in:body` or allow `allowEmptyValue` to be set only if `in:query` or `in:formData`.

Was there any other use cases you had in mind?

casualjim:
once you have the ability to specify types you probably also want the ability to specify the validations they have to pass. Things min length, pattern, min/max etc come to mind.
Feel free to ping me on slack https://slackin.goswagger.io

jucardi:
the constraints, you're absolutely right, I completely forgot about them, I'll work on those as well

jucardi:
@casualjim I added a section to parse the schema in parameters, covers `max`, `min`, `maxLength`, `minLength`, `default`, `enum`, `format`, it also does type assertion when using these constraints to ensure they are assigned to the right types.

Let me know if you have any observations. Thanks!

PR #1438, opened 2018-03-04
Fixes #1437 [format validation in simple schemas]

### LINKED ISSUE #1191: Adding length to a string with `uri` format will result i unworkable generated code
## Problem statement

Adding length to a string with `uri` format will result unworkable generated code

## Swagger specification

```
swagger: '2.0'
info:
  version: '2.0.0'
  title: Thing API
  description: Some description
basePath: /thing-api
consumes:
  - application/json
produces:
  - application/json
paths:
  '/things':
    get:
      operationId: POST_things
      summary: Get a Thing
      description: Create a new thing
      responses:
        '200':
          description: Its all ok
          schema:
            type: object
            properties:
              notification_url:
                $ref: '#/definitions/notification-url'
definitions:
  notification-url:
    description: A URL to be notified about the Thing
    type: string
    maxLength: 1000
    format: uri
```

## Steps to reproduce

```
swagger generate client --target=src
```

## Symptom

The generated file `src/models/notification_url.go` will contain the following code (in which `m.String()` won't compile because `strfmt.URI` is string):

```
type NotificationURL strfmt.URI

// Validate validates this notification url
func (m NotificationURL) Validate(formats strfmt.Registry) error {
	var res []error

	if err := validate.MaxLength("", "body", string(m), 1000); err != nil {
		return err
	}

	if err := validate.FormatOf("", "body", "uri", m.String(), formats); err != nil {
		return err
	}
```

## Environment
swagger version:
version: 0.12.0
commit: 8135eb6728e43b73489e80f94426e6d387809502

go version:
go version go1.9 darwin/amd64

### LINKED ISSUE #1437: Simple aliased model type does not validate format
## Problem statement
When redefining a type with format like in:
```
definitions:
  aDate:
    type: string
    format: date
```

Posting a small PR to fix this _and only this_

The generated model for aDate has empty Validate() method
## Environment
swagger version:  current master
go version: x.x.x  
OS:

fredbi:
- fixes #1437 
- fixes #1191


generator/formats.go
```diff
@@ -197,8 +196,8 @@ var customFormatters = map[string]struct{}{
 	"strfmt.UUID3":      struct{}{},
 	"strfmt.UUID4":      struct{}{},
 	"strfmt.UUID5":      struct{}{},
-	// Special
-	"io.ReadCloser": struct{}{}, // for "format": "binary"
```
casualjim: where did these go?

fredbi: customFormatters is used to determine IsCustomFormatter and decide to run a validation.
There is no validation to be carried on these interfaces.
The type mapping from "binary" to io.ReadCloser remains above and is important.

fredbi:
I found out there is a remaining bug around the same lines, for arrays. I am pushing another commit for this

fredbi:
Technical issue on AppVeyor grid

generator/formats.go
casualjim: the io.Writer is the same but for the client, so it should also remain. 
format: binary on the server is an  io.ReadCloser but on the client it's an io.Writer

fredbi: I saw Writer was for the client. Though I did not specifically test the client, I scooped through templates to figure out other usages of IsCustomFormatter than validation

fredbi:
So I removed these entries from the map to avoid this kind of wrong generation:
```
// Code generated by go-swagger; DO NOT EDIT.

package models

// This file was generated by the swagger tool.
// Editing this file might prove futile when you re-run the swagger generate command

import (
	"io"

	strfmt "github.com/go-openapi/strfmt"

	"github.com/go-openapi/errors"
	"github.com/go-openapi/swag"
	"github.com/go-openapi/validate"
)

// AnObjectOfInterfaces an object of interfaces
// swagger:model anObjectOfInterfaces
type AnObjectOfInterfaces struct {

	// a reader
	AReader io.ReadCloser `json:"aReader,omitempty"`
}

// Validate validates this an object of interfaces
func (m *AnObjectOfInterfaces) Validate(formats strfmt.Registry) error {
	var res []error

	if err := m.validateAReader(formats); err != nil {
		// prop
		res = append(res, err)
	}

	if len(res) > 0 {
		return errors.CompositeValidationError(res...)
	}
	return nil
}

func (m *AnObjectOfInterfaces) validateAReader(formats strfmt.Registry) error {

	if swag.IsZero(m.AReader) { // not required
		return nil
	}

	if err := validate.FormatOf("aReader", "body", "binary", m.AReader.String(), formats); err != nil {
		return err
	}

	return nil
}

// MarshalBinary interface implementation
func (m *AnObjectOfInterfaces) MarshalBinary() ([]byte, error) {
	if m == nil {
		return nil, nil
	}
	return swag.WriteJSON(m)
}

// UnmarshalBinary interface implementation
func (m *AnObjectOfInterfaces) UnmarshalBinary(b []byte) error {
	var res AnObjectOfInterfaces
	if err := swag.ReadJSON(b, &res); err != nil {
		return err
	}
	*m = res
	return nil
}

```

fredbi:
At the moment, binary does not support any validation. I do not consider this a bug atm, even though purists might jump at this.

fredbi:
Hey Ivan.
I put back the entries in the map (I am still struggling to figure out why they were there in the first place, but I am not arguing...)
I extended test cases and fixed some more issues with formatting under various nesting conditions.

PR #1439, opened 2018-03-04
Improve 667 multiple validations in model

fredbi:


casualjim:
won't fix, as discussed offline

fredbi:
How about making it a generation option (e.g. available in ModelSection from yaml config)?

Issue #866, opened 2017-01-12
Generator don't generates validation for embended models


westito:
## Problem statement
When I write the model schema directly into the query (and not use separete definition for the model) the validations like maxLength, required do not generates for the model. But when I but the schama in the definition section, and $ref to it, the validations are get generated.

## Swagger specification
```yaml
responses:
        200:
          description: Success
          schema:
            required:
              - data
            properties:
              data:
                required:
                  - account_id
                  - access_token
                properties:
                  account_id:
                    type: integer
                    format: int64
                  access_token:
                    type: string
                    maxLength: 64
```

casualjim:
a work around for this is to flatten the spec before generating it

PR #1455, opened 2018-03-13
* Fix model build issue with AdditionalItems (serialization, validation)

fredbi:
* Fix AdditionalItems validation when spec is flattened
* Contributes to #866

Models with AdditionalItems now build with `swagger generate model --skip-validation ...`

Strengthened CI test suite on codegen with server, model and client generations and builds.


fredbi:
Oops! I've seen a glitch in my CI script. I am fixing this right away

fredbi:
.. and appveyor is stuck once again

Issue #1446, opened 2018-03-08
Support for optional auth?


dmjones500:
## Problem statement

I'm trying to support optional authorization for a path, but it seems like go-swagger is requiring authorization each time.

Based on https://github.com/OAI/OpenAPI-Specification/issues/14, I've create a path that includes:

```
      security:
      - {}
      - jwt: []
```

This compiles fine, but when I submit request that doesn't include my API key, I get an immediate 401 response. Is this a supported configuration?

## Environment
swagger version: 

version: 0.13.0
commit: 8135eb6728e43b73489e80f94426e6d387809502

go version: go1.9.3 darwin/amd64  

OS:   Mac OSx

fredbi:
You might be interested with analysis in #1089

casualjim:
Optional auth should now be supported

tschaub:
I'm struggling to come up with a security spec that takes advantage of the new optional auth functionality in https://github.com/go-openapi/runtime/pull/98.  Should the configuration below be supported now?

```yml
security:
  - {}
  - SomeOtherAuthenticationType
```

fredbi:
Have you taken a look at the example [here](https://github.com/go-swagger/go-swagger/tree/master/examples/composed-auth)? Several combinations are examplified there.

tschaub:
I've created go-openapi/runtime#103 in hopes of getting optional authentication support with a spec like the one above.

tschaub:
Thanks @fredbi - I did look through the docs on [composed security requirements](https://github.com/go-swagger/go-swagger/tree/master/examples/composed-auth) - they cover some nice use cases.  Unless I'm misreading, I don't think they consider the "optional authentication" use case.

It looks like the changes in `go-openapi/runtime#98` were intended to support optional authentication.  And there is a test case that covers this (see https://github.com/go-openapi/runtime/pull/98/files#diff-766ab5a2cb406b60ad239ca4c3e484c9R194).  However, the tests passes with a route authenticator that looks like this:

```go
ra := RouteAuthenticator {
	allowAnonymous: true,
}
``

And this is different than the shape of a route authenticator that is built from the spec above (with `security: [{}, ...]`).  This spec generates a route authenticator that looks something like this:

```go
ra := RouteAuthenticator {
	allowAnonymous: true,
	Schemes: []string{""},
	# ...
}
```

fredbi:
How goes your spec example through the other toolkit packages, e.g. spec, analysis an validate?
[...]

Issue #1281, opened 2017-11-27
embedded spec discards information from original spec


databus23:
## Problem statement
Given this spec
```
swagger: '2.0'
consumes:
  - application/json
produces:
  - application/json
info:
  title: Test
  version: 1.0.0
securityDefinitions:
  keystone:
    type: apiKey
    in: header
    name: x-auth-token
security:
  - keystone: []
paths:
  /info:
    get:
      security: []
      responses:
        '200':
          description: OK
          schema:
            type: string

```

The embedded spec in the generated server is missing the `security: []` entry in the operation.
This is semantically different from the original spec where the global `security` setting is explicitly overridden to be empty (no authentication for this operation required).

The problem is caused by an `omitempty` [struct tag](https://github.com/go-openapi/spec/blob/a4fa9574c7aa73b2fc54e251eb9524d0482bb592/operation.go#L34). I'll open a PR there to fix this issue.

## Steps to reproduce
`swagger generate server --name test --target test swagger.yaml`

## Environment
swagger version: 0.13.0  
go version: 1.9.2
OS: Darwin

gregmarr:
@databus23 Is there a possibility that this will cause the `security: []` to appear on specs that don't currently have anything there, changing their semantics?

databus23:
no, what will appear there is `security: null` which is semantically the same as before.

gaplyk:
this changes cause an error in validator.

gaplyk:
@databus23  please reconsider your approach to fix server generator instead of fixing spec.
I can't fix spec generator to generate json files without "security": null

gregmarr:
@gaplyk That's expected according to @databus23 's post that it's semantically the same as before.  Can you provide information on why that's an error for you?

databus23:
@gaplyk Can you elaborate a bit on what the problem is? what is the problem with `security: null` in the spec generator?
I'm not sure what other ways are they to fix this. Leaving the `omiteempty` tag annotation in go-openapi/spec does not seem right to me as it causes loss of information when serialising a `spec.SwaggerSpec` struct to JSON.

casualjim:
if semantically the same how is omitempty an issue? can it be fixed with a custom marshal/unmarshal?

gregmarr:
He's saying that `security: null` is semantically the same as `omitempty`, but different than `security: []`.

databus23:
[...]

Issue #1214, opened 2017-10-15
Security definition order is not consistent between code generations


johnnyg:
## Problem statement

When we have multiple security definitions, the order they are generated in isn't consistent between runs.

## Swagger specification

`swagger: "2.0"
info:
  version: 0.0.0.0
  title: Test swagger spec for 1214
securityDefinitions:
  header:
    type: apiKey
    name: secret
    in: header
  query:
    type: apiKey
    name: secret
    in: query
security: ["header": [], "query": []]
paths:
  /foo:
    get:
      summary: "Add a new pet to the store"
      description: ""
      responses:
        200:
          description: "Success"`

## Steps to reproduce

Run swagger gen with above spec multiple times.

## Environment
swagger version: 0.12.0  
go version: 1.9.1  
OS: Linux

fredbi:
SecurityDefinitions are naturally stored as a map in structs.go
but it is actually used as a slice in regular ranges by templates, so it may be possible to convert all these maps into slices in the generation structures and sort them.

At the moment, that makes quite some extra work. Is there any genuine added value beyond stable generated output?

gregmarr:
Don't underestimate the value of stable generated output. It makes debugging generation errors much easier. We currently sort model and operation arrays after extracting then from the maps and before iterating over them.

fredbi:
Hi Greg. I don't underestimate. We have had this discussion and you know I am versionning everything...
This is precisely why I am investigating this one.
I already identified a quick improvement to sort AllOf schemas, that's a really easy one.
This one is a bit trickier but quite doable.
Problem is regression testing with this as I only have few working examples at my place to really test it's working. Testing is making 90% of my contribution time ratio now that I got into templates.

gregmarr:
Okay. Just wanted to say that I think stable generated output is enough genuine added value and that you don't have to look for more.

gregmarr:
Now that I read the description again, I'm even more convinced as this is about observable order in the output and not just the order in which files are generated.

fredbi:
Yes. It is just one of the few map structures left in GenXXX structs. That is what is making it random.

Issue #1089, opened 2017-07-03
Security only supports "OR" logic, not "AND"


gavbaa:
## Problem statement
When specifying a security block in the AND format (described here: https://swagger.io/docs/specification/authentication/ ), having only one pass allows requests to continue without a 401.

The documentation states:
```
security:
  - A
  - B

 # A OR B
security:
  - A
    B

 # A AND B
```

## Swagger specification
I've included the relevant bit here.  I've put a full test case at https://github.com/gavbaa/go-swagger-auth-test if you want to clone and review directly.  It's the TodoList Swagger file, with only the following security bit added.
```
securityDefinitions:
  AppTokenAuth:
    type: apiKey
    in: header
    name: App-Token
  TokenAuth:
    type: apiKey
    in: header
    name: Token

security:
  - AppTokenAuth: []
    TokenAuth: []
```
## Steps to reproduce
I implemented only the AppTokenAuthAuth function and left TokenAuthAuth returning an error:
```
	api.AppTokenAuthAuth = func(token string) (interface{}, error) {
		return true, nil
		//return nil, errors.NotImplemented("api key auth (AppTokenAuth) App-Token from header param [App-Token] has not yet been implemented")
	}

	// Applies when the "Token" header is set
	api.TokenAuthAuth = func(token string) (interface{}, error) {
		return nil, errors.NotImplemented("api key auth (TokenAuth) Token from header param [Token] has not yet been implemented")
	}
```

However, upon cURLing with just an App-Token, it moves onto the method handler instead of rejecting the request with a 401.

```
➜  ~ curl -H "App-Token: x" "http://127.0.0.1:58282"
"operation todos.Get has not yet been implemented"
```

## Environment
swagger version: 2.0.0 
go version: 1.8.1
OS:  OSX 10.12.5

mmmguitar:
New to go-swagger but I've also run into this issue and had a quick look into the code.  Long story short, the problem ultimately is with the go-openapi router + middleware collapsing the path/route security []map into just an [] of unique requirements, go-swagger then implements this / has to. 

The go-openapi which go-swagger depends on defines a struct model for a swagger spec, take a look at spec.Operation which has a spec.OperationProps which seems to define the properties on the route path.

spec.OperationProps defines security prop as an "[]map[string][]string"

This matches with the swagger spec / documentation, i.e there are a list of acceptable maps of security definitions, the list is the OR part and the map is the AND part.

[...]

PR #1478, opened 2018-04-15
Allow AND and Optional auth expressions

casualjim:
contributes to:
* #1446
* #1281
* #1214
* #1089


Gopkg.lock
```diff
@@ -44,10 +44,10 @@
   version = "v1.4.7"
 
 [[projects]]
-  branch = "master"
+  branch = "auth-fixes"
```
fredbi: Do not forget to revert to master once the final integration is done

Gopkg.lock
```diff
@@ -83,7 +83,7 @@
   revision = "2a2b323bab96e6b1fdee110e57d959322446e9c9"
 
 [[projects]]
-  branch = "master"
+  branch = "auth-fix"
```
fredbi: same remark as above

Gopkg.toml
```diff
@@ -22,7 +22,7 @@
 
 
 [[constraint]]
-  branch = "master"
+  branch = "auth-fixes"
```
fredbi: same as above

Gopkg.toml
```diff
@@ -38,7 +38,7 @@
   name = "github.com/go-openapi/loads"
 
 [[constraint]]
-  branch = "master"
+  branch = "auth-fix"
```
fredbi: same

generator/operation.go
```diff
@@ -271,7 +271,7 @@ type codeGenOpBuilder struct {
 	DefaultScheme       string
 	DefaultProduces     string
 	DefaultConsumes     string
-	Security            []analysis.SecurityRequirement
+	Security            [][]analysis.SecurityRequirement
```
fredbi: big contribution toward #1214 , but not solved yet: since the original spec structure is a map, the inner [] is still random and need sorting here.

casualjim: requirements aren't a map, the order is stable there. The only time a map structure is used is when describing a requirement is as a set for deduping but the items should be appended in a consistent order.

fredbi: I was referring to the OperationProps.Security in spec (https://github.com/go-openapi/spec/blob/bcff419492eeeb01f76e77d2ebc714dc97b607f5/operation.go#L34). This is an array of maps.

fredbi: Same in https://github.com/go-openapi/spec/blob/bcff419492eeeb01f76e77d2ebc714dc97b607f5/swagger.go#L84

fredbi: that means the requirements with "AND" (keys) come in any order right?

PR #1489, opened 2018-04-19
pin go-openapi/{analysis,spec,validate} versions

ags799:
Was using go-swagger 0.13.0 as a library until I noticed a compilation error between go-swagger and go-openapi/spec. I fixed it by specifying a commit for the go-openapi/spec dependency.

This commit is supposed to be for the 0.13.0 tag (commit b015bd). I couldn't open a PR unless I made the commit against master (I'm not that experienced with GitHub).

If this isn't merged, the workaround is for my client project to have
```toml
[[override]]
  name = "github.com/go-openapi/spec"
  revision = "a4fa9574c7aa73b2fc54e251eb9524d0482bb592"
```
in addition to its
```toml
[[constraint]]
  name = "github.com/go-swagger/go-swagger"
  version = "0.13.0"
```
in its `Gopkg.toml`. I'd much rather not need that `[[override]]` block (I'd much rather merge this PR).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -44,7 +44,6 @@
   version = "v1.4.2"
 
 [[projects]]
-  branch = "master"
   name = "github.com/go-openapi/analysis"
   packages = ["."]
   revision = "8ed83f2ea9f00f945516462951a288eaa68bf0d6"
@@ -115,10 +114,9 @@
   revision = "cf0bdb963811675a4d7e74901cefc7411a1df939"
 
 [[projects]]
-  branch = "master"
   name = "github.com/go-openapi/validate"
   packages = ["."]
-  revision = "a52193aca9d575f354d8be388254c134765ea793"
+  revision = "a762d5dd38f7c7e9e1a13c3e98344cfbe4aa15d9"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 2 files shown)


ags799:
Just want to be clear, I understand there are conflicts here. I'd like to create a `0.13.1` branch from the `0.13.0` tag, merge this PR into that new branch, and then release a new `0.13.1`.

But perhaps you all have other ideas.

casualjim:
How come you are depending on go-swagger/go-swagger ?

ags799:
@casualjim I'm working with APIs defined in [JSON Hyper Schema](http://json-schema.org/). Wish I was using Swagger 😢.

I'd still like to use go-swagger's code generation. So I've wrapped go-swagger with another program that converts JSON Hyper Schema into Swagger, then uses go-swagger to generate code from that.

Also, it looks like I need to specify revisions for the other go-openapi projects used by Swagger. More commits on the way.

JonKohler:
Where are we with this PR? This seems like a good idea overall.

ags799:
I think it needs an approving review? I don't use go-swagger anymore so idc.

fredbi:
I think it would be wiser to tag the goopenapi repos as releases

PR #1514, opened 2018-05-04
fix writing enum parameters (#1513)

### LINKED ISSUE #1513: Client parameters don't set body when model is an enum
## Problem statement

If the client body is an enum, the body doesnt get serialised by WriteToRequest.

## Swagger specification

```yaml
swagger: '2.0'
info:
  version: "1.0.0"
  title: "Test enum serialisation"
paths:

  /example:
    put:
      parameters:
      - name: "something"
        in: "body"
        description: "PUT enum example"
        required: true
        schema:
          $ref: '#/definitions/MyEnum'
      responses:
        202:
          description: "PUT OK"
        500:
          description: "Error"

definitions:

  MyEnum:
    type: "string"
    enum:
    - SOMEVALUE
    - SOMEOTHERVALUE
```

## Steps to reproduce

`swagger generate client -f ./swagger.yaml`

Note that put_example_responses.WriteToRequest() never calls r.SetBodyParam().

## Environment
swagger version: master
go version: 1.10.2 
OS:  osx

## Fix
Looks like the following in the template needs to include enums:
```
  {{ if or .Schema.IsInterface .Schema.IsStream (and .Schema.IsArray .Child) (and .Schema.IsNullable (not .HasDiscriminator)) }}if {{ .ValueExpression }} != nil {
```

stevecookform3:
- fixes #1513

### DIFF
--- a/generator/parameter_test.go
+++ b/generator/parameter_test.go
@@ -1705,3 +1705,26 @@ func TestGenParameter_Issue1392(t *testing.T) {
 		}
 	}
 }
+
+func TestGenParameter_Issue1513(t *testing.T) {
+	assert := assert.New(t)
+	var assertion = `r.SetBodyParam(o.Something)`
+	gen, err := opBuilderWithFlatten("put-enum", "../fixtures/bugs/1513/enums.yaml")
+	if assert.NoError(err) {
+		op, err := gen.MakeOperation()
+		if assert.NoError(err) {
+			buf := bytes.NewBuffer(nil)
+			opts := opts()
+			err := templates.MustGet("clientParameter").Execute(buf, op)
+			if assert.NoError(err) {
+				ff, err := opts.LanguageOpts.FormatContent("move_clusters_parameters.go", buf.Bytes())
+				if assert.NoError(err) {
+					res := string(ff)
+					assertInCode(t, assertion, res)
+				} else {
+					fmt.Println(buf.String())
+				}
+			}
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


generator/parameter_test.go
```diff
@@ -1705,3 +1705,26 @@ func TestGenParameter_Issue1392(t *testing.T) {
...
+func TestGenParameter_Issue1513(t *testing.T) {
+	assert := assert.New(t)
+	var assertion = `r.SetBodyParam(o.Something)`
+	gen, err := opBuilderWithFlatten("put-enum", "../fixtures/bugs/1513/enums.yaml")
+	if assert.NoError(err) {
+		op, err := gen.MakeOperation()
+		if assert.NoError(err) {
+			buf := bytes.NewBuffer(nil)
+			opts := opts()
+			err := templates.MustGet("clientParameter").Execute(buf, op)
```
stevecookform3: I copied this block from an earlier test case, but not that clear what its doing. Is this valid for checking the generation code?

fredbi: These lines are just to sanity check that the template executes correctly;
Unfortunately we don't have a real integration test on the CI.
We just run codegen asserts (you have a lot of that in this test file). They are here to validate that the codegen looks as expected, but not that it actually works...

fredbi:
Oh blast! The appveyor CI seems to be again stalled...

fredbi:
@stevecookform3 
looks like appveyor is (again! it occurs from time to time) stalled.  Could you please kick him by pushing the PR again (e.g. git push --force upstream) ?

stevecookform3:
@fredbi dont seem to be able to git push upstream directly, but i've pushed an empty commit to my fork which should kick it off again

fredbi:
OK. The way I am doing it is to git push --force upstream to my fork. This wakes up everything.

PR #1521, opened 2018-05-10
Fix complex anonymous additional properties

fredbi:
- fixes codegen on anonymous complex additional properties (e.g. with --skip-flatten)
- extends UT in expand mode


fredbi:
CI checks are blocking because of untested error paths.
However, this is impossible to test, since the only cause for error is an unresolved $ref, which causes a panic in spec expander.

fredbi:
Thanks Ivan.
Would you like that I remove the panic() in spec/expand()?
I made a ", err != nil { return err}" explorer testcase that improve test coverage a bit...

Issue #1497, opened 2018-04-23
Not possible to hide individual struct fields


neilgarb:
## Problem statement

I have a struct which is exposed by my API, but which I want to leave undocumented/unspec'ed. There is no way to exclude individual struct fields from the spec if they're exported and have a `json:""` tag.

## Swagger specification

N/A, because I'd like to hide the field from the generated spec altogether.

## Environment
swagger version: 2.0
go version: 1.10.1
OS:   macOS

## Motivation

My API needs to remain backwards-compatible. When I add functionality which I don't want to support indefinitely, I leave it undocumented.

## Proposed solution

Allow fields to be hidden by adding `hidden: true` to their comment, e.g.

```go
// swagger:model Foo
type Foo struct {
    // Bar is an example field which won't be included in the generated spec.
    //
    // hidden: true
    Bar string `json:"bar"`
}
```

inhuman:
> It will be even more awesome if it can be differentiated with "// swagger:ignoreParam" and "// swagger:ignoreResponse".

It would be very useful

yujintang:
type Foo struct {
    Bar string `json:"-"`
}
tags using `json:"-"`can solve this problem

casualjim:
You can use `// swagger:ignore`

casualjim:
You can achieve more finer grained granularity by making a struct with the common fields and embedding that

waffle-frame:
> You can use `// swagger:ignore`

Thanks, this works for me

PR #1527, opened 2018-05-13
Allow struct fields to be annotated with swagger:ignore

neilgarb:
Sometimes it's useful to includ a field in a request or response but leave it
undocumented. This change allows such struct fields to be annotated with
`swagger:ignore`, e.g.

```go
type Foo struct {
  // swagger:ignore
  Bar string `json:"bar"`
}
```

See #1497.

Signed-off-by: Neil Garb <neil@luno.com>


mosleim:
@casualjim is it possible to ignore only on param?

casualjim:
Use `json:"-"`

PR #1533, opened 2018-05-16
Ordering security definitions and requirements

### LINKED ISSUE #1214: Security definition order is not consistent between code generations

fredbi:
- fixes #1214 (stable codegen with security definitions)
- also sorts inner (joint) security requirements and associated scopes (although not currently used in templates)
- temporary work around for mishandling empty sec requirements in analysis (see https://github.com/go-openapi/analysis/pulls/29)


generator/operation.go
```diff
@@ -1076,3 +1075,40 @@ func (b *codeGenOpBuilder) setBodyParamValidation(p *GenParameter) {
...
+// makeSecuritySchemes produces a sorted list of security requirements for this operation.
+// As for current, these requirements are not used by codegen (sec. requirement is determined at runtime).
+// We keep the order of the slice from the original spec, but sort the inner slice which comes from a map,
+// as well as the map of scopes.
+func (b *codeGenOpBuilder) makeSecurityRequirements(receiver string) []GenSecurityRequirements {
+	if b.Security == nil {
+		if b.Doc.Spec().Security == nil && b.Operation.Security == nil {
+			// work around bug in analysis.SecurityRequirementsFor()
+			// TODO: remove second check this when fixed in analysis
+			// nil (default requirement) is different than [] (no requirement)
```
casualjim: The changes in analysis were merged, you can remove this

generator/structs.go
```diff
@@ -490,13 +489,13 @@ type GenApp struct {
...
-	SecurityDefinitions []GenSecurityScheme
+	SecurityDefinitions GenSecuritySchemes
 	Models              []GenDefinition
 	Operations          GenOperations
 	OperationGroups     GenOperationGroups
 	SwaggerJSON         string
-	// this is important for when the generated server adds routes
-	// ideally this should be removed after we code-generate the router instead of relying on runtime
+	// This is important for when the generated server adds routes.
+	// TODO: ideally this should be removed after we code-generate the router instead of relying on runtime
```
casualjim: There is a very distinct advantage to having that in runtime instead of as generated code.  So I don't think we should ever code generate the router.
If embedding the spec is an issue then it's probably a better solution to use a middleware that loads the files from some other medium (like disk, cdn, ...) and serves that as a static middleware and not embed the specs with `--exclude-spec`

generator/operation.go
fredbi: Yes: that was the plan...

I would like that we update vendor beforehand. I can take care of that. Would you like to merge the pending PR in spec before that?

PR #1538, opened 2018-05-18
Fix #1532

### LINKED ISSUE #1532: Allof with a polymorphic type can have conflicting field names
## Problem statement

It appears that having a polymorphic type containing a field named "data" included in an allOf definition causes an issue in the generated code, specifically in the UnmarshalJSON function:

```
func (m *UpdateObjectAllOf1) UnmarshalJSON(raw []byte) error {
	var data struct { // <-- here, "data" is a struct
		Data json.RawMessage `json:"data,omitempty"`
	}
	buf := bytes.NewBuffer(raw)
	dec := json.NewDecoder(buf)
	dec.UseNumber()

	if err := dec.Decode(&data); err != nil {
		return err
	}

	data, err := UnmarshalObject(bytes.NewBuffer(data.Data), runtime.JSONConsumer()) // <-- but here, UnmarshalObject returns something satisfying the "Object" interface, which the original "data" does not
```

I think this could be fixed by adding `Field` or something similar to the `{{ camelize .Name }}` here, if I'm understanding the code correctly: https://github.com/go-swagger/go-swagger/blob/master/generator/templates/tupleserializer.gotmpl#L102

## Swagger specification
```
definitions:
  Event:
    type: object
    discriminator: event
    properties:
      event:
        type: "string"
        enum:
          - update_object
          - delete
      id:
        type: string

    required:
      - event
      - id

  update_object:
    type: object
    allOf:
      - $ref: "#/definitions/Event"
      - type: object
        properties:
          data: # <--- This is what produces an issue
            $ref: "#/definitions/Object"

  Object:
    type: object
    discriminator: type
    properties:
      type:
        type: string
        enum:
          - Message
          - User
    required:
      - type

```

## Environment
swagger version: 6bc0267ab287a519ac9db319593ed5c0c32dce10
go version: 1.10
OS:   OS X 10.12

fredbi:
Dedupe var when conflicting with "data", "buf" or "dec" in unmarshaller for polymorphic types.


generator/templates/tupleserializer.gotmpl
```diff
@@ -62,6 +62,10 @@ func ({{.ReceiverName}} {{ pascalize .Name }}) MarshalJSON() ([]byte, error) {
 }
 {{ end }}
 
+{{- define "dedupe" }}{{/* add a prefix if name conflicting with a local variable */}}
```
casualjim: this should probably move to a function instead a template? Or even be generated an part of the GenXxx struct that describes this data

fredbi: Yeah I pondered between these 3... The dedupe is very "local", so eventually I put it close to where it is used...

fredbi:
Ok now in FuncMap

fredbi:
This one was intended as a quick fix only, with lowest impact on UT's expectations.
I think I'll close this one  now and provide a better fix, with heavier impact on UTs...

Issue #1545, opened 2018-05-24
swagger:parameters won't generate in spec when inside a function


josephspurrier:
## Problem statement

If a swagger:parameters annotation and struct is inside of a function, then it is not picked up by the generator. This would be really helpful because they it would cut down on the naming of the structs and I could use the same `request` struct name in every function.

## Swagger specification

2.1

## Steps to reproduce

Create a function and put the swagger:parameters inside of it. When you perform a `swagger generate spec -o ./swagger.json`, the parameters for the endpoint will not be in the spec. If I move the annotation and the struct out of the function, it works fine.

```go
// Post .
// swagger:route POST /v1/game game GamePost
//
// Add the game to the database.
//
// Responses:
//   201: CreatedResponse
//   400: BadRequestResponse
//   401: UnauthorizedResponse
//   500: InternalServerErrorResponse
func (p *Game) Post(c echo.Context) (err error) {
	// swagger:parameters GamePost
	type request struct {
		// in: body
		Body struct {
			StartAt string   `json:"start_at" form:"start_at" validate:"required"`
			EndAt   string   `json:"end_at" form:"end_at" validate:"required"`
			Winners []string `json:"winners" form:"winners" validate:"required"`
		}
	}
	...        
}
```

## Environment
swagger version: dev
go version: go1.10.1 darwin/amd64
OS:   MacOSX

josephspurrier:
It looks like this is because of parser.ParseFile() in util.go - it looks like the standard Go parser only uses objects in the package scope. Is there a reason why the generation only works when the are on objects in the package scope?

casualjim:
no specific reason, feel free to extend that to your needs. We rarely reject PR's

josephspurrier:
I was wrong, it didn't have anything to do with the standard Go parser, I just had to make changes to the Parse function for the parameters. There are still other cases that could be handled from this function, but for the time being, this should suffice. Thanks! Wonderful tool by the way - really awesome!

PR #1550, opened 2018-05-25
Allow parameter code generation when struct is inside a function.

josephspurrier:
Also removes the extra log left in the repo. Closes https://github.com/go-swagger/go-swagger/issues/1545.


josephspurrier:
I'll add some tests so the code coverage doesn't go down. I'm going to add in similar code for the responses as well.

casualjim:
excellent! I'll wait with merging then?

josephspurrier:
Yeah, give me a couple days and I'll update the pull request. Thanks!

josephspurrier:
It should be good to go. Let me know if I need to make any changes.

PR #1566, opened 2018-05-31
Render and validate maps in body parameters

fredbi:
## Render and validate maps in body parameters
- fix incorrect rendering of maps in body param (i.e systematic `ìnterface{}` type)
- validate maps (e.g. schema AdditionalProperties) in body parameter
- render simple maps (e.g. `map[string][]map[string][][][]int64`) as inlined types
- maps of schema does not create a schema of their own, just the complex part (e.g. `map[string][][]map[string]model.MyModel`)

### Fixes

- fixes #1536
- fixes #1562

### Summary of changes

- `operation.go`:**refactoring + new feature/fix**
  - refactored body param processing
  - treat anonymous maps as Child GenParameter structures (such as `[]map[string][][][]map[string]*model.MyMode`)
  - extended body param validation strategy selection to maps
  - fix: enums are considered slice validations, including for bod params
  
- `model.go`: **targeted fixes**
  - refactored slice validations
  - fix missing lift validation case for maps
  - align nullable status for maps of primitives or arrays of primitives (#1562)
  
- `types.go`: **cosmetic**
  - more comments in code (e.g. special nullability rules)
  - outsource debug func
  
- `structs.go`: **enrichments**
  - removed unused and unassigned `BodyParam` struct
  - enriched GenParameters to support map constructs
  
- `templates/server/parameter.gotmpl`: **refactoring + new feature/fix**
   - extended body param section to support maps as well as maps, with or without validations
   - large refactoring of nested validations to merge slice and map validations
   - follow guards as in models (IsZero | Required)
   
### Known issues

Known existing issues not resolved here:

1. fully supporting `--skip-flatten` (e.g. anonymous types + extra schemas in the operation package)
2. undue definitions created by flatten on maps of objects (non blocking)
3. incorrect validations of Zero Values in slices and maps in some cases (e.g. skipping validation with IsZero())
4. depending on the future solution to replace/remove IsZero|Required for slice and map elements, the nullability re-alignment above 
may be easily modified and change nullable priorities
5. Nice to have: nested array validations with slice validations should execute slice validation before, not after validating all elements

PR #1568, opened 2018-06-05
Anonymous body parameters and responses

### LINKED ISSUE #1559: Broken code is generated with --skip-flatten
## Problem statement

It seems swagger version 0.14 has an issue in the server and client code generation if --skip-flatten parameter is used. It generates models but those models are not used in the responses. Instead there is an attempt to refer the models namespace but with a struct name that doesn't exist in there. There is a struct defined under that name in restapi/operations/.../directory.

After code is generated using the spec bellow take a look at the following files:
restapi/operations/object/get_objects_responses.go
restapi/operations/object/get_objects.go

The first file has the following code:
```
/*GetObjectsOK Objects

swagger:response getObjectsOK
*/
type GetObjectsOK struct {

    /*
      In: Body
    */
    Payload []*models.GetObjectsOKBodyItems0 `json:"body,omitempty"`
}
```

The second one has declaration of GetObjectsOKBodyItems0:
```
// GetObjectsOKBodyItems0 get objects o k body items0
// swagger:model GetObjectsOKBodyItems0
type GetObjectsOKBodyItems0 struct {

    // Object description.
    // Required: true
    Description *string `json:"description"`

    // Object ID
    // Read Only: true
    // Format: uuid4
    ID strfmt.UUID4 `json:"id,omitempty"`

    // Object name.
    // Read Only: true
    Name string `json:"name,omitempty"`
}
```

The models/object.go file was also created but I don't see where that package is used in the generated code. The content of the Object struct from that file basically a duplicate of GetObjectsOKBodyItems0:
```
// Object object
// swagger:model Object
type Object struct {

    // Object description.
    // Required: true
    Description *string `json:"description"`

    // Object ID
    // Read Only: true
    // Format: uuid4
    ID strfmt.UUID4 `json:"id,omitempty"`

    // Object name.
    // Read Only: true
    Name string `json:"name,omitempty"`
}
```
 
## Swagger specification
```
swagger: "2.0"
info:
  title: Test API
  description: Proof of bug Test API
  version: 2.0.0
schemes:
  - https
basePath: /
consumes:
  - application/json
produces:
  - application/json
paths:
  "/objects":
    get:
      summary: Objects
      description: Returns list of all objects
      operationId: getObjects
      tags:
        - Object
      responses:
        "200":
          description: Objects
          schema:
            type: array
            items:
              $ref: "#/definitions/Object"
        default:
          $ref: "#/responses/default"
  "/objects/{id}":
    parameters:
      - name: id
        description: Object ID.
        in: path
        required: true
        type: string
        format: uuid4
    get:
      summary: Object details
      description: Returns object identified by the provided ID.
      operationId: getObjectById
      tags:
        - Object
      responses:
        "200":
          description: Object details
          schema:
            $ref: "#/definitions/Object"
        default:
          $ref: "#/responses/default"
responses:
  default:
    description: Unexpected error
    schema:
      $ref: '#/definitions/Error'
definitions:
  Error:
    properties:
      errorCode:
        type: integer
      errorMessage:
        type: string
      moreInfo:
        type: string
  Object:
    type: object
    required:
      - description
    properties:
      id:
        type: string
        format: uuid4
        description: Object ID
        readOnly: true
      name:
        type: string
        description: Object name.
        readOnly: true
        x-nullable: false
      description:
        type: string
        x-nullable: true
        description: Object description.
```

## Steps to reproduce
swagger generate server --target ./ -name test --spec swagger.yaml --skip-flatten

## Environment
swagger version: 0.14.0  
go version: 1.10.2  
OS:   Linux

fredbi:
* Fixes #1559 (broken codegen with --skip-flatten`
* Refactors model.go and operation.go: reduce code repetition and boiler plate with validations
* Enables model validation codegen on extra schemas in operations
* Body parameters and responses work with full expand (--skip-flatten)
* Models now supports producing extra schema with references to another package (`type.go`, change
to `typeresolver`)

> This is currently validated by UT only: since the current spec pre-processing 
> only leaves the choice between full flattening (no extra schemas possible) or
> full expansion (everything is an anonymous extra schema), there is no CLI
> generation option as for current which leaves complex anonymous structures
> with inner references to model definitions.
> This is thus a preliminary work to support more nuanced spec pre-processing scenarios.

* fixes enum slice validation for params

> **NOTE**: I tried to minimize conflicts with PR #1566, but they are unavoidable.
> Please merge in any order, I'll take care of the rebasing.


akhobov:
I gave this PR a quick test and found out it vastly improved the generated code!

However I saw that the `--skip-flatten` option still produces models that are not used anywhere in operations. Is that intended behaviour?

I added a simple spec in https://github.com/akhobov/go-swagger/commit/1cd9e2f2d7974a511d40dfc72216eb85c0ac940c that I used for my testing. If you generate a server using that spec you would see that each model in completely independent. For example:
* The SimpleObject model will have its own copy of the Data struct in the same file that is called SimpleObjectData even thought there is Data model defined in the file next to it.
* The FullObject model will again have a copy of Data struct defined as FullObjectAO0Data while the rest of the properties is a combined set of both SimpleObject and FullObject. Is that what the "expand all" functionality means?
* The *_responses.go files in restapi/operations directory have their own copies of both SimpleObject and FullObject defined as GetSimpleOKBodyItems0 and GetFullOKBody respectively with the Data property being GetSimpleOKBodyItems0Data and GetFullOKBodyAO0Data.

Overall I would say that the "flatten" mode now generates more predictable models that are actually used in operations. The only thing that I noticed when generated server using the same spec is the FullObject is represented as a composition of SimpleObject and FullObjectAllOf1. Is that what the result of the functionality that @fredbi referred in point 4 of #1559?

fredbi:
Thanks for cross checking this contribution.

Regarding the generation of non-used models, I would not say it is a _desired_ functionality, but yes, it is expected: when running with `--skip-flatten`, you spec is actually fully expanded before generation (yes, the option name is misleading, it should say: `--with-expand` instead).

This means that your definitions remain here and generate models, but that all operations get their own structure from the expansion (the `$ref` relationship is broken by expansion). We call these "anonymous structures". This is why they have their own copy.

Overall, fully expanding the spec is usually not what a user needs. It is here, it must be supported and so now it works. I see flattening and expanding as two extreme cases for spec $ref preprocessing: with full flattening, you get _no_ anonymous structure (so everything becomes a definition in models), while with expansion, _everything_ is anonymous, and any generated model (original definitions) gets unused.

This is why I am saying this is preparatory work: everything has to work in both cases if we want to make some progress with `flatten` and produce more nuanced (and more useful) options.

akhobov:
Thank you for the explanation. That was really helpful! I'm with you now.

Would be great if that explanation is added to the documentation. At the moment docs are missing --skip-flatten option completely, not mentioning the difference in model/operation generation process when this option is used.

fredbi:
Model doc WIP [in this branch](https://github.com/fredbi/go-swagger/tree/doc-additional-in-model/docs/use/schemas). Feel free to comment, PR, etc.. on this fork.

generator/parameter_test.go
```diff
@@ -2934,9 +2945,255 @@ func TestGenParameter_Issue1536_Maps(t *testing.T) {
...
 
+func TestGenParameter_Issue1536_MapsWithExpand(t *testing.T) {
+	log.SetOutput(ioutil.Discard)
+	defer func() {
+		log.SetOutput(os.Stdout)
+	}()
+
+	fixtureConfig := map[string]map[string][]string{
+		// load expectations for parameters in operation get_map_of_array_of_map_parameters.go
+		"getMapOfArrayOfMap": map[string][]string{ // fixture index
```
golangci: File is not gofmt-ed with -s

generator/response_test.go
```diff
@@ -515,3 +518,70 @@ func TestGenResponses_Issue1013(t *testing.T) {
...
+
+func TestGenResponse_15362_SkipFlatten(t *testing.T) {
+	log.SetOutput(ioutil.Discard)
+	defer func() {
+		log.SetOutput(os.Stdout)
+	}()
+
+	fixtureConfig := map[string]map[string][]string{
+		// load expectations for parameters in operation get_nested_required_responses.go
+		"getNestedRequired": map[string][]string{ // fixture index
```
golangci: File is not gofmt-ed with -s

generator/response_test.go
```diff
@@ -515,3 +518,70 @@ func TestGenResponses_Issue1013(t *testing.T) {
...
+				`	if err := validate.Required("pkcs", "body", o.Pkcs); err != nil {`,
+			},
+		},
+	}
+
+	assert := assert.New(t)
+	for fixtureIndex, fixtureContents := range fixtureConfig {
+		fixtureSpec := "fixture-1536-2-responses.yaml"
+		gen, err := opBuilder(fixtureIndex, filepath.Join("..", "fixtures", "bugs", "1536", fixtureSpec))
+		if assert.NoError(err) {
```
golangci: 562-585 lines are duplicate of `generator/parameter_test.go:1201-1224`

generator/operation_test.go
```diff
@@ -310,6 +414,57 @@ func opBuilderWithFlatten(name, fname string) (codeGenOpBuilder, error) {
 	}, nil
 }
 
+// opBuilderWithExpand prepares the making of an operation with spec expansion prior to rendering
+func opBuilderWithExpand(name, fname string) (codeGenOpBuilder, error) {
```
golangci: 418-465 lines are duplicate of `generator/operation_test.go:368-415`

PR #1494, opened 2018-04-21
Replace tylerb/graceful with inbuilt methods Go 1.8 onwards

tortuoise:
For simplicity, examples not re-generated in this commit.

PR #1574, opened 2018-06-13
Replace tylerb/graceful with inbuilt methods Go 1.8 onwards - new request

tortuoise:
New request in place of #1494 which will be closed.


casualjim:
I think your templates are using spaces instead of tabs or something, the indentation is off compared to the rest of the file

PR #1577, opened 2018-06-15
Update README.md to add Nutanix to the list

JonKohler:
Added Nutanix to the list of companies using go-swagger. 

We're using this for swagger validate internally to ratchet up the quality gate for our various internal swagger.json/swagger.yaml implementations, as well as using swagger generate for nusights, which is one of our telemetry frameworks.


fredbi:
Thank you. This section of the readme if precisely here for this.

PR #1578, opened 2018-06-15
Create .golangci.yml

JonKohler:
Initial commit for .golangci.yml - this will control the behavior of the new golangCI integration here. I highly suspect this will need to be tuned to meet the projects needs, but this will at least get us started with something to talk about.

As mentioned in https://github.com/go-swagger/go-swagger/issues/1576, the idea here is to leverage this open source service to up the quality gate for incoming PR's into the go-swagger project.

Also, this same configuration file can be used to run golangci's CLI locally, such that if any maintainers or contributors want to "tune up" the project, everyone will be playing by the same rules.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,19 @@
+linters-settings:
+  govet:
+    check-shadowing: true
+  golint:
+    min-confidence: 0
+  gocyclo:
+    min-complexity: 10
+  maligned:
+    suggest-new: true
+  dupl:
+    threshold: 100
+  goconst:
+    min-len: 2
+    min-occurrences: 2
+
+linters:
+  enable-all: true
+  disable:
+    - maligned


.golangci.yml
```diff
@@ -0,0 +1,19 @@
...
+  dupl:
+    threshold: 100
+  goconst:
+    min-len: 2
+    min-occurrences: 2
+
+linters:
+  enable-all: true
+  disable:
+    - maligned
```
gregmarr: Curious, if `maligned` is disabled, why is it configured above?

JonKohler: ah, I commented below, but this is actually what upstream golangCI runs on their code, so this is actually just the reference configuration. This certainly could be pruned, but I figured I'd start with upstream reference and go from there

casualjim:
can you update this branch, it's behind master

JonKohler:
done @casualjim

JonKohler:
checks are all passed now

JonKohler:
thanks for the review and merge everyone. I do suspect this is going to be a slightly iterative process to tun up the configuration, but its progress none the less!

fredbi:
yeah. I already changed cyclo warning to 20.

.golangci.yml
```diff
@@ -0,0 +1,19 @@
+linters-settings:
+  govet:
+    check-shadowing: true
+  golint:
+    min-confidence: 0
+  gocyclo:
+    min-complexity: 10
+  maligned:
+    suggest-new: true
```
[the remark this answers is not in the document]
jirfag: Yes, it is. Golangci lint uses forked maligned to print suggestions

PR #1579, opened 2018-06-15
intentionally bad change for golangci

JonKohler:
intentionally bad change for golangci

This should be flagged up and down by golangci, let's see what happens

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,19 @@
+linters-settings:
+  govet:
+    check-shadowing: true
+  golint:
+    min-confidence: 0
+  gocyclo:
+    min-complexity: 10
+  maligned:
+    suggest-new: true
+  dupl:
+    threshold: 100
+  goconst:
+    min-len: 2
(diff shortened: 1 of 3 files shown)


JonKohler:
welp, thats embarassing. let's try that again

JonKohler:
I need to reopen another junky PR to get it run again

fredbi:
Does the golangCI checks just for changes or run against the whole existing code?
I am running gometalinter from time to time, and we have a lot of warning to fix.

JonKohler:
It JUST checks the changes, as that would make PR's a nightmare for projects that

I did use gometalinter for a while, until I found golangci's CLI, and I really, really like it instead.

JonKohler:
Give this a try and see how you like it: https://github.com/golangci/golangci-lint

fredbi:
Yeah I looked at this link this morning to figure out changes re gometalinter.
What would be interesting is to run this in CI on sample generated code as well.

Issue #1572, opened 2018-06-09
Improve spec preprocessing : flatten vs expand


fredbi:
## Problem statement

Here are some points that I am considering faulty with `go-openapi/analysis/flatten.go`.

1. Too many anonymous OAIGen definitions generated. 
> OAIGen are generated whenever there is a name conflict when flattening the spec. 
> In some obvious cases, however, this conflict may be resolved without creating an additional definition.
> Flatten should issue a warning to the user to explain this "OAIGen" creation.

2.  Have an option to stop flattening as soon as references are good enough for the generator 
> At this stage it is not required for codegen to create all $ref's for complex things 
> (this is _minimal flattening_ ).
> All codegen will assume that this step has been completed, code checking for remote $ref
> (e.g. `!ref.HasFragmentsOnly`) may be removed

3.  Full flattening should consider maps of simple things as simple things (e.g. don't create a definition for map[]{my object}, the definition for {myobject} should be enough)

4. Have an option to remove unused definitions (e.g. `#/parameters` which are expanded), so as to produce a pristine spec.

I propose the following modification to `swagger` CLI options:
- `--with-flatten=minimal` (default)
- `--with-flatten=full` (correspond to current default)
- `--with-expand` | `--with-flatten=expand` (corresponds to current `--skip-flatten`)
- `--skip-flatten` is desupported

`--with-flatten` may support multiple options, e.g:
`--with-flatten:full,verbose,remove-unused`

Some more suggestions to come later on:
5. In full flatten mode, more options to control how allOf are handled
6. In full flatten mode propagating x-go-name (as prefix) to newly created definitions: creating new definitions introduces some naming conventions. If a x-go-name has been set in the original structure, new names are no consistent with their container...
> Example: if an array of complex items has `x-go-name: zzz`, the items definition created should have `x-go-name: zzzItems`
 
See also #1559 

Points (1), (2) and (3) and (4) are currently under testing on my dev.
I'll do my best to issue a "how-to flatten a spec" doc with this series of PR's.

Points (5) and (6) would require some preparatory work on models.

## Environment
swagger version: 0.14
go version: x.x.x  
OS:

gregmarr:
Sounds awesome.

orangle:
good ideas

fredbi:
Thanks. It is coming up this week in go-openapi/analysis, which introduces these new features.
For go-swagger, I would need PR #1566 and #1568  to be merged, so all testing goes well. 
[...]

PR #1585, opened 2018-06-19
Update vendor 20180630

fredbi:
This PR updates vendor dependencies.

It follows PR #1568 

It is essentially focused on testing go-openapi/analysis#30 (currently forced to a branch at my fork) 
and in particular, non-regression testing with the new flatten behavior (e.g. maps are simple types).

This testing uncovered some bugs, which are fixed in a distinct commit (now squashed).

Complementary integration work on go-swagger (CLI update, more testing with newly available options) will come as a separate PR.

Cheers,

Fred


fredbi:
Contributes to #1572

PR #1592, opened 2018-06-25
Tuple serializer

### LINKED ISSUE #1571: Tuple unmarshaller fails on nil fields
## Problem statement
In a tuple model, all tuple elements are described as required fields of a go struct (e.g. P0, P1, ...).

The unmarshaller currently fails to properly initialize this structure, as it expects a non nil value.

## Environment
swagger version: 0.14
go version: x.x.x  
OS:

fredbi:
* Fixes #1571
* Fix issue with arrays in additional items (index var conflict)
* Fix issue with anonymous objects as tuple elements

TODO
* [ ] Fix issue with special allOfs (e.g. allOf: [ ..., { x-nullable: true}])


fredbi:
Merge conflict is actually major here - I am taking care of it.

fredbi:
Testing this with anonymous elements uncovered a new issue with liftSpecialAllOf(), unrelated with tuples.
It so happens that the only test case for this was in a tuple element... This deserves a dedicated fix.

PR #1598, opened 2018-07-02
Refactor debug code

### LINKED ISSUE #634: Debug prints regorganization
## Problem statement

I see lot of places where the code looks like this:

```
if Debug {
    log.Println("absolute path", fa)
}
```

Event though it is trivial and debug only, it contributes into the code size overall reducing readability.
## Proposition.

 I would consider moving it out into just one function that supports some kind of log levels or I would just use an existing logging library.

Thoughts?

fredbi:
* Fixes #634


generator/debug.go
```diff
@@ -15,28 +15,47 @@
...
-func debugLog(format string, args ...interface{}) {
+// debugLogAsJSON unmarshals its last arg as pretty JSON
+func debugLogAsJSON(frmt string, args ...interface{}) {
 	if Debug {
-		_, file, pos, _ := runtime.Caller(2)
-		log.Printf("%s:%d: "+format, append([]interface{}{filepath.Base(file), pos}, args...)...)
+		_, file, pos, _ := runtime.Caller(1)
+		if len(args) > 0 {
+			bbb, _ := json.MarshalIndent(args[len(args)-1], "", " ")
+			generatorLogger.Printf("%s:%d: %s\n%s", filepath.Base(file), pos, fmt.Sprintf(frmt, args[0:len(args)-1]...), string(bbb))
```
golangci: line is 124 characters

generator/model.go
```diff
@@ -698,17 +678,11 @@ func (sg *schemaGenContext) MergeResult(other *schemaGenContext, liftsRequired b
...
-	}
+	debugLog("building properties %s (parent: %s)", sg.Name, sg.Container)
 
 	for k, v := range sg.Schema.Properties {
-		if Debug {
-			bbb, _ := json.MarshalIndent(sg.Schema, "", "  ")
-			log.Printf("building property %s[%q] (tup: %t) (BaseType: %t) %s\n", sg.Name, k, sg.IsTuple, sg.GenSchema.IsBaseType, bbb)
-			log.Printf("property %s[%q] (tup: %t) HasValidations: %t)", sg.Name, k, sg.IsTuple, sg.GenSchema.HasValidations)
-		}
+		debugLogAsJSON("building property %s[%q] (tup: %t) (BaseType: %t) %s", sg.Name, k, sg.IsTuple, sg.GenSchema.IsBaseType, sg.Schema)
```
golangci: line is 132 characters

generator/model.go
```diff
@@ -1799,11 +1752,7 @@ func (sg *schemaGenContext) buildMapOfNullable(sch *GenSchema) {
 }
 
 func (sg *schemaGenContext) makeGenSchema() error {
-	if Debug {
-		log.Printf("making gen schema (anon: %t, req: %t, tuple: %t) %s\n", !sg.Named, sg.Required, sg.IsTuple, sg.Name)
-		b, _ := json.MarshalIndent(sg.Schema, "", "  ")
-		log.Println(string(b))
-	}
+	debugLogAsJSON("making gen schema (anon: %t, req: %t, tuple: %t) %s\n", !sg.Named, sg.Required, sg.IsTuple, sg.Name, sg.Schema)
```
golangci: line is 128 characters

generator/shared.go
```diff
@@ -618,9 +616,7 @@ func (g *GenOpts) write(t *TemplateOpts, data interface{}) error {
 	}
 
 	if t.SkipExists && fileExists(dir, fname) {
-		if Debug {
-			log.Printf("skipping generation of %s because it already exists and skip_exist directive is set for %s", filepath.Join(dir, fname), t.Name)
-		}
+		debugLog("skipping generation of %s because it already exists and skip_exist directive is set for %s", filepath.Join(dir, fname), t.Name)
```
golangci: line is 139 characters

generator/debug_test.go
```diff
@@ -0,0 +1,84 @@
...
+		generatorLogger.SetOutput(os.Stdout)
+	}()
+	generatorLogger.SetOutput(tmpFile)
+
+	debugLog("A debug")
+	tmpFile.Close()
+
+	flushed, _ := os.Open(tmpName)
+	buf := make([]byte, 500)
+	flushed.Read(buf)
```
golangci: Error return value of `flushed.Read` is not checked

generator/debug_test.go
```diff
@@ -0,0 +1,84 @@
...
+		FieldOne string `json:"fieldOne"`
+	}{
+		FieldOne: "content",
+	}
+	debugLogAsJSON("A long debug:%t", true, sch)
+	tmpJSONFile.Close()
+
+	flushed, _ = os.Open(tmpJSONName)
+	buf2 := make([]byte, 500)
+	flushed.Read(buf2)
```
golangci: Error return value of `flushed.Read` is not checked

fredbi:
N.B: this has nothing to do with putting an external logger, like ZeroLog or some other coloured/json logger.

It is only about code readability (especially in model.go) and relevance of the gocov reports.

I also removed the unit test I add some months ago which was here to make sure no issue was introduced with all these `if Debug { ...}` and is no more relevant.

The "debugAsJson()" func is just here so simplify the debug code and remove the repeated calls ro `json.MarshallIndent()`

PR #1600, opened 2018-07-04
regen examples

### LINKED ISSUE #1496: Bug in oauth2 example
## Problem statement

In the oauth2 readme, an http redirect is written to an `gW` variable.

```go
func login(r *http.Request) string {
	// implements the login with a redirection and an access token
	var accessToken string
	http.Redirect(wG, r, config.AuthCodeURL(state), http.StatusFound)
	log.Println("Access token:", accessToken)
	return accessToken
}
```

I did not found any reference to it, so I looked it up in the example code, and found it [here](https://github.com/go-swagger/go-swagger/blob/master/examples/oauth2/restapi/configure_oauth_sample.go#L98-L109). In the middleware function the `http.ResponseWriter` is stored to a package variable.

This is of course a bug since different HTTP requests are in different goroutines and access this variable concurrently.

## Environment
swagger version: master

fredbi:
Fix #1496 (race in oauth2 sample code)


fredbi:
I am expecting golangCI to skip the examples... Let's see

fredbi:
Yep it skipped it. Pfew

PR #1612, opened 2018-07-09
Feature/x-order

### LINKED ISSUE #1609: Vendor extension x-example for Dredd not working
## Problem statement

I would like to use [Dredd](https://github.com/apiaryio/dredd) to test the generated Swagger against the API. The problem is I'm using path parameters which required a default or an example by Dredd, but since it's a required field, I have to use an example value. I don't think Swagger can generate the vendor extension I need because when I add `// x-example: USERID` to the struct, it will remove the "x-" in the spec which is not the behavior I need.

## Swagger specification

This is the spec it generates with `example` even thought I specify `x-example` in the Go code annotation.

```json
...
        "summary": "Delete a user.",
        "operationId": "UserDestroy",
        "parameters": [
          {
            "type": "string",
            "example": "USERID",
            "x-go-name": "UserID",
            "name": "user_id",
            "in": "path",
            "required": true
          }
        ],
...
```

This is the spec I need with `x-example`:

```json
...
        "summary": "Delete a user.",
        "operationId": "UserDestroy",
        "parameters": [
          {
            "type": "string",
            "x-example": "USERID",
            "x-go-name": "UserID",
            "name": "user_id",
            "in": "path",
            "required": true
          }
        ],
...
```

## Steps to reproduce

Create go code like this:
```go
	// swagger:parameters UserShow
	type request struct {
		// in: path
		// x-example: USERID
		UserID string `json:"user_id" validate:"required"`
	}
```

Then run Swagger generate to generate the spec.

## Environment
swagger version: dev (from master)
go version: 1.10.1 darwin/amd64
OS:   MacOS 10.13.5

liber-takano:
`x-order`:  To sort struct properties order.

It useful with msgpack-array.

```
  sessionData:
    type: object
    properties:
      sessionId:
        x-order: 0
        type: string
        description: session unique
      deviceId:
        x-order: 1
        x-go-custom-tag: hello1
        type: string
        description: deveide unique
      uMain:
        $ref: '#/definitions/u_main'
        x-order: 2
        x-go-custom-tag: hello2
        description: Sample
```

Output is sorted by x-order, no alphabet order.

```
type SessionData struct {

	SessionID string `json:"sessionId,omitempty"`

	DeviceID string `json:"deviceId,omitempty" hello1`

	UMain *UMain `json:"uMain,omitempty" hello2`
}
```

### DIFF
--- a/generator/model.go
+++ b/generator/model.go
@@ -837,9 +837,12 @@ func (sg *schemaGenContext) buildProperties() error {
 			emprop.GenSchema.ValueExpression += "()"
 		}
 
+		emprop.GenSchema.Extensions = emprop.Schema.Extensions
+
 		// set custom serializer tag
 		if customTag, found := emprop.Schema.Extensions[xGoCustomTag]; found {
 			emprop.GenSchema.CustomTag = customTag.(string)
+			delete(emprop.GenSchema.Extensions, xGoCustomTag)
 		}
 		sg.GenSchema.Properties = append(sg.GenSchema.Properties, emprop.GenSchema)
 	}
--- a/fixtures/codegen/x-order.yml
+++ b/fixtures/codegen/x-order.yml
@@ -1,0 +1,109 @@
+swagger: '2.0'
+info:
+  description: Test for x-order
+  version: 0.0.1
+  title: Test for x-order
+  contact:
+    email: takano@liberent.co.jp
+schemes:
+  - http
+paths:
+  /user/startSession:
+    post:
+      summary: Test for x-order
+      description: Test for x-order
+      operationId: StartSession
+      consumes:
+        - application/json
+        - application/x-msgpack-array
+      produces:
+        - application/json
+        - application/x-msgpack-array
+      parameters:
+        - in: body
+          name: clientProfile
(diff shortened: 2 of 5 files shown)


generator/model.go
```diff
@@ -837,9 +837,12 @@ func (sg *schemaGenContext) buildProperties() error {
 			emprop.GenSchema.ValueExpression += "()"
 		}
 
+		emprop.GenSchema.Extensions = emprop.Schema.Extensions
+
 		// set custom serializer tag
 		if customTag, found := emprop.Schema.Extensions[xGoCustomTag]; found {
 			emprop.GenSchema.CustomTag = customTag.(string)
+			delete(emprop.GenSchema.Extensions, xGoCustomTag)
```
fredbi: @liber-takano I merged this without paying much attention at first.
Now that I am looking at it more at ease, I am really wondering why you want to delete this.
How about other template users who use Extensions?
Why is the redundant presence of xGoCustomTag here an issue for you?

liber-takano: For xGoCustomTag, it is okay to make your judgment top priority.

PR #1615, opened 2018-07-10
Fixes #1614

### LINKED ISSUE #1614: Stack overflow error in swagger validate
When running `swagger validate ./public/swagger.v1.json` probably after #1608 was merged we are having error:
```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

For more details see: https://drone.gitea.io/go-gitea/gitea/1918/8

fredbi:


fredbi:
I assume the codecov difference is due to a change in codecov's interpretation of what is a line passed and what is not, because not much as changed really with this...

PR #1619, opened 2018-07-14
make planning of operations parallel

koron:
`generate xxx` (`planning operations` phase) is too slow.

`codeGenOpBuilder.MakeOperation()` takes about 500msec for **an operation** in my env.
And I have over 100 operations in my swagger.yml, then...

    500msec/op * 100op = 50s

It takes about 1min.

I make it run in parallel to make the phase faster.

### DIFF
--- a/generator/support.go
+++ b/generator/support.go
@@ -16,6 +16,7 @@ package generator
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


koron:
Should I include `golang.org/x/sync/semaphore` and its dependencies into `vendor/`?
Or just copy minimum files with some modifications?

fredbi:
Do not cherry pick dependencies: add you dependency to Gopkg.toml and update vendor.

fredbi:
There is something that I did not understand here. Maybe you could explain.
When running a MakeOperation, there is a lot of digging inside the spec.
And there many maps accessed via pointers.

How do you protect these maps from being accessed concurrently (even just for reading)?

koron:
I created better PR https://github.com/go-swagger/go-swagger/pull/1620 for this problem.
So I close this.

PR #1626, opened 2018-07-19
Add contributed template option and stratoscale's templates

### LINKED ISSUE #1625: Contributed templates
Add new feature to support contributed templates.
As a result of https://github.com/Stratoscale/swagger and after discussion with @fredbi we would like to add an option for defining "contributed" templates which will be delivered with the swagger binary.

The suggested solution is to add a `--template` flag to the generate command, and specify the contributed template name. This will result in overriding the default templates from templates which are included in the binary.

This solution also need to give a way for custom templates to control the generated options, such as excluding main or rewriting the configureapi file.

michaelf-stratoscale:
* Add `/contrib/stratoscale` to `generator/templates` which include stratoscale's templates
* Add `--template` flag which chooses a contributed template by name
* Add `--regenerate-configureapi` flag to `generate server` command, which forces rewriting of the configureapi

Closes #1625


generator/template_repo.go
```diff
@@ -285,6 +284,29 @@ func (t *Repository) LoadDir(templatePath string) error {
 	return nil
 }
 
+// LoadContrib loads template from contrib directory
+func (t *Repository) LoadContrib(name string) error {
+	log.Printf("loading contrib %s", name)
+	const pathPrefix = "generator/templates/contrib/"
+	basePath := filepath.Join(pathPrefix, name)
+	err := filepath.Walk(basePath, func(path string, info os.FileInfo, err error) error {
```
golangci: argument err is overwritten before first use

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
-	SkipSupport       bool     `long:"skip-support" description:"no supporting files will be generated when this flag is specified"`
-	ExcludeMain       bool     `long:"exclude-main" description:"exclude main function, so just generate the library"`
-	ExcludeSpec       bool     `long:"exclude-spec" description:"don't embed the swagger specification"`
-	WithContext       bool     `long:"with-context" description:"handlers get a context as first arg (deprecated)"`
-	DumpData          bool     `long:"dump-data" description:"when present dumps the json for the template generator instead of generating files"`
-	FlagStrategy      string   `long:"flag-strategy" description:"the strategy to provide flags for the server" default:"go-flags" choice:"go-flags" choice:"pflag"`
-	CompatibilityMode string   `long:"compatibility-mode" description:"the compatibility mode for the tls server" default:"modern" choice:"modern" choice:"intermediate"`
-	SkipValidation    bool     `long:"skip-validation" description:"skips validation of spec prior to generation"`
+	Name                   string   `long:"name" short:"A" description:"the name of the application, defaults to a mangled value of info.title"`
```
golangci: line is 141 characters

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	Operations             []string `long:"operation" short:"O" description:"specify an operation to include, repeat for multiple"`
```
golangci: line is 128 characters

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	Tags                   []string `long:"tags" description:"the tags to include, if not specified defaults to all"`
+	Principal              string   `long:"principal" short:"P" description:"the model to use for the security principal"`
+	DefaultScheme          string   `long:"default-scheme" description:"the default scheme for this API" default:"http"`
+	Models                 []string `long:"model" short:"M" description:"specify a model to include, repeat for multiple"`
+	SkipModels             bool     `long:"skip-models" description:"no models will be generated when this flag is specified"`
```
golangci: line is 123 characters

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	SkipOperations         bool     `long:"skip-operations" description:"no operations will be generated when this flag is specified"`
```
golangci: line is 131 characters

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	SkipSupport            bool     `long:"skip-support" description:"no supporting files will be generated when this flag is specified"`
```
golangci: line is 134 characters

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	ExcludeMain            bool     `long:"exclude-main" description:"exclude main function, so just generate the library"`
+	ExcludeSpec            bool     `long:"exclude-spec" description:"don't embed the swagger specification"`
+	WithContext            bool     `long:"with-context" description:"handlers get a context as first arg (deprecated)"`
+	DumpData               bool     `long:"dump-data" description:"when present dumps the json for the template generator instead of generating files"`
```
golangci: line is 148 characters

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	FlagStrategy           string   `long:"flag-strategy" description:"the strategy to provide flags for the server" default:"go-flags" choice:"go-flags" choice:"pflag"`
```
golangci: line is 166 characters

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	CompatibilityMode      string   `long:"compatibility-mode" description:"the compatibility mode for the tls server" default:"modern" choice:"modern" choice:"intermediate"`
```
golangci: line is 171 characters

generator/client.go
```diff
@@ -35,6 +35,10 @@ func GenerateClient(name string, modelNames, operationIDs []string, opts *GenOpt
 		return errors.New("gen opts are required")
 	}
 
+	if opts.Template != "" {
+		templates.LoadContrib(opts.Template)
```
golangci: Error return value of `templates.LoadContrib` is not checked

generator/support.go
```diff
@@ -59,6 +59,9 @@ func newAppGenerator(name string, modelNames, operationIDs []string, opts *GenOp
 	}
 
 	templates.LoadDefaults()
+	if opts.Template != "" {
+		templates.LoadContrib(opts.Template)
```
golangci: Error return value of `templates.LoadContrib` is not checked

cmd/swagger/commands/generate/shared.go
```diff
@@ -75,6 +75,7 @@ type shared struct {
 	ServerPackage         string         `long:"server-package" short:"s" description:"the package to save the server specific code" default:"restapi"`
 	ClientPackage         string         `long:"client-package" short:"c" description:"the package to save the client specific code" default:"client"`
 	Target                flags.Filename `long:"target" short:"t" default:"./" description:"the base directory for generating the files"`
+	Template              string         `long:"template" description:"Load contributed templates" choice:"stratoscale"`
```
fredbi: I would prefer 'TemplateSet' here, I think it is more explicit about the intent.

cmd/swagger/commands/generate/server.go
```diff
@@ -24,55 +24,60 @@ import (
...
+	SkipValidation         bool     `long:"skip-validation" description:"skips validation of spec prior to generation"`
+	RegenerateConfigureAPI bool     `long:"regenerate-configureapi" description:"Force regeneration of configureapi.go"`
```
fredbi: Do you really need this? This option is actually already available in the config of the template itself.
You may override it directly without adding this bool.

cmd/swagger/commands/generate/contrib.go
```diff
@@ -0,0 +1,14 @@
+package generate
+
+import "github.com/go-swagger/go-swagger/generator"
+
+// contribOptionsOverride gives contributed templates the ability to override the options if they need
+func contribOptionsOverride(opts *generator.GenOpts) {
+	switch opts.Template {
+	case "stratoscale":
+		// Stratoscale template needs to regenerate the configureapi on every run.
+		opts.RegenerateConfigureAPI = true
```
fredbi: This may be overriden directly in the section running templates. There is a `SkipExists` flag on each template.

cmd/swagger/commands/generate/support.go
```diff
@@ -42,8 +42,11 @@ func (s *Support) getOpts() (*generator.GenOpts, error) {
 		Principal:     s.Principal,
 		DumpData:      s.DumpData,
 		DefaultScheme: s.DefaultScheme,
+		Template:      s.Template,
 		TemplateDir:   string(s.TemplateDir),
-	}, nil
+	}
+	contribOptionsOverride(opts)
```
fredbi: may be guarded by `!= "" or any default value here.

generator/template_repo.go
```diff
@@ -285,6 +284,29 @@ func (t *Repository) LoadDir(templatePath string) error {
 	return nil
 }
 
+// LoadContrib loads template from contrib directory
```
fredbi: Do you really need this?
Afaik, dependencies are managed automatically by the template repo.
This works when there are external files, I did not check with preloaded assets.

posener: @fredbi this function replaces the "default" template by the one defined in the "contrib/<stratoscale>" directory.
The "contrib/<stratoscale>" was added to the bindata.go file, but they only replace the default files when the --template flag is set.
You may suggest another way to switch the default file and we will implement it.

cmd/swagger/commands/generate/server.go
michaelf-stratoscale: @fredbi what do you mean here?

generator/template_repo.go
fredbi: What I mean is why don't you just mimick what we do manually with the yaml config file?

When using the config file, I change the path of my templates and everything which needs override is well resolved without a single change in the repo_template code.

If there is a way to do that from config rather than altering this code, all the better, don't you think?

cmd/swagger/commands/generate/server.go
fredbi: I am adding here a sample config file of mine which overrides things.
In my opinion, you should alter this struct defaults (e.g. before loading the user's own config file)  rather than going with more variables in `GenOpts`.

Note the SkipExists flag and the Source entry referring to the template with implied path (prefixed with asset if we assume this is preloaded).

```yaml
layout:
  application:
    - name: configure
      # Generates configure_api.go. 
      # This is a modified bersion of the standard version, with 
      # automatic registration of API operations handlers and 
      # custom OAuth2 authorization method.
      source: asset:serverConfigureapi
      target: "{{ joinFilePath .Target .ServerPackage }}"
      file_name: "configure_{{ .Name }}.go"
      # Set to false now that our template links to implem
      skip_exists: false
      skip_format: false
      # Generates main for server [disabled]
    - name: main
      source: asset:serverMain
      target: "{{ joinFilePath .Target \"cmd\" (dasherize (pascalize .Name)) }}-server"
      file_name: "main.go"
      # Generates embedded spec
    - name: embedded_spec
      source: asset:swaggerJsonEmbed
      target: "{{ joinFilePath .Target .ServerPackage }}"
      file_name: "embedded_spec.go"
      # Generates server
    - name: server
      source: asset:serverServer
      target: "{{ joinFilePath .Target .ServerPackage }}"
      file_name: "server.go"
      # Generates xxx_api.go (standard template).
    - name: builder
      source: asset:serverBuilder
      target: "{{ joinFilePath .Target .ServerPackage .Package }}"
      file_name: "{{ snakize (pascalize .Name) }}_api.go"
      # Generates doc server. This is a stripped down version of the standard  
      # server, with an interface to allow for composition
    - name: doc
      source: asset:serverDoc
      target: "{{ joinFilePath .Target .ServerPackage }}"
      file_name: "doc.go"
      # Generates common methods for implementation package [specific]
    - name: implementation_common
      source: server/implementation_common.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "implementation_{{ (snakize (pascalize .Name)) }}_common.go"
      skip_exists: false
      skip_format: false
      # Unit test for common methods for implementation package [specific]
    - name: implementation_common_test
      source: server/implementation_common_test.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "implementation_{{ (snakize (pascalize .Name)) }}_common_test.go"
      skip_exists: false
      skip_format: false
      # Generates responders for implementation package [specific]
    - name: implementation_responder
      source: server/implementation_responder.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "implementation_{{ (snakize (pascalize .Name)) }}_responder.go"
      skip_exists: false
      skip_format: false
    - name: implementation_responder_test
      source: server/implementation_responder_test.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "implementation_{{ (snakize (pascalize .Name)) }}_responder_test.go"
      skip_exists: false
      skip_format: false
    - name: implementation_parameter_test
      source: server/implementation_parameter_test.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "implementation_{{ (snakize (pascalize .Name)) }}_parameter_test.go"
      skip_exists: false
      skip_format: false
    - name: implementation_gen_fixtures
      source: server/gen_fixtures.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "gen_fixtures.sh"
      skip_exists: false
      skip_format: true
  models:
    - name: definition
      source: asset:model
      target: "{{ joinFilePath .Target .ModelPackage }}"
      file_name: "{{ (snakize (pascalize .Name)) }}.go"
  operations:
    - name: parameters
      source: asset:serverParameter
      target: "{{ if gt (len .Tags) 0 }}{{ joinFilePath .Target .ServerPackage .APIPackage .Package  }}{{ else }}{{ joinFilePath .Target .ServerPackage .Package  }}{{ end }}"
      file_name: "{{ (snakize (pascalize .Name)) }}_parameters.go"
    - name: responses
      source: asset:serverResponses
      target: "{{ if gt (len .Tags) 0 }}{{ joinFilePath .Target .ServerPackage .APIPackage .Package  }}{{ else }}{{ joinFilePath .Target .ServerPackage .Package  }}{{ end }}"
      file_name: "{{ (snakize (pascalize .Name)) }}_responses.go"
    - name: handler
      source: asset:serverOperation
      target: "{{ if gt (len .Tags) 0 }}{{ joinFilePath .Target .ServerPackage .APIPackage .Package  }}{{ else }}{{ joinFilePath .Target .ServerPackage .Package  }}{{ end }}"
      file_name: "{{ (snakize (pascalize .Name)) }}.go"
  operation_groups:
    # Generates API skeleton methods for implementation package [specific]
    - name: implementation_groups
      source: server/implementation_groups.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "implementation_{{ snakize (pascalize .Name) }}.go"
      skip_exists: false
      skip_format: false
    - name: implementation_groups_test
      source: server/implementation_groups_test.gotmpl
      target: "{{ joinFilePath .Target \"implementation\" }}"
      file_name: "implementation_{{ (snakize (pascalize .Name)) }}_test.go"
      skip_exists: false
      skip_format: false
```

cmd/swagger/commands/generate/shared.go
fredbi: We agree this is a choice string, not a flag, as discussed in the associated issue.

generator/template_repo.go
fredbi: Consider applying overrides here: https://github.com/go-swagger/go-swagger/blob/8914a0f9bc4c11970552f133f8c9af106c1ac511/generator/shared.go#L222

cmd/swagger/commands/generate/support.go
michaelf-stratoscale: @fredbi what do you mean here?

fredbi: Shouldn't we avoid calling when not used?
```go
if Template != "" {
  contribOptionsOverride(opts)
}
````

Comment (fredbi) [reply cmd/swagger/commands/generate/support.go]: Oh yes I did not figure it out. This is not very important

File: cmd/swagger/commands/generate/contrib.go (hunk 14 above)
Comment (fredbi) [reply cmd/swagger/commands/generate/contrib.go]: You have to consider that the user might chose to override the stratoscale template with a derivation of her own making, override again this skip decision for some reason.

I would say that we should just override the defaults in the templates section before we load the user's own config file.

File: cmd/swagger/commands/generate/server.go (hunk 13 above)
Comment (michaelf-stratoscale) [reply cmd/swagger/commands/generate/server.go]: It looks like a good solution, the only problem is that in the way it works right now, when i load a configuration file it does not append or override only the things that i set, it completely erase the defaults and uses only what is written in the configuration file.
Do you know of  way to set only the things that i write and the rest will be as the default?

File: cmd/swagger/commands/generate/contrib.go [hunk 17]
```diff
@@ -0,0 +1,16 @@
+package generate
+
+import (
+	"github.com/go-swagger/go-swagger/generator"
+)
+
+// contribOptionsOverride gives contributed templates the ability to override the options if they need
+func contribOptionsOverride(opts *generator.GenOpts) {
+	switch opts.Template {
+		generator/shared.go	case "stratoscale":
```
golangci: expected '}', found 'IDENT' generator (and 1 more errors)

cmd/swagger/commands/generate/server.go
michaelf-stratoscale: @fredbi as we spoke on slack, `contribOptionsOverride` is used to override the default flags. this function is called only from one place and allow the user to override with his configuration file

cmd/swagger/commands/generate/shared.go
michaelf-stratoscale: is this still relevant?

cmd/swagger/commands/generate/server.go
posener: @fredbi why are you enforcing line lengths? It might make the code less readable.
According to [effective go](https://golang.org/doc/effective_go.html):
> Line length
Go has no line length limit. Don't worry about overflowing a punched card. If a line feels too long, wrap it and indent with an extra tab.

fredbi: it is a standard rule in golangci to improve readability.

If you find a nice way to tune this in .golangci.yml, so it does not take literal strings in its count, but otherwise works on regular statements, I am interested.

N.B: you may figure out locally using `golangci-lint run` (to be installed from github), which now replaces `gometalinter` on our repos...

posener: looks like the flag is `--disable lll`

generator/template_repo.go
```diff
@@ -285,6 +284,32 @@ func (t *Repository) LoadDir(templatePath string) error {
...
+			err := t.addFile(target, string(MustAsset(aname)), true)
+			if err != nil {
+				return err
+			}
+			log.Printf("added contributed template %s from %s", target, aname)
+			filesAdded++
+		}
+	}
+	if filesAdded == 0 {
+		return fmt.Errorf("No files added from template: %s", name)
```
golangci: error strings should not be capitalized or end with punctuation or a newline

generator/client_test.go
```diff
@@ -33,6 +33,51 @@ func TestClient_InvalidSpec(t *testing.T) {
...
+		t.Fatalf("Failed to create a test target directory: %v", err)
+	}
+	log.SetOutput(ioutil.Discard)
+	defer func() {
+		_ = os.RemoveAll(targetdir)
+		log.SetOutput(os.Stdout)
+	}()
+	opts := testGenOpts()
+	opts.Target = targetdir
+	opts.Spec = "../fixtures/petstores/petstore.json"
```
golangci: string `../fixtures/petstores/petstore.json` has 2 occurrences, make it a constant

generator/client_test.go
```diff
@@ -43,9 +33,55 @@ func TestClient_BaseImportDisabled(t *testing.T) {
...
+		name      string
+		template  string
+		wantError bool
+		prepare   func(opts *GenOpts)
+	}{
+		{
+			name:      "InvalidSpec",
+			wantError: true,
+			prepare: func(opts *GenOpts) {
+				opts.Spec = "../fixtures/bugs/825/swagger.yml"
```
golangci: string `../fixtures/bugs/825/swagger.yml` has 2 occurrences, make it a constant

.golangci.yml
```diff
@@ -17,3 +17,4 @@ linters:
   enable-all: true
   disable:
     - maligned
+    - lll
```
fredbi: Would prefer that we relax this constraint rather than disabling it altogether.
i.e; 
```yaml
 lll:
# max line length, lines longer will be reported. Default is 120. '\t' is counted as 1 character.
line-length: 200
```

should pass and still we guard against unwanted code

cmd/swagger/commands/generate/contrib.go
fredbi: OK, since this option is now available for every template to override the config setting, contrib template or not.

cmd/swagger/commands/generate/server.go
fredbi: like mentioned above, I would prefer relaxing the check than disabling it

cmd/swagger/commands/generate/shared.go
fredbi: ok let it be like this

generator/templates/contrib/stratoscale/README.md
```diff
@@ -0,0 +1,311 @@
...
+   
+```go
+for _, auth := range route.Authenticators {
+    for scopeName, scopeValues := range auth.Scopes {
+        for _, scopeValue := range scopeValues {
+            ...
+        }
+    }
+}
+```
```
fredbi: This markdown would be better located in the `./examples` directory.
You are welcome to add some `example/contributed-templates/stratoscale` directory.
Would be nice to add in there code for a full generated example.
The maintenance of examples (regeneration)) is normally carried out manually from time to time using the script in `hack/regen-sample.sh`. Please add your sample doc and code and update this script so we can maintain the example up-to-date with every maintenane.

.golangci.yml
fredbi: ok. We discussed it and I rest my case. Let's disable it then

cmd/swagger/commands/generate/server.go
fredbi: ok. We discussed it and I rest my case. Let's disable it then

PR #1627, opened 2018-07-19
Many shutdown fixes

Phlamethrower:
It looks like there are a few bugs with the shutdown code. Mainly problems with the `(*Server) Shutdown()` function (which I like to use in unit tests - starting and stopping temporary server instances to check API behaviour), but also some issues with shutdowns which are triggered by signals.

The result of these changes should be that:

* `(*Server) Shutdown()` will now cause the server to gracefully shutdown, and `(*Server) Serve()` to return without generating an error.
  * Currently it'll exit via `os.Exit(1)` or generate a panic, with a slim chance of first waiting for one of the listeners to perform a graceful shutdown.
* SIGINT and SIGTERM will now cause the server to gracefully shutdown, and `(*Server) Serve()` to return without generating an error.
  * Currently it'll immediately exit via `os.Exit(1)` or generate a panic, with no graceful shutdown.
* If you have multiple listeners, `api.ServerShutdown()` will only be called once, and only if all the `http.Server`s shut down without error.
  * Currently `api.ServerShutdown()` will be called once per listener (while other listeners may still be processing traffic), which I'm guessing wasn't intended

What's the best way of adding some unit tests to the project to check that shutdown is working correctly? Although I can find plenty of spec yaml files in the repository that I think are used for unit tests, I can't see anything obvious which tests for the behaviour of the generated server code. Without unit tests, it's just a matter of time until someone comes along and breaks things again (I have a feeling that this code has been broken for a long time).


fredbi:
Please squash your commits before merging.

casualjim:
I will push a 0.15.1 with this fix over the weekend

Phlamethrower:
Thanks!

mmmguitar:
Hey, just wondering if that 0.15.1 release mentioned above is still planned for?  We're on 0.15.0 but are a little reticent upgrading to the 0.16.0, seems to be a few issues and certainty one is breaking us immediately (https://github.com/go-swagger/go-swagger/issues/1683 but its a simple workaround).  Whenever we roll containers over / upgrade etc they panic on shutdown / I can see the fix in here.  

Thanks.

fredbi:
There won't be any 0.15.1. Just 0.16.

mmmguitar:
ok, thanks.

Issue #1602, opened 2018-07-05
Fail to generate model: unhandled type to resolve JSON pointer


zhuangqh:
## Problem statement

We are using go-swagger v0.12 to generate types, which works fine. We hope to upgrade go-swagger to the latest version without incompatibility and benefit from other feature. I think #1572 help a lot. However, I fail to generate model with latest version of go-swagger.

run
```bash
swagger generate model -f swagger.yml
```

I got error
```
2018/07/05 14:09:30 validating spec /mnt/hgfs/vmshare/go/src/github.com/alibaba/pouch/hack/../apis/swagger.yml
2018/07/05 14:09:33 preprocessing spec with option:  minimal flattening
unhandled type to resolve JSON pointer: spec.Response
```

debug mode
```
...
analysis:2018/07/05 14:29:08 flatten.go:1401: normalizeRef
analysis:2018/07/05 14:29:08 flatten.go:1222: name pointers
analysis:2018/07/05 14:29:08 flatten.go:1336: deepestRef for #/responses/401ErrorResponse
unhandled type to resolve JSON pointer: spec.Response
```

## Swagger specification
https://github.com/alibaba/pouch/blob/master/apis/swagger.yml

## Environment
swagger version: build from master a6de9fc6ed034d2bc27c0818571f888ec59084eb
go version: 1.9.7  
OS:   ubuntu 18.04

fredbi:
okay I see where the bug is located.
I omitted in test cases the case where the expanded shared response still hold a $ref.
This should be the same problem for #/parameters.
I'll fix this today, it is not a hard one

zhuangqh:
I’m truly grateful for your help. @fredbi

fredbi:
I acknowledge the bug and shall make the code more resilient, but I have to say it comes from an  invalid specification:
line 791, your schema incorrectly points to the response descriptor. That is actually what makes the code to fail: I did not expect a schema to point to a response object.

fredbi:
With this fix and using --withFlatten=verbose, you should get a warning about what is likely to be a wrong $ref. However, the flatten shall recover from this by assuming you mean the underlying response schema.

fredbi:
I was pondering about this. I was not sure I have a valid case for this

zhuangqh:
@casualjim I agree with you. It would be better if go-swagger could tell me what's wrong and how to fix it.

fredbi:
done with error and hopefully more informative error message

[...]

PR #1631, opened 2018-07-22
Improvements to minimal flattening

### LINKED ISSUE #1614: Stack overflow error in swagger validate

### LINKED ISSUE #1621: Client generation fails with 0.15.0, works with 0.14.0
## Problem statement

Just tested the `quay.io/goswagger/swagger:0.15.0` docker image as a replacement for 0.14.0 for generating a client.

The error output was:

```
name inlined schema: object has no field /v4/users/%7Bemail%7D/
```

With 0.14.0. the generation works fine based off the identical spec.

## Swagger specification

The spec is in this folder:

https://github.com/giantswarm/api-spec/tree/14d57a701b20135b174e97a7e755e0b49f28c097

The main file is `spec.yaml`, there are three more referenced YAML files in the same folder. The place where the generation broke is likely [line 293 or spec.yaml](https://github.com/giantswarm/api-spec/blob/14d57a701b20135b174e97a7e755e0b49f28c097/spec.yaml#L293):

## Steps to reproduce

1. `git clone https://github.com/giantswarm/gsclientgen.git`
2. Edit `Makefile` to replace `quay.io/goswagger/swagger:0.15.0` with `quay.io/goswagger/swagger:0.15.0`
3. Run `make generate`

This will use the current `master` of the spec linked above. To make sure the exact same version is used, edit `Makefile` once more and change the `BRANCH` variable in the top from `master` to `14d57a701b20135b174e97a7e755e0b49f28c097`.

## Environment

swagger version: The spec uses Swagger 2.0
OS: Linux (official docker image)

fredbi:
* fixes #1621
* fixes #1614
* improves #1602 and #1614 (flatten consistent with validate regarding pointers to responses/parameters as schemas)

Carry out vendor update, including:
* go-openapi/swag (comparing floats)
* mailru/easyjson


fredbi:
Stumbled on yet another bug/regression. This one will have to wait another fix to pass.

fredbi:
all fixes applied in go-openapi dependant packages

fredbi:
codecov/patch non-significant here

fredbi:
Where is appveyor check gone? Why does this new licensing check never complete?

fredbi:
Looks like appveyor is run but the CI report is dropped

PR #1633, opened 2018-07-23
[WIP] Ensure non-nil passed to additionalProperties in json.Unmarshall(...)

drewsonne:


generator/model_test.go
```diff
@@ -2436,6 +2436,32 @@ func TestGenModel_Issue910(t *testing.T) {
 	}
 }
 
+// This tests that additionalProperties with validation is generated properly.
+func TestGenModel_Issue1632(t *testing.T) {
```
golangci: 2440-2463 lines are duplicate of `generator/model_test.go:2192-2215`

PR #1636, opened 2018-07-24
Support go.mod from Go 1.11

shaxbee:
Crude module resolution using regex on go.mod file, tested with go 1.11 built from master by generating server, compiling and running it.


casualjim:
can't it call vgo and capture the output?

shaxbee:
Vgo has been merged into go command in 1.11 but the module resolver is in
internal package.
Calling out external program feels a bit heavy handed.

PR #1638, opened 2018-07-28
Make format resolving strictly to require a valid pair of type and format.

co3k:
The before implementation checks format only but it causes wrong type decisions.

e.g. {"type": "string", "format": "uint64"} had been converted to uint64.
     This behaivior is correct in case of "type" is "integer" or "number", but the "string" case is not.

This commit changes this decision algorithm. "format" is not treated as same of "type" anymore.
"format" must be available when it is specified with the valid (matched) "type". If not, the "format" will be ignored
and it just decides the type of Go code by using "type" value.

### DIFF
--- a/generator/formats.go
+++ b/generator/formats.go
@@ -92,59 +92,82 @@ var stringFormatters = map[string]string{
 	"uint64":  "swag.FormatUint64",
 }
 
-// typeMapping contains a mapping of format (or type name) to go type
+// typeMapping contains a mapping of type name to go type
 var typeMapping = map[string]string{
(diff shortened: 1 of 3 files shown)


co3k:
Oops, the unit test has been passed but "Test codegen builds" is failed in CircleCI. How should I do test it?

casualjim:
You build your binary with `go install` and run `./hack/codegen-nonreg.sh`

fredbi:
@co3k you can run locally this build script using ./hack/codegen-nonreg.sh

PR #1640, opened 2018-07-29
Mangle file names with a specific meaning for go build

### LINKED ISSUE #1596: Handle platform-specific file names
## Problem statement

I just got bitten by this. If you have a field name that ends in something like "_windows", then the file produced will end with "_windows.go", and will only be included in the build if you are on that OS. Likewise for other fields that happen to match a possible GOARCH or GOOS value (see https://golang.org/pkg/go/build/)

I could imagine several ways to handle this:
- Stop using underscores in file names, since it's not recommended anyway
- Warn or err when a path component would overlap with GOARCH/GOOS
- Rather than warn or err, use some replacement

fredbi:
This includes all GOOS, GOARCH and "_test"

* fixes #1596

### DIFF
--- a/generator/shared.go
+++ b/generator/shared.go
@@ -48,6 +48,7 @@ type LanguageOpts struct {
 	reservedWordsSet map[string]struct{}
 	initialized      bool
 	formatFunc       func(string, []byte) ([]byte, error)
+	fileNameFunc     func(string) string
 }
 
 // Init the language option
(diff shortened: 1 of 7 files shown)


generator/shared.go
```diff
@@ -113,6 +166,21 @@ func GoLangOpts() *LanguageOpts {
...
 	}
+	opts.fileNameFunc = func(name string) string {
+		// whenever a generated file name ends with a suffix
+		// that is meaningful to go build, adds a "swagger"
+		// suffix
+		parts := strings.Split(swag.ToFileName(name), "_")
+		_, isArch := goarchValues[parts[len(parts)-1]]
+		_, isOs := goosValues[parts[len(parts)-1]]
+		_, isOtherReserved := goOtherReservedSuffixes[parts[len(parts)-1]]
+		if isArch || isOs || isOtherReserved {
```
gregmarr: This checks all three maps even when the first one matches.  It would be better to short circuit as soon as one matches.  For that matter, there is no reason that you need 3 maps here.  A single map with all reserved suffixes should be fine.  Then you only need to access the last part once and can do this:

    if goReservedSuffixes[len(parts)-1] {
    ...

fredbi: ok done. Thanks

fredbi:
ok done

PR #1641, opened 2018-07-29
Unmarshalling array of polymorphic type in body parameter

### LINKED ISSUE #1637: Server cannot unmarshal array of polymorphic values
## Problem statement

Not sure if this is related to #842, but when an endpoint expects an array of polymorphic values, the server fails with the following error message:

```
{"code":400,"message":"parsing contents body from \"\" failed, because json: cannot unmarshal object into Go value of type models.Value"}
```

## Swagger specification

``` yaml
swagger: '2.0'
info:
  version: '0.0.1'
  title: Test
schemes:
  - http
basePath: /v1
produces:
  - application/json
consumes:
  - application/json
paths:
  /test:
    put:
      operationId: test
      parameters:
      - name: payload
        in: body
        required: true
        schema:
          type: array
          items:
            $ref: '#/definitions/Value'
      tags:
        - Test
      responses:
        204:
definitions:
  Value:
    type: object
    discriminator: ValueType
    required:
    - ValueType
    properties:
      ValueType:
        type: string
  StringValue:
```

## Steps to reproduce

- Implement a dummy handler  
- Build and start the server  
- execute the following curl:

```
curl 'http://localhost:PORT/v1/test -X PUT -H 'Content-type: application/json' -d '[{"ValueType":"StringValue","Foobar":"hello"}]'
```

## Environment
swagger version: 0.12.0  
go version: go1.9.1 darwin/amd64  
OS:   MacOS but should be irrelevant

## Workaround

When `payload` is to be an object that contains the array instead of just the array, it works.

``` yaml
- name: payload
    in: body
    required: true
    schema:
      type: object
      properties:
        values:
          type: array
          items:
            $ref: '#/definitions/Value'
```

Since the issue can be worked around it is not a blocking issue for me, I am submitting this issue more out of curiosity as to what provokes this behavior.

fredbi:
* Fixes #1637 (unmarshalling array of polymorphic type in body parameter)

> NOTE: polymorphism still gets a number of uncovered edge cases.
> This provides minimal patching for this particular case.

### DIFF
--- a/fixtures/bugs/1637/.gitignore
+++ b/fixtures/bugs/1637/.gitignore
@@ -1,0 +1,3 @@
+gen-*
+!gen-fixtures.sh
+*.log
--- a/fixtures/bugs/1637/fixture-1637.yaml
+++ b/fixtures/bugs/1637/fixture-1637.yaml
@@ -1,0 +1,44 @@
+swagger: '2.0'
+info:
+  version: '0.0.1'
+  title: Test
+schemes:
+  - http
+basePath: /v1
+produces:
+  - application/json
+consumes:
+  - application/json
+paths:
+  /test:
+    put:
+      operationId: test
+      parameters:
+      - name: payload
+        in: body
+        required: true
+        schema:
+          type: array
+          items:
+            $ref: '#/definitions/Value'
+      tags:
+        - Test
+      responses:
+        default:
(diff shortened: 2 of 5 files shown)


fredbi:
Again codecov report is out of this world... Removed 1 line => relative coverage decreases...

PR #1642, opened 2018-07-29
Improves error handling at startup for https server

### LINKED ISSUE #1418: Replace graceful with built-in methods (1.8 onwards)
## Problem statement
go-swagger's generate server uses a graceful.Server, the main functionality of which is now built-in to Go 1.8. Perhaps the dependency on graceful can be removed altogether ?

## Environment
swagger version: 2.0 / go-swagger dev
go version: 1.8  
OS:   NA

### LINKED ISSUE #1473: Server locks up on adding flags with go-flags
## Problem statement

When using **go-flags** I have run into the issue that when I try to use it, the server simply gets stuck adding it to go-flags.

```bash
func configureFlags(api *operations.NetroundsPocAPI) {
	api.CommandLineOptionsGroups = []swag.CommandLineOptionsGroup{
		swag.CommandLineOptionsGroup{
			ShortDescription: "Netrounds PoC Options:",
			LongDescription:  "",
			Options: struct {
				Example1 string `long:"example1" description:"Sample for showing how to configure cmd-line flags"`
			}{},
		},
	}
}
```

Which I assume is a bug?

### Using Pflags

So I tried switching to `--flag-strategy=pflag`

It would make sense in this mode to add a comment in `configureFlags` that it is not applicable (it will never be run by the generated `main.go`)

## Swagger specification

Not applicable

## Steps to reproduce

Use the above `configureFlags` example and things lock up.

## Environment
swagger version:
Used docker image:
```
quay.io/goswagger/swagger          latest              3694e4352134
```
Which used:
```
version: dev
commit: 2d443fa7eea569af9f20dd6f0302aef6c3b10651
```

go version:
```
go version go1.9.4 linux/amd64
```

OS: Linux (run via Docker on Windows)

### LINKED ISSUE #1558: Deadlock when TLS certificate is missing
## Problem statement
Server deadlocks when TLS ceretificate is missing. Deadlock happens at this line - https://github.com/go-swagger/go-swagger/blob/25e637c5028dee7baf8cdf5d172ccb28cb8e5c3e/generator/templates/server/server.gotmpl#L391

There is no error check when doing certificate load and parsing - https://github.com/go-swagger/go-swagger/blob/25e637c5028dee7baf8cdf5d172ccb28cb8e5c3e/generator/templates/server/server.gotmpl#L376

Adding error check that returns err solves the problem.

## Steps to reproduce
Run server with --tls-certificate pointing to non-existant file.

fredbi:
* Avoids panic when certificate or key passed is invalid
* Checks that CA certificate passed is correctly parsed

Acknowledges the following fixes from previous PR (#1627):

* fixes #1473
* fixes #1558
* fixes #1418 (from PR #1574)

NOTE: fixes are acknowledged using a test script which reproduce the issues.
However, this script has not been included in the CI suite.


generator/templates/server/server.gotmpl
```diff
@@ -361,51 +363,58 @@ func (s *Server) Serve() (err error) {
...
-      caCertPool := x509.NewCertPool()
-      caCertPool.AppendCertsFromPEM(caCert)
-      httpsServer.TLSConfig.ClientCAs = caCertPool
-      httpsServer.TLSConfig.ClientAuth = tls.RequireAndVerifyClientCert
-    }
+		if s.TLSCACertificate != "" {
+			// include specified CA certificate
+			caCert, caCertErr := ioutil.ReadFile({{ if .UseGoStructFlags }}string({{ end }}s.TLSCACertificate{{ if .UseGoStructFlags }}){{ end }})
+			if caCertErr != nil {
+				log.Fatal(caCertErr)
```
casualjim: this needs to use the log from the server not from the package

generator/templates/server/server.gotmpl
```diff
@@ -361,51 +363,58 @@ func (s *Server) Serve() (err error) {
...
+			}
+			caCertPool := x509.NewCertPool()
+			ok := caCertPool.AppendCertsFromPEM(caCert)
+			if !ok {
+				log.Fatal("cannot parse CA certificate")
```
casualjim: this needs to use the log from the server not from the package

generator/templates/server/server.gotmpl
```diff
@@ -361,51 +363,58 @@ func (s *Server) Serve() (err error) {
...
+			CurvePreferences: []tls.CurveID{tls.CurveP256},
+		{{- if .UseModernMode }}
+			// Use modern tls mode https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility
 			NextProtos: []string{"http/1.1", "h2"},
 			// https://www.owasp.org/index.php/Transport_Layer_Protection_Cheat_Sheet#Rule_-_Only_Support_Strong_Protocols
 			MinVersion: tls.VersionTLS12,
-      // These ciphersuites support Forward Secrecy: https://en.wikipedia.org/wiki/Forward_secrecy
+			// These ciphersuites support Forward Secrecy: https://en.wikipedia.org/wiki/Forward_secrecy
 			CipherSuites: []uint16{
 				tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
```
casualjim: can you add 

```
tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
```

to these ciphersuites since we're no longer supporting 1.7

generator/templates/server/server.gotmpl
fredbi: maybe even better to return an error... I followed the previous lines of code with a lot of Fatalf, but frankly I would prefer to replace all those by return err / return fmt.Errorf(xxx)
What do you think?

generator/templates/server/server.gotmpl
fredbi: ok

generator/templates/server/server.gotmpl
casualjim: these are fatal errors for the server, and I'd prefer no change to the public interface

gregmarr: Looks like most of this function does `return err` if it's in the main function, and `s.Fatalf` if it's in a goroutine.  The only difference is this section:
https://github.com/go-swagger/go-swagger/blob/master/generator/templates/server/server.gotmpl#L408-L418

gregmarr: and this line https://github.com/go-swagger/go-swagger/blob/master/generator/templates/server/server.gotmpl#L393

casualjim: in that case either way works, returning errors instead of exiting probably helps  people using this in tests.

generator/templates/server/server.gotmpl
fredbi: done

Issue #1629, opened 2018-07-20
Redirect sometimes fails in the go-swagger generated code


huikang:
## Problem statement

I used go swagger to generate the http handler code. One of the handler needs to call the http.Redirect(...) to redirect the request to another web page.

```go
http.Redirect(wG, r, redirectURL, http.StatusFound)
```

In order to get the writer, I added the following code in the go swagger middleware part

```
var wG http.ResponseWriter

// The middleware configuration is for the handler executors. These do not apply to the swagger.json document.
// The middleware executes after routing but before authentication, binding and validation
func setupMiddlewares(handler http.Handler) http.Handler {
	ourFunc := func(w http.ResponseWriter, r *http.Request) {
		wG = w
		handler.ServeHTTP(w, r)
	}
	return http.HandlerFunc(ourFunc)
}
```

However, sometimes the redirect fails to redirect the client browser to the redirect URL. Instead, the browser stays on the same page. It occurs occasionally. In addition, I saw the following log in the API server

```
2018/07/20 17:03:46 http: multiple response.WriteHeader calls
2018/07/20 17:03:46 http: multiple response.WriteHeader calls

```

Could anyone help look at this issue? Thanks.

## Swagger specification

version: 0.13.0
commit: 8135eb6728e43b73489e80f94426e6d387809502

fredbi:
I have hit this one too.

It is a warning issued by the http package.
It is because we always send a response in the end, even though the redirect status has already been set.
One should check for possible already redirected response before forcing a response.

huikang:
@fredbi , thanks for your explanation. Do you have any code snippet that fixes the redirect issue? Thanks in advance.

fredbi:
not yet

huikang:
The problem is that the http.Redirect method in Go does not provide any return value.

Also is it confirmed that the "http: multiple response.WriteHeader calls" causes the redirect failure?

casualjim:
do you have wg outside of the middleware like that?

huikang:
In the above code snippet, ``wG`` is a global variable and the only one. Because I need to use ``wG`` in the http.Redirect method.

casualjim:
yeah but how does that work with 2 concurrent requests?

huikang:
I thought the go-swagger generated code handles concurrent requests. For example, user 1 hits the redirect path of the server, which calls the redirect method to redirect to an external page.
Meanwhile, user 2 can access the same path.

[...]

PR #1657, opened 2018-08-11
Better example for redirect in operation

fredbi:
Better example code re question #1629

### DIFF
--- a/examples/oauth2/README.md
+++ b/examples/oauth2/README.md
@@ -182,9 +182,7 @@ func configureAPI(api *operations.OauthSampleAPI) http.Handler {
 	})
 
 	api.GetLoginHandler = operations.GetLoginHandlerFunc(func(params operations.GetLoginParams) middleware.Responder {
-        // implements the login operation
-		login(params.HTTPRequest)
-		return middleware.NotImplemented("operation .GetLogin has not yet been implemented")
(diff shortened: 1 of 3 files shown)


examples/oauth2/README.md
```diff
@@ -210,12 +208,12 @@ We set the following implementation for authentication in `restapi/implementatio
 - Redirecting to the login page
 
 ```go
-func login(r *http.Request) string {
-	// implements the login with a redirection and an access token
-	var accessToken string
-	wG := r.Context().Value(ctxResponseWriter).(http.ResponseWriter)
-	http.Redirect(wG, r, config.AuthCodeURL(state), http.StatusFound)
-	return accessToken
+func login(r *http.Request) middleware.Responder {
```
casualjim: I think the nicer example here would be by adding a response that is for code 302 and returning that instead.
That keeps it all typed. That response needs to define a location header

fredbi: This one was just to follow our previous discussion on handling redirects _with_ http.Redirect(),
and illustrating usage of the ResponderFunc()... and no more publish sample code with context here.

Of course, what you suggest is to avoid this call altogether and process redirection as a regular 302 response with header, etc. Since this significantly deviates from the "oauth2" example, I suggest this could be a complete dedicated example.

PR #1661, opened 2018-08-19
Improve go.mod support in generator

mikkeloscar:
This improves the `go.mod` support added in #1636 by adding support for
targets not at the module level.

Imagine you generate code like this:

```bash
$ swagger generate client --name my-api --target ./pkg/api
```

Then it would assume `go.mod` is under `pkg/api` and not detect that go
modules are used. The fix is to walk up the directory to find the first
occurance of `go.mod` and assume that as the module base URL.

### DIFF
--- a/generator/moreschemavalidation_test.go
+++ b/generator/moreschemavalidation_test.go
@@ -298,7 +298,7 @@ func TestMoreModelValidations(t *testing.T) {
 		fixtureSpec := fixture.SpecFile
 		runTitle := strings.Join([]string{"codegen", strings.TrimSuffix(path.Base(fixtureSpec), path.Ext(fixtureSpec))}, "-")
 		t.Run(runTitle, func(t *testing.T) {
-			t.Parallel()
+			//t.Parallel()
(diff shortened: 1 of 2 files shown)


mikkeloscar:
I don't understand why the build is failing, I can run the tests fine locally. Any ideas if it's related to my changes?

fredbi:
Looks related to some previous changes I made... I have to investigate there: a bug did sneak in undetected

fredbi:
I suggest as a temporary workaround to pass your PR to comment the "Parallel()" statement here https://github.com/go-swagger/go-swagger/blob/dd867fd63c30269ac217004c102f47a1774d4f5a/generator/moreschemavalidation_test.go#L301

I assume my latest attempt to improve this did introduce some smelly behavior.

You could try this and see if your pr passes. I'll try to fix things up later on.

mikkeloscar:
@fredbi commenting out the parallel config did make the build run successfully.

fredbi:
Found the concurrency bug in unit test. Going to fix it and specifically test for this...

PR #1709, opened 2018-09-19
Allow to include/exclude routes by their tags

scf2k:
--include-tag, --exclude-tag options are added to the spec generator. By specifying which it's possible to white/black list the route tags. 

Signed-off-by: Vladimir Alaev <scf2k4@gmail.com>


scan/routes.go
```diff
@@ -112,3 +119,21 @@ func (rp *routesParser) Parse(gofile *ast.File, target interface{}) error {
...
+			if includeTags[tag] {
+				return true
+			}
+		} else if len(excludeTags) > 0 {
+			if excludeTags[tag] {
+				return false
+			}
+		}
+	}
+	if len(includeTags) > 0 {
```
golangcibot: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>'

casualjim: this can be simplified to `return includeTags <= 0`

casualjim:
can you address the golang ci comment and sign your commit?

scf2k:
done

PR #1717, opened 2018-09-20
Fix do not generate models with x-go-type extension

### LINKED ISSUE #1710: Models rendered to file even if imported from external package using x-go-type
## Problem statement
After my changes in #1705 imports are working correctly now, but as a result of parsing whole hierarchy of models during `makeGenDefinitionHierarchy` (to retrieve all external packages and their aliases) instead of skipping ones with `x-go-type`, models are rendered to file even if they are imported from external package.

Unfortunately I have no more time for digging in generator code, but my idea for this is to add field called for example `external` of type `bool` and set it to true for models containing `x-go-type` extension, to mark models if they should be rendered to file or not.

## Environment
swagger version: master
go version: 1.11
OS: Ubuntu 18.04

moskal91:
Fixes #1710

### DIFF
--- a/generator/types.go
+++ b/generator/types.go
@@ -342,13 +342,6 @@ func (t *typeResolver) resolveArray(schema *spec.Schema, isAnonymous, isRequired
 	result.IsNullable = false
 	result.IsEmptyOmitted = t.IsEmptyOmitted(schema)
 
-	if !isAnonymous {
-		tpe, pkg, alias := knownDefGoType(t.ModelName, *schema, t.goTypeName)
-		result.GoType = tpe
-		result.Pkg = pkg
-		result.PkgAlias = alias
-	}
-
 	if schema.AdditionalItems != nil {
 		result.HasAdditionalItems = (schema.AdditionalItems.Allows || schema.AdditionalItems.Schema != nil)
 	}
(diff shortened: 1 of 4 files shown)


casualjim:
Change looks good but can you sign and sign-off on the commit?

sign off:
https://github.com/go-swagger/go-swagger/pull/1717/checks?check_run_id=18650494

sign:

generator/types.go
```diff
@@ -342,13 +342,6 @@ func (t *typeResolver) resolveArray(schema *spec.Schema, isAnonymous, isRequired
 	result.IsNullable = false
 	result.IsEmptyOmitted = t.IsEmptyOmitted(schema)
 
-	if !isAnonymous {
```
fredbi: This section is rather sensitive.
Could you please explain why you needed to remove this?

moskal91: This was added in my previous PR and I'm not sure what it exactly does, so reverted it to original state for now, as it does not affect result of my case.

fredbi: Be extremely cautious here, because types maybe primitives with pointers. There are some flaws here and it is difficult to modify this without changing many other stuff.

casualjim: This was code that was added by @moskal91 in his previous PR and now removed.
There is no risk afaik with this

Issue #1573, opened 2018-06-13
Generated code don't pass on go vet


si-m:
## Problem statement
I am getting this vet response:
```
service/restapi/server.go:104: can't check non-constant format in call to Printf
service/restapi/server.go:115: can't check non-constant format in call to Fatalf
```
## Steps to reproduce
```sh
go vet -v $(go list ./...)
```
## Environment
swagger version: 0.14.0  
go version: go1.10.1
OS:  macOS High Sierra 10.13.4
#

fredbi:
This is a warning croaked by vet in verbose mode: the vet command return code is still a success.

To remove this check: `go vet -v -printf=false`

Also vet is changing in go, and there are some bugs reported in go 11 vet command, specifically on this "printf" check.

fredbi:
Reminder: should be tested again with go1.13

fredbi:
Closed: go vet has evolved and does not produce this warning any longer. Tested using go1.13

PR #1723, opened 2018-09-22
use different bindata binary

casualjim:
updates to build with go 1.11
fixes a number of shellcheck issues


fredbi:
this new --vet off option... does it mean that our generated stuff does not pass vet with go 11?

fredbi:
Can't merge PRs any longer.

casualjim:
That is correct because of the debugLog and debugLogAsJSON methods, the format strings are of wrong type and/or missing a format param.

fredbi:
We have the same kind of construct in generated code -see issue #1573.
IMHO go vet just got overboard with 1.10

Issue #1714, opened 2018-09-20
un-graceful shutdown after 15 seconds on generated restapi/server.go#handleShutdown


dehoust:
In handleShutdown a context with a constant timeout of 15 seconds is passed on to server.Shutdown

https://github.com/go-swagger/go-swagger/blob/32f94c7110162041004c3c54a2f3cba41c6d205c/generator/templates/server/server.gotmpl#L541
...
https://github.com/go-swagger/go-swagger/blob/32f94c7110162041004c3c54a2f3cba41c6d205c/generator/templates/server/server.gotmpl#L552

which leads to an un-graceful shutdown after 15 seconds with "context deadline exceeded", closing also active connections.

We have to deal with longer living active connections. Please make that timeout a parameter that can be set in configureServer on restapi/configure_.

fredbi:
Fixed with #1730, #1732

PR #1732, opened 2018-10-01
Fix graceful-timeout variable

gregorhollmig:
During enhancing go-swagger generator as mentioned in issue #1714, as an accident in #1730 the command line parameter `--graceful-timeout=50s` was interpreted as float and for `WithTimeout` multiplied by `*time.Second`

The code doesn't fail but `context deadline exceeded` will be immediately executed.

This fixes the issue.


fredbi:
Hello. How come you can update a template and commit without regenerating bindata.go ?

fredbi:
Yes it should. There is a script .githools/pre-commit/50xxx for that

fredbi:
Ah, again the commit signing involved... Best to rebase and sign in one pass (_every commit_ is required to be signed...)

gregorhollmig:
Is further action required by my side for the PR to be merged? 
The signing is sth. I haven't done before, I need to get used to that...

fredbi:
this checks _all_ your commits...

casualjim:
one is sign-off the other is sign with pgp key.
You're missing the sign with pgp key. Just the last commit will be fine.

casualjim:

casualjim:
git hooks don't get installed automatically you have to enable them manually first and then it should start to work.

gregorhollmig:
Okay @casualjim  so ... I signed with `commit -s` isn't that fine? I believe I did the same for #1730 . DCO check is also fine, what is the issue?

fredbi:
@gregorhollmig  -s only add your email to the commit text.  I think the new CI check also verifies that your commit is _cryptographically_ signed with a PGP signature. You can check this for example, when doing `git log --show-signature`. This means configuring PGP keys, and configuring git to use these keys to PGP-sign your commits.

gregmarr:
@casualjim there is no mention in CONTRIBUTING.md that PGP signature is required.  Is this an intentional change?  If so, it should be added there.

fredbi:
Yep, this came a bit unannounced... This could also be detailed here: https://github.com/go-swagger/go-swagger/tree/master/docs/guidelines

casualjim:
This PR is not the right place to have this discussion

PR #1744, opened 2018-10-07
Handle relative path when using Go modules

mikkeloscar:
This is a follow up to #1661 which correctly handles relative paths when using go modules.

#1661 didn't account for situations where you want to generate server code in a sub package of your module, so it generated the wrong imports because it used the go module path as base import instead of go module + relative package.

E.g. if you had a module `github.com/a/b` and wanted to generate go-swagger code from the package `github.com/a/b/my-server` then the import paths would be `github.com/a/b/restapi/..` instead of `github.com/a/b/my-server/restapi/..`.

### DIFF
--- a/generator/shared.go
+++ b/generator/shared.go
@@ -250,14 +250,14 @@ func GoLangOpts() *LanguageOpts {
 
 		}
 
-		mod, err := tryResolveModule(tgtAbsPath)
+		mod, goModuleAbsPath, err := tryResolveModule(tgtAbsPath)
 		switch {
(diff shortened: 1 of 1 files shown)


casualjim:
This makes generation go a lot slower for some reason

mikkeloscar:
I have noticed this as well. I don't think it's related to this particular change but I suspect it has to do with using Go modules in general. I assume when using Go modules you hit the network a lot when invoking go during the generation.

I have not yet verified this though.

PR #1749, opened 2018-10-08
Updated contributing guidelines

fredbi:

### DIFF
--- a/docs/guidelines/README.md
+++ b/docs/guidelines/README.md
@@ -6,14 +6,23 @@ A quick guide on how to contribute to `go-swagger` and the other `go-openapi` re
 
 Repos follow standard go building and testing rules.
 
+Cloning `go-swagger`:
+```bash
+mkdir -p $GOPATH/src/github.com/go-swagger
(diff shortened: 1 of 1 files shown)


docs/guidelines/README.md
```diff
@@ -23,16 +32,17 @@ More advanced tests are run by CI. See [below](#continuous-integration).
 All PR's are welcome and generally accepted (so far, 95% were accepted...).
 There are just a few common sense rules to be followed.
 
-1. PRs which are not ready to merge should be prefixed with "WIP:"
-2. Generally, contributors should squash their commits (use `git rebase -i ...`)
+1. PRs which are not ready to merge should be prefixed with `WIP:`
+2. Generally, contributors should squash their commits (use `git rebase -i master`)
```
casualjim: this is no longer required, we can squash at merge time

fredbi: Yes. Hence the "generally" and "should" rather than "always" "must"... It remains a fair practice since it forces contributor to select relevant commit messages.

PR #1780, opened 2018-11-01
fixes #1778. Allow 'result' name for API parameter

### LINKED ISSUE #1778: non-compiling code if an API method contains a parameter named 'result'
## Problem statement

Compilation of generated code fails if an API method contains a parameter named 'result'.

## Swagger specification

```
  /some/api/path:
    get:
      description: "Something"
      operationId: SomeOperation
      tags: ['sometag']
      produces:
      - application/json
      parameters:
      - {$ref: '#/parameters/page'}
      - {$ref: '#/parameters/pagesize'}
      - name: type
        in: query
        description: some description of type parameter
        type: array
        items:
          type: string
        collectionFormat: csv
      - name: result
        in: query
        description: some description of result parameter
        type: array
        items:
          type: boolean
        collectionFormat: csv
```

## Steps to reproduce

Generate the spec and try to compile the code.

## Environment
swagger version: 0.17.2  
go version: 1.11.1  
OS:   Windows 10, Ubuntu 14

## Additional
Problem lies in method Build() :
```
result.Path = golangswaggerpaths.Join(_basePath, _path)
...
result := swag.JoinByFormat(resultIR, "")

	if len(result) > 0 {
		qsv := result[0]
		if qsv != "" {
			qs.Set("result", qsv)
		}
	}
```

mtfelian:


mtfelian:
Fixes #1778

generator/templates/server/urlbuilder.gotmpl
```diff
@@ -139,7 +139,7 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}URL) SetBasePath(bp string) {
 
 // Build a url path and query string
 func ({{ .ReceiverName }} *{{ pascalize .Name }}URL) Build() (*url.URL, error) {
-  var result url.URL
+  var _result url.URL
```
casualjim: I think it's better to make a combination here or ensure that the identifier is unique, has some randomness to it.
Otherwise we'll end up in the same situation as before because the magical string has changed from result to _result

fredbi: problem with randomness is that generation is no more stable, which is another problem...

fredbi: for my part, the `_result` is good enough. It's no silver bullet, but reduces the likelyhood of conflict.

gregmarr: The `_result` also matches `_path` and `_basePath` in the function.  If a combination (nothing random please) were used, such as `{{ .ReceiverName }}_{{ . Name }}_result` then it should be added to those other variables as well.

casualjim: yes sure, I just meant not `s/result/_result/` something that has a much higher likelihood of being unique

casualjim:
can you do this sign off on the commit? ```git commit --amend --signoff```

generator/templates/server/urlbuilder.gotmpl
mtfelian: @fredbi released new bindata.go, so what we will do about _result? I agree that if we change this in some manner, other variables should be changed also in same manner. I thought that variables named starting with _ deliberately to avoid conflicts. May be should forbid any property name starting with _ ? Any way it is a bad practice to use such names.

fredbi:
Looks good to me

fredbi:
@mtfelian  could you please rebase your PR, so we may proceed with the merge? Thank you

mtfelian:
@fredbi rebased the branch sorry for the delay.

fredbi:
Thanks. Almost there! But I can't merge... I assume this is because of the missing GPG-signature

Issue #1665, opened 2018-08-23
can't support annotation multiple - lines


liyuankui:
## Problem statement

```
// swagger:parameters getFooByID getFooFooByID
// swagger:parameters getBarByID getBarBarByID 
type idParams struct {
	// in: path
	ID string `json:"id"`
}
```

Expect: the param valid for all 4 APIs. 

As: only Foo* can get this param. 

Motivation:
1. common params like `id` may be used by many many many APIs. 
2. the line will be very lengthy and also easily causing conflicts during git collaboration
3. If multiple line annotation can be recognized, then each API can have it's own line to declare

## Swagger specification

parameters

## Steps to reproduce

compile

## Environment
swagger version: dev
go version: 1.10
OS:   Darwin 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64

casualjim:
You can change that here: https://github.com/go-swagger/go-swagger/blob/master/scan/parameters.go#L171-L187

feel free to send a PR, i think it shouldn't break and then it should probably be fine.

PR #1783, opened 2018-11-05
Absorbing api names across multiple lines

liyuankui:
#1665


casualjim:
Can you add a test that demonstrates it behaves as expected?

liyuankui:
@casualjim tests added, thanks a lot for the check

PR #1788, opened 2018-11-09
fix client cannot unmarshal struct which contains null

### LINKED ISSUE #1729: Client cannot unmarshal struct
## Problem statement
Let's say, I have a struct called Pet
```go
type struct Pet{
    Type string
    Name string
}
```
I also have a struct called Person
```go
type struct Person{
    Name string
    Pet Pet
} 
```
I get the Person via API from server, the json looks like
```json
{
"name":"Alex"
"pet":null
}
```
Then I use the client of go-swagger to unmarshal the json, I will get runtime panic.
Since the field 'Type' of Pet is required in my definition.  When it unmarshal the value of 'pet', it will validate if the value of 'Type' is exist, if it's not exist, runtime panic happens.  Actually, the value is null, so the client cannot get a value for the field 'Type'.

Do you think my scenario is reasonable?

## Swagger specification
Swagger 2.0
## Environment
swagger version: 0.15.0
go version: 1.10.3
OS: Mac OS

AlexInGitHub:
fixes #1729


casualjim:
Can you run: `go generate ./generator` . that way bindata.go will get updated 
We use this bindata: https://github.com/kevinburke/go-bindata

Can you also sign off on your commit? https://github.com/go-swagger/go-swagger/pull/1788/checks?check_run_id=30358019

casualjim:
Can you please run: `go generate ./generator`? 
That way generator/bindata.go will get updated, and your changes will actually get picked up.
We use this bindata: https://github.com/kevinburke/go-bindata

AlexInGitHub:
Done, thanks

AlexInGitHub:
Hi @casualjim, Does the failed test cases means I don't have the permission to update the template?

casualjim:
I think irrespective of the naming discussion, the test isn't passing, so the test needs an updated assertion. You can run the tests with `go test ./generator/...`

AlexInGitHub:
I updated the cases, could you help to review and merge the code ?

Issue #1624, opened 2018-07-18
Can't have server code in subdirectory


maqdev:
## Problem statement

When passing -s argument as below which includes /, I was hoping that would allow the server code to be in a subdirectory:

`swagger generate server -f ./api/swagger/swagger.yaml -s pkg/srv`

However the import statements, for example in `configure_srv.go` are snake casing the package name:

`"github.com/xxx/app/pkg_srv/operations"`

## Swagger specification
2.0

## Environment
swagger version: 0.15.0
go version: 1.10.3
OS:  OS-X

fredbi:
The generated code is put in a hierarchy of packages which may be customized with the
following options:
```sh
      -a, --api-package=                                                          the package to save the operations (default: operations)
      -m, --model-package=                                                        the package to save the models (default: models)
      -s, --server-package=                                                       the package to save the server specific code (default: restapi)
      -c, --client-package=                                                       the package to save the client specific code (default: client)
      -t, --target=                                                               the base directory for generating the files (default: ./)
```

The organization is determined as follows:
```
-- {target}
    -- {client-package:-client}
    -- {model-package:-models}
    -- {server-package:-restapi}
       -- {api-package:-operations}
          -- {tags}
```

You may customize the folder names, but can't specify sub-packages as package names.

gregmarr:
We fixed this issue for models in December: https://github.com/go-swagger/go-swagger/issues/1326

gregmarr:
Commit is here https://github.com/go-swagger/go-swagger/pull/1329/commits/ae57fbc6faae9e97310ca7d96f7d8a45d861e48b

maqdev:
@fredbi actually you can specify sub-package for the models as mentioned by @gregmarr.
A workaround is to create a sub-directory, then run swagger inside it and it generates correct package names.

PR #1790, opened 2018-11-09
WIP: Support server generation in a subdirectory #1624

MrLuje:
example : `swagger generate server -f ./api/swagger/swagger.yaml -s pkg/srv`

`configure_srv.go` will have **srv** as package name (last part of **pkg/srv**)
Imports to operations will be like **[...]/pkg/srv/operations**

I added a test but I guess it won't be run for now and I'm not totally sure how I should reference it, from a go file or a script, like codegen-nonreg.sh

### DIFF
--- a/fixtures/enhancements/1624/.gitignore
+++ b/fixtures/enhancements/1624/.gitignore
@@ -1,0 +1,3 @@
+gen-*
+!gen-fixtures.sh
+*.log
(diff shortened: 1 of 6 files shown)


fredbi:
The problem is more complicated than that. This PR addresses one very particular use case, at the cost of more inconsistent behavior.

MrLuje:
@fredbi No problem, I should have asked about it before starting anything

PR #1801, opened 2018-11-17
Use JSON for examples of type object

leon-barrett:
The OpenAPI 2.0 spec allows examples to be any JSON value. Currently,
go-swagger supports most types (e.g. ints and strings), but not objects.
This adds support to parse object examples as JSON, allowing users to
document examples for maps.


fredbi:
Could you please add some unit test to assert this behaves as expected?

casualjim:
there is a test already

PR #1806, opened 2018-11-18
Deprecates `--with-context` option

### LINKED ISSUE #1684: swagger generation with --with-context result in compiles error
## Problem statement
RE generate the code for example custom-server with context results in compilation error.
There is missing returned http.Request for generated code and missing Authorizer method.

Compilation Error:
# github.com/go-swagger/go-swagger/examples/tutorials/custom-server/gen/restapi/operations
examples/tutorials/custom-server/gen/restapi/operations/get_greeting.go:43:12: assignment mismatch: 2 variables but 3 values
examples/tutorials/custom-server/gen/restapi/operations/greeter_api.go:190:46: cannot use o (type *GreeterAPI) as type middleware.RoutableAPI in argument to middleware.NewRoutableContext:
	*GreeterAPI does not implement middleware.RoutableAPI (missing Authorizer method)

Even after manually fixing that, following error
examples/tutorials/custom-server/cmd/greeter/main.go:39:60: cannot convert func literal (type func(operations.GetGreetingParams) middleware.Responder) to type operations.GetGreetingHandlerFunc

## Swagger specification
Just using the custom-server in example
https://github.com/go-swagger/go-swagger/blob/master/examples/tutorials/custom-server/swagger/swagger.yml

## Steps to reproduce
server:
rm -rf gen && swagger generate server --exclude-main -A greeter -t gen -f ./swagger/swagger.yml --with-context
client:
swagger generate client -A greeter -t gen -f ./swagger/swagger.yml

## Environment
swagger version: dev
go version: go1.10.2 darwin/amd64
OS:   
macOS

fredbi:
The command line flags remains active (older CLI scripts still work),
but the flag is ignored.

Templates are cleaned up from this unsupported variant, including stratoscale.

* fixes #1684


cmd/swagger/commands/generate/server_internal_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestDeprecated(t *testing.T) {
+	var buf bytes.Buffer
+	log.SetOutput(&buf)
+	defer log.SetOutput(os.Stdout)
+	s := Server{
+		WithContext: true,
+	}
+	s.getOpts()
```
golangcibot: Error return value of `s.getOpts` is not checked

jawspeak:
Sorry for commenting on an old PR, but SEO results were not helpful to figure this out. This is a top result when searching for the deprecated command line flag `--with-context` when generating server handlers.

What's the way to generate handlers such that they can access the current request context?

casualjim:
You can get the request context from the `HTTPRequest` property on the generated `XxxParams` struct. You have full access to the `net/http.Request` there

jawspeak:
Amazing! Thanks again @casualjim !

PR #1807, opened 2018-11-19
Handle array examples; document Example usage

leon-barrett:
I previously added JSON parsing for examples on object types; this adds
it for array types. It also adds basic documentation for Example.


docs/use/spec/model.md
```diff
@@ -74,6 +75,7 @@ type User struct {
 	// the email address for this user
 	//
 	// required: true
+        // example: user@provider.net
```
casualjim: How come this alignment is off ?

Issue #1816, opened 2018-11-27
generated embedded_spec.go removing refs


kenjones-cisco:
## Problem statement

Starting around the end of the week of 11/16 when generating code, I'm noticing that the embedded_spec.go is no longer include the ref paths, they are being set to `#`.

Examples:

```
           "400": {
-            "$ref": "#/responses/400"
+            "$ref": "#"
           },
           "401": {
-            "$ref": "#/responses/401"
+            "$ref": "#"
           },
           "403": {
-            "$ref": "#/responses/403"
+            "$ref": "#"
           },
           "404": {
-            "$ref": "#/responses/404"
+            "$ref": "#"
           },
           "default": {
-            "$ref": "#/responses/500"
+            "$ref": "#"
           }
         },

             "schema": {
               "type": "array",
               "items": {
-                "$ref": "#/definitions/service_view"
+                "$ref": "#"
               }
             },

           "200": {
             "description": "Service",
             "schema": {
-              "$ref": "#/definitions/service_view"
+              "$ref": "#"
             },

             "schema": {
               "type": "array",
               "items": {
-                "$ref": "#/definitions/unit_view"
+                "$ref": "#"
               }
             },

           "200": {
             "description": "Service Unit",
             "schema": {
-              "$ref": "#/definitions/unit_view"
+              "$ref": "#"
             },

       "properties": {
         "audit": {
-          "$ref": "#/definitions/audit"
+          "$ref": "#"
         },

           "type": "object",
           "title": "Service Unit Quotas",
           "additionalProperties": {
-            "$ref": "#/definitions/quota"
+            "$ref": "#"
           }

     "400": {
       "description": "Invalid Request Error. Check error message in the response body for details.\n",
       "schema": {
-        "$ref": "#/definitions/error"
+        "$ref": "#"
       },

```

## Swagger specification

```
swagger: '2.0'
info:
  version: "1.0.0"
  title: Service Manager

schemes:
  - http

produces:
  - application/vnd.cia.v1+json

tags:
  - name: services

basePath: /v1

paths:
  /services:
    parameters:
      - name: "X-Request-Id"
        in: "header"
        description: Unique identifier associated with request
        required: true
        type: string
        format: uuid
    get:
      tags: [services]
      operationId: searchServices
      summary: Search for Services
      description: |
        To apply filter criteria use the following approach:

        `?<field><operator><value>`

        Example: `?name=Test`

      parameters:
        - name: name
          in: query
          description: Search records by name
[...]

PR #1817, opened 2018-11-27
Task: Regenerate examples

kenjones-cisco:
#1816


examples/authentication/restapi/embedded_spec.go
```diff
@@ -46,33 +46,33 @@ func init() {
             "name": "info",
             "in": "body",
             "schema": {
-              "$ref": "#/definitions/social_id"
```
kenjones-cisco: Example of issue reported in #1816 
Just by regenerating the examples from master

kenjones-cisco:
I will happily regenerate again once the bug has been fixed. This seemed like the best approach to clearly demonstrate the issue being seen.

casualjim:
Yes you're right. it's good

examples/authentication/restapi/embedded_spec.go
fredbi: Thanks for reporting this

casualjim:
Can you update and regenerate with the code from master?

kenjones-cisco:
Regenerated and the the issue has been corrected.

fredbi:
Great!

fredbi:
@kenjones-cisco I think the  `[]` which now becomes `null` is a new regression failure, probably with the causes as the issue you detected. If you agree, I'll try and fix this new one over the week end.

kenjones-cisco:
Agreed, I'm good with waiting until after the new update.

thanks @fredbi !

fredbi:
After a quick look, I confirm this is _exactly_ the same cause: this new gob encode/decode.
Now I realize every spec object with special json unmarshalling needs to be checked if in need of a gob encoder and decoder or we'll have the same issue with other edge cases.

fredbi:
@kenjones-cisco I finished the update with PR #1837. I kept your own commit and just applied the fix.

PR #1821, opened 2018-11-29
Add HandlerAPI() function to stratoscale template

maxatome:
It allows to access API instance.

The use case is to allow logging of route pattern (eg. `/pet/{id}`).

The route pattern can be retrieved using the following code, but the API instance is needed:
```go
route, ok := api.Context().LookupRoute(req) // req is a *http.Request
if ok {
  return route.PathPattern // /pet/{id} here
}
```

Note: I `go generate`d generator/bindata.go but did not `go fmt`ed it coz it seems it was not before...


fredbi:
@posener, @michaelf-stratoscale  could you please review this PR?

michaelf-stratoscale:
@maxatome i don't understand where are you going to use this `api`, this file is auto generated and shouldn't be changed, so when you return it to `Handler` you are not using it.

maxatome:
@michaelf-stratoscale in my main() func I have:

```go
handler, err := restapi.Handler(restapi.Config{...})
if err != nil {
  logrus.Fatal(err)
}
```
then I chain this handler to a more generic one that handles authentication, request signature, logs and so on:
```go
handler = GenericMiddleware(handler, GenericMiddlewareConfig{...})
```
In this generic middleware, I log incoming requests, their method, path, status, duration.

Later I can analyse these logs to know how much time is spend in this method or this other one...

But the Request.Path is not a good input for comparing routes amongst themselves. On the other hand, the route.PathPattern is a good input, as it does not depend on user parameters.

To access this route.PathPattern, I need the api.Context() and so the api instance.

In my case, with this PR, I could do:
```go
handler, api, err := restapi.HandlerAPI(restapi.Config{...})
if err != nil {
  logrus.Fatal(err)
}
handler = GenericMiddleware(handler, GenericMiddlewareConfig{
  CanonicalizePath: func(req *http.Request) string {
    route, ok := api.Context().LookupRoute(req) // req is a *http.Request
    if ok {
      return route.PathPattern // /pet/{id} here
    }
    return ""
  },
  ...
})
```
So my log feature in my generic middleware just have to call the CanonicalizePath closure with the request to get the path pattern from the request it currently handles...

Perhaps you know another way to achieve this?

michaelf-stratoscale:
@maxatome please make sure it don't break existing functionality

maxatome:
@michaelf-stratoscale I tested it on my side and I didn't see any problem.

PR #1822, opened 2018-11-29
Fix paths when looking for packages in vendor dir

mhr3:
Was hitting `can't determine selector path from ...` error when trying to use a field from a vendored import path, this fixed the issue


casualjim:
This is superseded by #1991

PR #1827, opened 2018-12-01
Fix #1824

### LINKED ISSUE #1824: Regression with gob encoder optimization
## Problem statement

Since go-openapi/loads adopted gob to deep clone specs (commit: https://github.com/go-openapi/loads/pull/19/commits/0d17535daa37578ec3b60494288bd3164cb9a36b), some regressions occurred on spec objects which have custom json marshalling.

A first regression on "$ref" (bug: #1816) has been fixed with #1818.

Another one on "security" object has been found when analyzing #1817.

We need to add some more GobEncode/GobDecode in go-openapi/spec to ensure this optimization causes no more regression.

## Swagger specification
https://github.com/go-swagger/go-swagger/blob/master/examples/generated/swagger.json

## Steps to reproduce
Regenerate examples, observe diff.
Observe that:
```yaml
security: 
  api_key: []
```

becomes:
```yaml
security: 
  api_key: null
```
## Environment
swagger version:  current master
go version: 1.11
OS:   ubuntu

fredbi:
* fixes #1824

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -150,11 +150,11 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:a35c031b701153df4bd35671a1892ebd460dc65247f4f8f14c817578fa3d4216"
+  digest = "1:54af405aae840f810418f3d25211012af1c4bfd56d07d3de6482a7a3b762528a"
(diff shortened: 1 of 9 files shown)


fredbi:
should be good to go now

Issue #1828, opened 2018-12-04
generate/spec: route responses tag description


d7561985:
## Problem statement
Using swagger generate spec from go code.
While `swagger:route` filling `Responses` case  and use `description` we get no valid swagger code without description

## Steps to reproduce
```
// swagger:route DELETE /cats Cats deleteCats
//  Delete
//  Delete cat
//
//     Responses:
//       200: description:Message Success
```
expect
```
responses:
  "200":
     description: Message Success
```
but  got
```
responses:
  "200":
     $ref: '#/responses/'
```
## Environment
swagger version: current master  
go version: 1.11
OS:   macOS

fredbi:
Duplicates #989

PR #1829, opened 2018-12-04
route: add support description response tag at path

d7561985:
https://github.com/go-swagger/go-swagger/issues/1828


casualjim:
would it be possible to add a test for this behavior?

d7561985:
i also think about it. i'll do it on this week.

d7561985:
@casualjim done.

PR #1841, opened 2018-12-21
WIP: Onboards go-openapi/spec refactoring

fredbi:


gregmarr:
Is this PR just to get the CI stuff?  Doesn't seem like we'd want to commit using a fork of spec instead of the main repo.

fredbi:
Sure @gregmarr, don't worry! This is just to verify everything is okay before merging the spec repo (which is also still WIP). When we are okay with tests and merged dependency, this would just go back to original Gopkg.tom.

PR #1849, opened 2018-12-28
Models: add support for $ref indirections

### LINKED ISSUE #1260: Responses defined in schema: restapi/operations/xxx_responses.go:54:18: cannot make type models.YY
## Problem statement
Server generation.

When response schema is specified by $ref which defines an array structure, the generated model is not correct and does not compile, with a "cannot make [type]" error.
Namely, the initialization of the structure ( {{ Schema.Zero}} initialization method) does not reflect the generated nature of the type.

- generated type with $ref uses an 'OAI' type alias and a struct:
```
type xxx struct {
   OAIxxx 
}
```
- generated OAI alias:
`type OAIxxx []string`

When no ref is used (e.g. array declared inline in the definitions of the 
swagger file), there is no "OAI" alias type and compilation proceeds normally.

When the type is an object and not an array, the generator recognize this type as a complex type and does not attempt to use the make built-in to initialize it.

The problem really here is to assimilate type X and type Y struct { X }, which the compiler does not accept.

One would expect the code generated with $ref to be identical to the one with inline schema.

## Swagger specification
2.0

## Steps to reproduce
swagger specification YAML
[test3-ter-swagger.yaml.txt](https://github.com/go-swagger/go-swagger/files/1462315/test3-ter-swagger.yaml.txt)

JSON schema referred to
[test3-ter-model-schema.json.txt](https://github.com/go-swagger/go-swagger/files/1462317/test3-ter-model-schema.json.txt)

```
swagger generate server -f ../../models/test3-ter-swagger.yaml
```

Move to cmd/{server} directory then:
```
>go build
# .../restapi/operations
../../restapi/operations/get_all_responses.go:54:18: cannot make type models.B
```
## Environment
swagger version: build from source origin/master
```
>git describe
0.12.0-49-g783e5bb
```

go version: 1.9
OS:   linux ubuntu.14.04

fredbi:
* introduces golang _aliased_ type declarations (e.g. type A = B) to support definitions with indirections as $ref
* fix response template to accomodate these situations
* special provisions are taken for aliases on polymorphic types: use golang aliased type syntax, but redefine unmarshalling factories

* fixes #1260

> **NOTES**:
> For complex (non-polymorphic) types, the type composition construct (e.g. type A struct { B }) remains unchanged.
>
> For simple types (slices, maps, primitives) or polymorphic types, golang type aliases are used.
> See fixtures/bugs/1260/fixture-realiased-types.yaml for examples.

**Generated code using this feature requires go1.9+**


erikh:
is pointer access/nullable support in this patch? I am beating my head against the fixed models atm not doing this.

fredbi:
No there is no such things. Just felt like I had to fix one of the very first issues I raised... not related with nullable whatsoever

PR #1855, opened 2019-01-11
Add test case without output file for expand.go

saromanov:
Hi! I've just added one test case for cover the brand where output file is not defined on `expand.go`. Please review this


fredbi:
@saromanov could you please PGP-sign your commit?

saromanov:
> @saromanov could you please PGP-sign your commit?

@fredbi Signed it

PR #1872, opened 2019-02-05
Added --skip-validators

tandr:
Added a --skip-validators to "generate models|client|server" commands, so generated code does not contain "Validate()" function. 

This allows to generate significantly simpler code, and allows easier extraction of dependencies for generated code.

### DIFF
--- a/cmd/swagger/commands/generate/client.go
+++ b/cmd/swagger/commands/generate/client.go
@@ -32,6 +32,7 @@ type Client struct {
 	DefaultProduces string   `long:"default-produces" description:"the default mime type that API operations produce" default:"application/json"`
 	SkipModels      bool     `long:"skip-models" description:"no models will be generated when this flag is specified"`
 	SkipOperations  bool     `long:"skip-operations" description:"no operations will be generated when this flag is specified"`
+	NoValidators    bool     `long:"skip-validators" description:"no validators will be generated when this flag is specified"`
 	DumpData        bool     `long:"dump-data" description:"when present dumps the json for the template generator instead of generating files"`
 	SkipValidation  bool     `long:"skip-validation" description:"skips validation of spec prior to generation"`
 }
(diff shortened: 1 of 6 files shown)


fredbi:
We had deprecated this option some months ago, I just can't remember exactly why.

What I can say however, is that there are some traps there and this would require some more thorough testing to make sure this really works as expected.

I can see there are a lot of missing parts in this PR. As it is, it would only introduce bugs.

For instance: https://github.com/go-swagger/go-swagger/blob/81789e7a15fdae4906c34b64e48d8bd8bc0cc16c/generator/operation.go#L1073
And also: https://github.com/go-swagger/go-swagger/blob/81789e7a15fdae4906c34b64e48d8bd8bc0cc16c/generator/templates/server/parameter.gotmpl#L7

Well, I'll let @casualjim figure out if the intent shall be retained or not. What I can say is that this can only be perceived as incomplete and VIP.

tandr:
Thank you for the quick run down. I did not know about preexisting option that was removed, but since I need that functionality (otherwise I wouldn't bother :) ), I will just keep it for myself.

Ideally I would love an option to generate just "plain old Go struct" + maybe serialization code without  too many dependencies to pull. With this change as it is I can manually remove remaining validation code (couple lines, but I think I can modify tmpl to do it automatically if skip-validators supplied) and reduce dependencies (for example, drop strfmt) even more.

fredbi:
I see. What you have made works specifically for this use case: generating barebones models.

But it cannot be merged as-is: there are many other use-cases when generating servers and clients.

Anyway, you got the toolkit, so you may freely tamper with templates as you see fit.

tandr:
"Barebone models" is a very good definition of what I actually needed, thank you.

If I may ask - what are these "other use-cases" you have mentioned?

fredbi:
With the examples given above, it seemed obvious to me that this PR would make client and server generation fail in many cases (models would not implement the Validatable interface and the templates currently don't skip this). Idem for schemas directly generated in the operations package (not in models). 

So at best, this PR would cover the `generate model` case.

Even there, one of the tricky undertakings of the model generation is to determine when to map attributes as pointers. This is essentially linked to validation rules. So I assume that a model without validation should not have pointers. Unfortunately, setting this flag is not enough for this.

PR #1876, opened 2019-02-06
Add support for property extensions

### LINKED ISSUE #1875: Support for parsing property vendor extensions
Hi !

Would be awesome to have support for property vendor extensions !

It would allow more flexibility and is widely supported in the swagger/openAPI community. (For example, by the go-openapi package we rely on here)

For now, this package only supports automatic x-go-name and a few other hard coded extensions, but support for custom extensions would allow people to take control of this.

**Example :**
```
type User struct {
	// the name for this user
        //
        // Extensions:
	// ---
	// x-property-value: value
	// x-property-array:
	//   - value1
	//   - value2
	// x-property-array-obj:
	//   - name: obj
	//     value: field
	// ---
	Name string `json:"name"`
}
```
**Result :**
```
definitions:
  User:
    type: object
      name:
        description: the name for this user
        type: string
        x-property-value: value
        x-property-array:
          - value1
          - value2
        x-property-array-obj:
          - name: obj
            value: field
```
PR incoming 😄 🎁

fdelayen:
fixes #1875 😄

### DIFF
--- a/scan/schema_test.go
+++ b/scan/schema_test.go
@@ -60,6 +60,21 @@ func TestSchemaParser(t *testing.T) {
 	assert.False(t, prop.ExclusiveMinimum, "'score' should not have had an exclusive minimum")
 	assert.Equal(t, 27, prop.Example)
 
+	expectedNameExtensions := spec.Extensions{
+		"x-go-name": "Name",
+		"x-property-array": []interface{}{
+			"value1",
+			"value2",
+		},
+		"x-property-array-obj": []interface{}{
+			map[string]interface{}{
+				"name":  "obj",
+				"value": "field",
+			},
+		},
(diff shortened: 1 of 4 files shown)


scan/schema_test.go
```diff
@@ -60,13 +60,29 @@ func TestSchemaParser(t *testing.T) {
...
+			"value2",
+		},
+		"x-property-array-obj": []interface{}{
+			map[string]interface{}{
+				"name":  "obj",
+				"value": "field",
+			},
+		},
+		"x-property-value": "value",
+		"x-go-name":"Name",
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

PR #1895, opened 2019-03-05
Update authorizers.go

najeal:
Example modification:
Check all scopes are in role claims

### DIFF
--- a/examples/composed-auth/auth/authorizers.go
+++ b/examples/composed-auth/auth/authorizers.go
@@ -95,12 +95,17 @@ func HasRole(token string, scopes []string) (*models.Principal, error) {
 			isInScopes := false
 			claimedRoles := []string{}
 			for _, scope := range scopes {
+				isInScopes = false
 				for _, role := range claims.Roles {
 					if role == scope {
 						isInScopes = true
-						// we enrich the principal with all claimed roles within scope (hence: not breaking here)
+						// we enrich the principal with all claimed roles within scope
 						claimedRoles = append(claimedRoles, role)
+						break
 					}
+				}
+				if isInScopes == false {
+					break
 				}
 			}
 			if isInScopes {


examples/composed-auth/auth/authorizers.go
```diff
@@ -95,13 +95,18 @@ func HasRole(token string, scopes []string) (*models.Principal, error) {
 			isInScopes := false
 			claimedRoles := []string{}
 			for _, scope := range scopes {
+				isInScopes = false
```
fredbi: What motivates your change here?
The initial intent of this demo was precisely to show how to merge claims from several scopes.
If you break on first match, this defeats the purpose...

examples/composed-auth/auth/authorizers.go
```diff
@@ -95,13 +95,18 @@ func HasRole(token string, scopes []string) (*models.Principal, error) {
...
 						isInScopes = true
-						// we enrich the principal with all claimed roles within scope (hence: not breaking here)
+						// we enrich the principal with all claimed roles within scope
 						claimedRoles = append(claimedRoles, role)
+						break
 					}
 				}
+				if isInScopes == false {
+					break
+				}
```
fredbi: It looks like you want to achieve something else than the what I wanted to demonstrate here.
It might be interesting, although I can't clearly see what is the expected behavior here.
A good way to illustrate how to play with this differently could be to contribute a full example with the specifics of your use case explained.

examples/composed-auth/auth/authorizers.go
najeal: I wanted to use the HasRole function but the function succeeds when the scope are [foo, bar] and the user just has [foo] role. Then I propose changes for the next people which copy/paste the function.
For each scope iteration it starts for failing if we don't find the same role claimed in the role loop. If the role is found, we take the next scope and start by failing it until we find the same role claimed, etc
Hope my explanation is clear

examples/composed-auth/auth/authorizers.go
najeal: As I said above, I break the scope loop if the scope has not been found in the roles claimed.
The inScope will be false and the HasRole function will fail.
My purpose in the modification is to force all scopes to be present in the jet roles claimed to succeed the function.

fredbi: The intent documented with this function was "// We verify that the claimed role is **one of** the passed scopes".
I understand that you want a **all of** check. 
If this is the case, please add a function with this semantics and propose an additional path in the example that demonstrates this feature.

najeal: Ok, I understand. Thanks you

PR #1900, opened 2019-03-11
Fix some typos

knweiss:
Note: I did not recreate all the generated files after changing the server templates.


gregmarr:
The fixes themselves look good.  Just need to do the updates of all the generated stuff that goes along with it.

knweiss:
@gregmarr Okay, I gave it a shot. Please verify.

However, I've noticed two issues:
1. I did not update `examples/contributed-templates/stratoscale/` as the `go generate` call in `examples/contributed-templates/stratoscale/restapi/configure_petstore.go` is missing.
2. The `go generate` call in `examples/todo-list/restapi/configure_todo_list.go` uses the default flag-strategy but the existing generated code uses the optional `--flag-strategy=pflag`. To keep the diff as small as possible I've kept using this flag strategy but did not edit/change the (generated) `go generate` call.

fredbi:
> I did not update examples/contributed-templates/stratoscale/ as the go generate call in examples/contributed-templates/stratoscale/restapi/configure_petstore.go is missing.

This is intended, the stratoscale contrib template set generates itself the configure file.


Overall I think it is good to go.

PR #1902, opened 2019-03-14
WIP: switched to go mod

onokonem:


cmd/swagger/commands/generate/spec.go
```diff
@@ -26,6 +26,9 @@ import (
 	"github.com/go-swagger/go-swagger/scan"
 	"github.com/jessevdk/go-flags"
 	"gopkg.in/yaml.v2"
+
+	_ "github.com/davecgh/go-spew/spew"
```
golangcibot: a blank import should be only in a main or test package, or have a comment justifying it (from `golint`)

fredbi:
I think we have switched master to go modules now. @onokonem does this one add something different? Or could it be just closed?

onokonem:
Please close

PR #1903, opened 2019-03-17
development: strict types could be returned from handlers

onokonem:
to make code more safe


generator/shared.go
```diff
@@ -1257,3 +1268,10 @@ func sharedValidationsFromSchema(v spec.Schema, isRequired bool) (sh sharedValid
 	}
 	return
 }
+
+func stringTernary(cond bool, ifTrue string, ifFalse string) string {
```
golangcibot: `stringTernary` - `ifTrue` always receives `"Strict"` (from `unparam`)

casualjim:
Do you have an example of what the end result looks like? Maybe in the examples folder?

onokonem:
of course I do :)


[configure_issue_tracker.go](https://github.com/go-swagger/go-swagger/files/2975515/configure_issue_tracker.go.txt)

this is a file generated from `examples/task-tracker`. there are  some more changes but they all are related to the changes in this file.

The main idea is to force the handler to return one of the replays predefined for this particular method.

onokonem:
and yes, I can create a strict (and build api) examples based on the existing ones. just tell me which one suits best

casualjim:
Either the todo list api or the issue tracker example would be great

onokonem:
done

onokonem:
BTW, I need help adding tests for these options.

casualjim:
so essentially the change is it returns the more precise responder types instead of the interface?
And the builder changes because of that, including the not implemented error?

onokonem:
exactly

casualjim:
for tests we typically take a spec, generate code and verify that the files have the expected content through regex matchers etc.

casualjim:
I suppose this is not ready for merge yet is that correct?

onokonem:
from my point of view it is ready to be merged

it is in use in prod in some place at the moment

the parts are missing:

- tests to satisfy the coverage checker
- tests to actual test the functionality

unfortunately it looks like I'll not complete the tests

casualjim:
Unfortunately we need tests to accept this contribution, so that it doesn't inadvertently breaks. The last year we spent increasing our test coverage and stability.

onokonem:
> Unfortunately we need tests to accept this contribution, so that it doesn't inadvertently breaks. The last year we spent increasing our test coverage and stability.

do you have your test system documented somewhere?

I might be able to assign somebody less busy/lazy to this task

casualjim:
I don't think we have it documented somewhere but it's pretty straight forward. You make a spec or take one of the pre-existing ones, and then call render. We have tons of tests that all look like this:

https://github.com/go-swagger/go-swagger/blob/master/generator/discriminators_test.go#L38-L65

Or more in line with this changeset:
https://github.com/go-swagger/go-swagger/blob/master/generator/server_test.go#L347-L379

fredbi:
@onokonem I've had a look at this work and was surprised about the way you are duplicating a lot of templates.
Is this change so hard on response, etc templates so as to require an entire copy of this source?

Couldn't the same be achieved with proper {{ if .Stricts }} ... {{ else }} ... {{ end }} with a single template?

onokonem:
> Couldn't the same be achieved with proper {{ if .Stricts }} ... {{ else }} ... {{ end }} with a single template?

@fredbi , of course it could

But these templates are overcomplicated already.

fredbi:
> But these templates are overcomplicated already.

So you think that duplicating the source would make the overall solution simpler to maintain and safer to use? I don't think anyone is going to apply their changes everywhere and maintain both sets of templates.

Another way to "fork out" templates is to add "contrib" templates, which live their own lives next to the main templates set.

onokonem:
> Another way to "fork out" templates is to add "contrib" templates, which live their own lives next to the main templates set

I've tried that first, but found I need one more parameter passed to templates.

So I've added them

For my own task I've added the templates we are using to the project.

And yes, I'm ready to discuss how this has to be refactored to be added to the upstream, as it is a little pain to maintain the fork.

generator/shared.go
```diff
@@ -580,6 +588,9 @@ type GenOpts struct {
 	CompatibilityMode      string
 	ExistingModels         string
 	Copyright              string
+
+	Strict          bool
```
fredbi: GenOpts are available from the template, so there is no need to duplicate templates

generator/shared.go
```diff
@@ -1257,3 +1268,10 @@ func sharedValidationsFromSchema(v spec.Schema, isRequired bool) (sh sharedValid
 	}
 	return
 }
+
+func stringTernary(cond bool, ifTrue string, ifFalse string) string { // nolint: unparam
```
fredbi: the // nolint: hint is in my opinion unnecessary. You could just declare this has: (cond bool, ifTrue, ifFalse string) string

onokonem:
great, will do on weekend

onokonem:
replaced by https://github.com/go-swagger/go-swagger/pull/2312

Issue #125, opened 2015-11-22
Keep object's parameters order as specified in swagger spec definition


aleksandr-vin:
Consider a `swagger` option to produce sorted-order fields in models' structs or a spec-persistent order.

As for me: I place parameters in spec in a specific order to describe the object in domains. But `swagger generate` sorts them alphabetically, and such order is taken for ORM system to generate tables in DB... which are rather hard to read.

casualjim:
I agree that this is a good thing to have, at least a way to order that's not alphabetical.
But for now the implementation uses a map and there is no weight option defined yet, so alphabetical is the only way to get a stable sort (and stable diffs across generations)

AlekSi:
#1939 was merged. Can this one be closed now?
When you plan to make a new release with this feature?

PR #1939, opened 2019-05-09
support keep spec properties order

efi-tectonic:
This PR solves issue #125 (Keep object's parameters order as specified in swagger spec definition).

Before loading the spec , if argument--keep-spec-order exists , it will inject x-order attribute to each property which is equal to its position in the spec file.

---
This change is [Reviewable](https://reviewable.io/reviews/go-swagger/go-swagger/1939)


casualjim:
would it be possible to add a test demonstrating this behavior?

efi-tectonic:
> would it be possible to add a test demonstrating this behavior?

1. Where should I add the test?
2. Do you what to assert the generated go file ?

casualjim:
This is used in the generation process right?
you can add the test in ./generator/model_test.go for example
I think in this case you can provide a go file with the expected result and see that the output of the render matches.

efiko12:
Am I missing tests for codecov checks ?

casualjim:
Codecov is always unhappy, it's not a required check

BrendanMartin:
Can this work for `generate spec` as well?

terranvigil:
> Can this work for `generate spec` as well?

It doesn't appear to but would also like to see this happen.

PR #1949, opened 2019-05-23
urlbuilder template fixed to support query params named 'append'

### LINKED ISSUE #1948: Query parameter can't be named 'append' when enum appearing also
## Problem statement

When a query parameter is named named 'append', AND there is an enumeration in a different query parameter, a portion of the generated code won't build because a variable named 'append' is created.

## Swagger specification

```
{
  "swagger": "2.0",
  "info": {
    "description": "Simple API",
    "version": "2.0.0",
    "title": "Simple API"
  },
  "tags": [
    {
      "name": "general",
      "description": "General settings and information"
    }
  ],
  "schemes": [
    "http",
    "https"
  ],
  "paths": {
    "/api/stuff": {
      "get": {
        "tags": [
          "general"
        ],
        "summary": "Retrieves stuff",
        "operationId": "getStuff",
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "append",
            "in": "query",
            "required": false,
            "default": false,
            "type": "boolean"
          },
          {
            "name": "results",
            "in": "query",
            "required": false,
            "type": "array",
            "minItems": 1,
            "maxItems": 5,
            "uniqueItems": true,
            "items": {
              "type": "string",
              "enum": [
                "Counts",
                "IDs",
                "Amounts",
                "Indexes",
                "rawIndexes"
              ]
            }
          }
        ],
        "responses": {
          "200": {
            "description": "Success, stuff returned"
          }
        }
      }
    }
  }
}
```
## Steps to reproduce

```
swagger generate server -f sample.json
```

```
go build cmd/simple-server/main.go
# myproj/restapi/operations/general
restapi/operations/general/get_stuff_urlbuilder.go:64:22: cannot call non-function append (type string)
```

The code being produced by the generator produces the following unbuildable code:

from generated file: get_version_urlbuilder.go, the line in bold is line 64.
<pre>
// Build a url path and query string
func (o *GetStuffURL) Build() (*url.URL, error) {
        var _result url.URL

        var _path = "/api/stuff"

        _basePath := o._basePath
        _result.Path = golangswaggerpaths.Join(_basePath, _path)

        qs := make(url.Values)

        var append string
        if o.Append != nil {
                append = swag.FormatBool(*o.Append)
        }
        if append != "" {
                qs.Set("append", append)
        }

        var resultsIR []string
        for _, resultsI := range o.Results {
                resultsIS := resultsI
                if resultsIS != "" {
                        <b>resultsIR = append(resultsIR, resultsIS)</b>
                }
        }

        results := swag.JoinByFormat(resultsIR, "")

        if len(results) > 0 {
                qsv := results[0]
                if qsv != "" {
                        qs.Set("results", qsv)
                }
        }

        _result.RawQuery = qs.Encode()

        return &_result, nil
}
</pre>

## Environment
swagger version: v0.19.0 
go version: go1.12.1 darwin/amd64 
OS:   MacOS

jimturnquist:
fixes #1948

Update urlbuilder_test.go to support added query name mangling.

Signed-off-by: Jim Turnquist <jim@boonlogic.com>


fredbi:
Yes there are many codegen assertions in CI. So you have to scan tests in generator to find out when urlbuilder template is tested.

jimturnquist:
Do I need to generate the bindata.go before putting up the pull request.  I was under the impression that git hooks would take care of that.  Is that correct?

fredbi:
yes they normally do, but it seems that in this case it did not work. We use bindata from Kevin Burke's fork.

jimturnquist:
Any thoughts on where I should go from here.  I believe that coverage decrease was due to the bindata.go regeneration.

Actually, are we expecting the number of changes in bindata.go based on the change I made to the templates?  I used Kevin Burkes go-bindata (master) as recommended.

fredbi:
Codecov is not a required check so don't worry: it is obvious the bindata is causing this. Looks like we haven't regenerated for quite a while...

PR #1962, opened 2019-06-05
do not force set the api logger to log.Printf

djgilcrease:
the logger already defaults so do not force set it in `SetAPI` since that would overwrite one I set in the API before calling `SetAPI`


generator/templates/server/server.gotmpl
```diff
@@ -256,7 +256,6 @@ func (s *Server) SetAPI(api *{{ .Package }}.{{ pascalize .Name }}API) {
 	}
 
 	s.api = api
-	s.api.Logger = log.Printf
```
fredbi: You should regenerate bindata.go (this is done automatically if you enable .githooks on this repo).
Note that bindata is regenerated using go-bindata fork from Kevin's Burke

djgilcrease: done

fredbi:
@djgilcrease  This one just needs an update with regenerated bindata

PR #1969, opened 2019-06-16
Fix basic auth for stratoscale template

### LINKED ISSUE #1765: Basic auth in stratoscale templates
## Problem statement

The code, generated for basic auth schema out of stratoscale templates, is not compilable. The generated code (configure_user_service.go) out of the fragment below is:

```
// Config is configuration for Handler
type Config struct {
...
	// AuthUserAuth for basic authentication
	AuthUserAuth func(user string, pass string)
...
}


// Handler returns an http.Handler given the handler configuration
// It mounts all the business logic implementers in the right routing.
func Handler(c Config) (http.Handler, error) {
...
	api.UserAuthAuth = func(user string, pass string) (interface{}, error) {
		if c.AuthUserAuth == nil {
			return "", nil
		}
		return c.AuthUserAuth(user, pass)
	}
...
}
```

It is not compilable, because the signature of `AuthUserAuth `does not correspond to `func(user string, pass string) (interface{}, error)` with respect to the return types.
The fix seems to be trivial, change the definition `AuthUserAuth func(user string, pass string)` to `AuthUserAuth func(user string, pass string) (interface{}, error)`, which in terms of templates means changing 
```
	{{ if .IsBasicAuth -}}
	// Auth{{ pascalize .ID }} for basic authentication
	Auth{{ pascalize .ID }} func(user string, pass string)
	{{ end -}}

```
to
```
	{{ if .IsBasicAuth -}}
	// Auth{{ pascalize .ID }} for basic authentication
	Auth{{ pascalize .ID }} func(user string, pass string) (interface{}, error)
	{{ end -}}

```
## Swagger specification

(fragment)

```
  /token:
    post:
      tags:
      - Token
      summary: Issues a token for an authenticated user
      description: >-
        Issues a token for an authenticated user
      operationId: issueToken
      produces:
      - application/json
      responses:
        200:
          description: OK
          schema:
            $ref: "./AccessToken.yaml"
        400:
          $ref: "./responses.yaml#/400"
        401:
          $ref: "./responses.yaml#/401"
        403:
          $ref: "./responses.yaml#/403"
        404:
          $ref: "./responses.yaml#/404"
        412:
          $ref: "./responses.yaml#/412"
      security:
      - userAuth: []

securityDefinitions:
  userAuth:
    type: basic
```

## Steps to reproduce

`swagger generate server...`

## Environment
swagger version: 0.17.0  
go version: 1.11  
OS:  Win 10 (Linux subsystem / Ubuntu 18.04)

JoakimSoderberg:
Use the correct signature for the basic authentication interface.
This caused the compilation to fail. Fixes #1765


casualjim:
can you run `go generate ./generator` so that the template changes get embedded in the binary?

PR #1978, opened 2019-06-21
Generated client: handles unexpected successful responses

### LINKED ISSUE #1518: panic: interface conversion: interface {} is *mypackage.PostActionsDefault, not *mypackage.PostActionsCreated
## Problem statement

I have a spec file which has a default response to catch unexpected codes from the server
```
      responses:
        201:
          description: ok
          schema: {$ref: '#/definitions/action'}
        400:
          description: error
          schema: {$ref: '#/definitions/error'}
        default:
          description: unexpectederror
          schema: {type: text}
```

when the server does respond with either a non 201, 2XX reply like 204, the generated client panics because the code flows through the success case and tries to map it to the action object.

## Swagger specification
2.0

## Steps to reproduce
use default response and only declare 201 as expected output. But make server misbehave and send a 204 instead.

## Environment
swagger version: 
version: 0.13.0
commit: 8135eb6728e43b73489e80f94426e6d387809502 
go version: 
go version go1.9.2 darwin/amd64

OS:   
OSX

fredbi:
Generated client: handles unexpected successful responses
and returns default response with error rather than panicking

Fixes #1518


casualjim:
This one needs an update before it can proceed

PR #1981, opened 2019-06-22
Client Template Interface

rokf:
Added an interface for client methods. With an interface clients can be mocked for tests.


casualjim:
Can you run `go generate ./generator`
Otherwise the template changes won't be picked up by the build

rokf:
Looks like those are not the only files that have to be updated since some tests are failing :smile:

fredbi:
Superseded by #2151 (commits from the original author have been kept)

PR #1982, opened 2019-06-23
fixed push latest tag on releases

### LINKED ISSUE #1873: Why is docker image "latest" NOT aligned with the latest tag?
Hi, 
As for now, on https://quay.io/repository/goswagger/swagger?tab=tags docker image "latest" is aligned with v0.17.0 even though there are newer tags (v0.18.0 for example). Why is it not aligned with latest tag?
Thanks,
Dror

fredbi:
Fixes #1873 (for next release)

@casualjim do you have the possibility to manually tag v0.0.19 as latest on quay.io?
 
### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,7 @@ jobs:
               upload_to_github
               upload_to_bintray
 
+              export EXTRA_TAG=latest
               deploy_docker
             fi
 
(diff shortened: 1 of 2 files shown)


casualjim:
This PR doesn't fix the issue.
The issue is that the regular build overrides the latest tag, so that needs to be removed from the circle ci build.
I won't do it for the previous release but we can do it for the next one

fredbi:
ok got it I'll give it another try

fredbi:
@casualjim hope now this corrects the image tagging issue

PR #1984, opened 2019-06-24
Honour JSON ,string directive of ptr field when generating spec

wych42:
This is an update of pull request: #1511 

For example:

```
// swagger:model Foo
type Foo struct {
    Bar *int64 `json:"bar,omitempty,string"`
}
```

The generated spec should be type "string" and format "int64".


scan/schema.go
```diff
@@ -1323,3 +1313,22 @@ func parseJSONTag(field *ast.Field) (name string, ignore bool, isString bool, er
...
+// isFieldStringable check if the field type is a scalar. If the field type is
+// *ast.StarExpr and is pointer type, check if it refers to a scalar.
+// Otherwise, the ",string" directive doesn't apply.
+func isFieldStringable(tpe ast.Expr) bool {
+	if ident, ok := tpe.(*ast.Ident); ok {
+		switch ident.Name {
+		case "int", "int8", "int16", "int32", "int64",
+			"uint", "uint8", "uint16", "uint32", "uint64",
+			"float64", "string", "bool":
+				return true
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			return true
```

Issue #1724, opened 2018-09-23
Support go modules


casualjim:
Supporting go modules will require a number of changes. Some are mechanical and can be scripted, others mean changing the way our code works.  

## Tagging changes

The first one is fairly easy. On the next release we tag versions with the *v* prefix. And we add a go.mod and go.sum file in every package under the go-openapi organization as well as in this go-swagger/go-swagger repository.

* [x] **Done**

## Turn on go vet during tests

Currently our generator package is doing a cute thing in the debugLog and debugLogAsJSON methods.
go vet is unhappy about it, so we've turned it off until we can address the format string issues.

https://github.com/search?l=&q=debugLog+repo%3Ago-swagger%2Fgo-swagger+extension%3Ago+path%3Agenerator&type=Code

* [x] **Done**

## Package refactoring

For our current version we don't need to rewrite packages or import paths, it's only if/once we decide to do v2 then we need to use different import paths. 

## Generated code changes

The generator should optionally add go.mod or edit the go.mod file with the required dependencies. Although the tooling should pick up on that too the first time the code gets run. But if people want to pin to older versions etc then this can help them out with that.

Initial work has done some of this: https://github.com/search?q=%22go.mod%22+repo%3Ago-swagger%2Fgo-swagger+extension%3Ago+path%3Agenerator&type=Code

## Generator support

We do lookups in the go path for guessing imports and so on. These will need to be made compatible with go modules. Currently we basically strip off or join with `${GOPATH-"$HOME/go"}/src` 

We would need to change this logic to be aware of modules and look up with that tool.

https://github.com/search?utf8=%E2%9C%93&q=GOPATH+repo%3Ago-swagger%2Fgo-swagger+extension%3Ago+extension%3Ago+path%3Agenerator%2F&type=Code&ref=advsearch&l=&l=

## Scanner support

The scanner uses `go/build` from the go compiler infrastructure, to discover the files and packages that need to be introspected.
This needs to move to the newer library: https://godoc.org/golang.org/x/tools/go/packages so that it's module aware.

https://github.com/search?utf8=%E2%9C%93&q=GOPATH+repo%3Ago-swagger%2Fgo-swagger+extension%3Ago+path%3Ascan&type=Code&ref=advsearch&l=&l=

related #1677

fredbi:
RE:

> Turn on go vet during tests

also affects generated code, see also #1573. This was actually introduced by go1.10.

[...]

PR #1995, opened 2019-07-04
Set govet again

fredbi:
**Housekeeping**

* Enabled go vet check again in hack testing scripts and in appveyor tests.
* Regen examples from latest master (brings some minor differences in embedded spec codegen, stems from updated go-openapi/spec dependency)

Contributes #1724 


fredbi:
Added some guard in new scanner test suite: some unstable behavior did cause tests to panic when it hits the problem. This is still brittle, but it does not panick...

PR #1998, opened 2019-07-04
Codegen edges cases

### LINKED ISSUE #1937: Swagger generates and uses different models in case of definition name starts with Digit
In swagger.yml file my definitions named like "400Schema", "404Schema" and etc. When I generate source from swagger, it was created models in the form "NR404Schema" but restapi code was used it as "X404Schema". Hence restapi could not find proper model objects. 
It is happening only in case when definitons' name starts with digit.

fredbi:
updated from latest tag of go-openapi/swag

Handling of names not starting with a letter (mismatches between GoType and model name)

This makes the pascalize and swag.ToGoName funcs work with similar rules.

* fixes #1937

### DIFF
--- a/fixtures/bugs/1937/.gitignore
+++ b/fixtures/bugs/1937/.gitignore
@@ -1,0 +1,3 @@
+gen-*
+!gen-fixtures.sh
+*.log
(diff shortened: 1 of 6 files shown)


fredbi:
ah this one hit the scanner random issue again... Can't rebuild CI job, though

PR #1999, opened 2019-07-04
updated dependencies

fredbi:
This results from taggging runtime (v0.19.3) and carrying out a `go get -u $(go list) ./...`  on this repo.

This builds therefore embeds all recents changes from the go-openapi libs, as of 2019/07/04.
No change affected the generation of examples (tried it out).


fredbi:
Codecov yields false negative...

PR #2004, opened 2019-07-07
refactor circleci build

casualjim:
* creates a fast pass that just runs the tests without race and builds a binary
* runs the following things in parallel
  * tests with race detection and coverage reporting
  * codegen-reg tests
  * canary tests
* builds and publishes a docker container on merges to master, tagged as dev
* when the build is for a tag (`git push --follow-tags`):
  * build a docker container tagged with the version and latest 
  * build cross-platform binaries and packages and publish to github and bintray


hack/deploy.sh
```diff
@@ -71,7 +71,7 @@ deploy_docker() {
   docker login -u $API_USERNAME -p $QUAY_PASS https://quay.io
   docker push quay.io/goswagger/swagger:$CIRCLE_TAG
 
-  if [[ -n "${EXTRA_TAG}" ]] ; then
+  if [[ -n "${EXTRA_TAG-""}" ]] ; then
```
fredbi: Why this script is not removed altogether? It seems no to be used any longer with this new CI.

PR #2017, opened 2019-07-11
2016 - Allow dynamic TLS configurations in servers

### LINKED ISSUE #2016: Cannot use dynamic certificate selection on servers
## Problem statement
The generated server code forces a static TLS configuration, which is not suitable for all use cases.

## Swagger specification
N/A

## Steps to reproduce
In `configureTLS`, set `tlsConfig.Certificates = nil` and `tlsConfig.GetCertificates` to any function that returns a valid certificate.

## Environment
swagger version: 0.18.0 
go version: 1.11.2  
OS: Ubuntu 18.04

Notes: About to submit PR to fix.

mmussomele:
The generated server code only checks the length of the https server's
tls config certificate list, forcing static TLS configurations. But, it
should be valid to set the tls config's GetCertificate function to allow
dynamic certificate selection.

Fixes #2016


mmussomele:
Relevant Go core PR: https://go-review.googlesource.com/c/go/+/8792/ (submitted in 2015 to allow this behavior by regardless of client SNI)

casualjim:
Can you run `go generate ./generator` so that the template changes are picked up in the binary?

mmussomele:
Updated using https://github.com/kevinburke/go-bindata, was that the correct one? It seems the original is missing

PR #2019, opened 2019-07-18
Add method `Payload()` to error types

### LINKED ISSUE #2018: Add `Payload() *models.TheError` to error types
## Problem statement

Client code generated by go-swagger includes structs for transporting error responses like 404, 409, etc:

```go
// NewCreateOAuth2ClientBadRequest creates a CreateOAuth2ClientBadRequest with default headers values
func NewCreateOAuth2ClientBadRequest() *CreateOAuth2ClientBadRequest {
	return &CreateOAuth2ClientBadRequest{}
}

/*CreateOAuth2ClientBadRequest handles this case with default header values.

genericError
*/
type CreateOAuth2ClientBadRequest struct {
	Payload *models.GenericError
}

func (o *CreateOAuth2ClientBadRequest) Error() string {
	return fmt.Sprintf("[POST /clients][%d] createOAuth2ClientBadRequest  %+v", 400, o.Payload)
}

func (o *CreateOAuth2ClientBadRequest) readResponse(response runtime.ClientResponse, consumer runtime.Consumer, formats strfmt.Registry) error {

	o.Payload = new(models.GenericError)

	// response payload
	if err := consumer.Consume(response.Body(), o.Payload); err != nil && err != io.EOF {
		return err
	}

	return nil
}
```

For each error code, a new struct is being generated. This makes getting the payload for these errors very cumbersome, as one has to do multiple type assertions.

Instead, it would be super useful if these errors expose a method `Payload` that simply returns `o.Payload`:

```go
func (o *CreateOAuth2ClientBadRequest) Payload() *ThePayloadType {
	return o.Payload
}
```

That way, we could simply create an interface and type assert against that, if the inner payload is required!

## Swagger specification

Not relevant.

## Steps to reproduce

Generate a client using any swagger definitions that has HTTP Error Response Codes.

## Environment
swagger version: 0.19.0  
go version: 11.12.5
OS: irrelevant

aeneasr:
Adding method `Payload() *TheError` allows for developers to type assert against an interface, making it possible to access the payload without needing multiple type assertions.

Closes #2018


generator/bindata.go
```diff
@@ -1,44 +1,42 @@
-// Code generated by go-bindata. DO NOT EDIT.
+// Package generator Code generated by go-bindata. (@generated) DO NOT EDIT.
```
fredbi: Could you update your bindata generator? we use Burke's fork

generator/templates/client/response.gotmpl
```diff
@@ -35,6 +35,11 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}) Error() string {
 	return fmt.Sprintf("[{{ upper .Method }} {{ .Path }}][%d] {{ if .Name }}{{ .Name }} {{ else }}unknown error {{ end }}{{ if .Schema }} %+v{{ end }}", {{ if eq .Code -1 }}{{ .ReceiverName }}._statusCode{{ else }}{{ .Code }}{{ end }}{{ if .Schema }}, o.Payload{{ end }})
 }
 
+{{ if .Schema }}
+func ({{ .ReceiverName }} *{{ pascalize .Name }}) GetPayload() {{ if and (not .Schema.IsBaseType) (not .Schema.IsInterface) .Schema.IsComplexObject (not .Schema.IsStream) }}*{{ end }}{{ if (not .Schema.IsStream) }}{{ .Schema.GoType }}{{ else }}io.Writer{{end}} {
```
fredbi: But is that signature not defeating the purpose of having a simple common interface since we can have different signature?
Why are we allowing the possibility to generate a wrong `*io.Writer` here?
Also wouldn't it be some `ReadCloser` instead?

aeneasr: It uses 1:1 the code from [here](https://github.com/go-swagger/go-swagger/pull/2019/files/e623aab4b143d7a2f1d29d66a0da51b58b1b95df#diff-6b20cc8a4514b56c8fc0488c02ece5f0R23)

generator/bindata.go
aeneasr: How do I achieve that? I ran `brew install go-bindata` but running `go generate ./generator` doesn't make a difference.

generator/templates/client/response.gotmpl
aeneasr: > But is that signature not defeating the purpose of having a simple common interface since we can have different signature?

Not really, most error messages share the schema and if they don't then they should have different signatures. It follows the same logic basic fetching is using.

Alternatively we could expose the response body but then we have to de/encode it again.

generator/bindata.go
fredbi: `go get -u github.com/kevinburke/go-bindata/go-bindata`

generator/templates/client/response.gotmpl
fredbi: ok I got your point. So we have many GetPayload() and these are going to be really useful only when the share the schema. So I am okay with that approach.
The io.Writer was probably wrong in the first place. This for the case when the response is `type: binary`

casualjim: The io.writer is not wrong, it's the sync when you consume a stream you want to write it  to somewhere

PR #2034, opened 2019-08-03
fix:walkImports should recursively traverse the imported package

d0ngw:


casualjim:
What is the problem this solves?
Can you add a test that demonstrates it's required?

d0ngw:
> What is the problem this solves?
> Can you add a test that demonstrates it's required?

[swagger-demo.zip](https://github.com/go-swagger/go-swagger/files/3464608/swagger-demo.zip)

The attachment is a test demo, please download and unzip it.
When I try to generate the swagger docment using go-swagger, it will fail and print the follow message: 

`can't find source file for struct: swagger-demo/model.AutoID`

This patch can fix the problem, please check it, thank you.

casualjim:
if you don't `cd controller` in your gendoc.sh script, it seems to work for me

```sh
unzip -d . swagger-demo.zip
cd swagger-demo
swagger generate spec
```
results in 
```json
{
  "swagger": "2.0",
  "paths": {
    "/hello": {
      "post": {
        "description": "say hello",
        "tags": [
          "User",
          ""
        ],
        "operationId": "helloParam",
        "responses": {
          "default": {
            "$ref": "#/responses/helloResponse"
          }
        }
      }
    }
  },
  "responses": {
    "helloResponse": {
      "schema": {
        "type": "object",
        "properties": {
          "ct": {
            "type": "integer",
            "format": "int64",
            "x-go-name": "CreateTime"
          },
          "id": {
            "type": "integer",
            "format": "int64",
            "x-go-name": "ID"
          },
          "name": {
            "type": "string",
            "x-go-name": "Name"
          },
          "ut": {
            "type": "integer",
            "format": "int64",
            "x-go-name": "UpdateTime"
          }
        }
      }
    }
  }
}
```

d0ngw:
Yes, if don't cd controller and it works for me.

But our project has different types of apis, one for internal and one for public api.
I want to generate a swagger document for two types of apis in a separate file, and I need to cd the target dir and run the command. 

That's why the  'gendoc.sh' need to cd dir.

Issue #2036, opened 2019-08-07
stratoscale client template breaks on multiline descriptions


avorima:
## Problem statement

The stratoscale template can't handle multiline strings as path entry descriptions.

## Swagger specification

2.0

## Steps to reproduce

Create a swagger spec file containing the following path:

```yaml
/test:
  get:
    description: |
      Some
      multi
      line
      string
    responses:
      '200':
        description: Ok
```

Run:

```
swagger generate client --template=stratoscale
```

This will result in `operations_client.go` containing the following snippet:

```go
...
type API interface {
    // GetTest Some
    multi
    line
    string
    ...
}
```

## Environment
swagger version: 0.18.0  
go version: 1.12.5  
OS: linux 5.1.18

PR #2037, opened 2019-08-07
Fix multi-line comments in stratoscale client template

avorima:
Fixes: #2036


casualjim:
can you run `go generate ./generator` so that the changes are picked up in the binary

avorima:
@casualjim I updated the generated code, PTAL

Issue #2035, opened 2019-08-07
Setting additionalProperties to false does not have any effect


connesc:
## Problem statement

I want to reject unknown fields provided in request body. According to [the FAQ](https://goswagger.io/faq/faq_model.html#extra-sections-in-post-body), it seems to be possible by simply adding `additionalProperties: false` to my definition.
However, this does not have any effect.

Is this a regression or just a misunderstanding?

## Swagger specification

```yaml
swagger: '2.0'
info:
  title: Example
  version: '0.1'
consumes:
- application/json
produces:
- application/json
paths:
  /test:
    post:
      operationId: test
      parameters:
      - name: body
        in: body
        schema:
          type: object
          properties:
            foo:
              type: boolean
          additionalProperties: false
      responses:
        200:
```

## Steps to reproduce

The following request should raise a 400 error, but does not:

```shell
curl -H 'Content-Type: application/json' -d '{"foo": true, "bar": false}' "http://127.0.0.1:$PORT/test"
```

## Environment
swagger version: v0.19.0
go version: go1.12.7 linux/amd64
OS: Linux

fredbi:
https://goswagger.io/use/models/schemas.html#known-limitations-with-go-swagger-models

connesc:
Thanks for the pointer, I didn't notice this doc before.

So at this point, I guess my best bet is to set `additionalProperties` to `true`, so that I can check that the `map[string]interface{}` is empty in my own code. However, it remains annoying because the Swagger doc will basically say the reverse of what it should.

Any idea for a better workaround?

fredbi:
No, no other idea.

As explained in the link above, the trade-off in our implementation is currently to ignore additional properties when set to false, so this does not bear any validation semantics.
This is the same for `additionalItems`.

That being said, I am not against some opt-in behavior, e.g. `--strict-additional-properties` that would enforce validation (i.e. unwanted extra properties make the message invalid). You can add this kind of additional check in the custom unmarshaller template for schemas with additionalProperties.

connesc:
@fredbi let me know if #2040 fits what you had in mind. At least, it perfectly solves my issue.
Thanks a lot for your help!

connesc:
For future references, two alternatives are available:
- using `--strict-additional-properties` as implemented in #2040 (currently unsure whether it will be merged)
- using a custom `runtime.Consumer` as pointed out by @casualjim [here](https://github.com/go-swagger/go-swagger/pull/2040#issuecomment-520485853)

PR #2040, opened 2019-08-12
Add support for --strict-additional-properties

connesc:
Implements what have been proposed in #2035.

When `--struct-additional-properties` is set, models with `additionalProperties: false` (which is the default) are given a custom `UnmarshalJSON` with [`Decoder.DisallowUnknownFields`](https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields) enabled.


casualjim:
To get this functionality you can also just register a different jsonconsumer with those settings. Might be easier: https://github.com/go-openapi/runtime/blob/master/json.go

You then register that for the mimetype and you get the behavior without any changes to the generated code

fredbi:
Looks good to me.

connesc:
Thanks for the advise @casualjim, it seems to work perfectly as well!

So, now there are two interesting options for the same goal, with subtle differences:
- Registering a custom `runtime.Consumer` does not have any performance overhead.
- Generating a custom `json.Unmarshaler` enforces the Swagger spec directly in the generated code.

Also, I'm not sure how both options interact with other generated custom unmarshalers (e.g. for `allOf` or polymorphic types).

@casualjim @fredbi I let you decide whether this should be merged or not. I will just rebase the PR and regenerate `bindata.go` to fix conflicts.

fredbi:
@connesc nice analysis and work. Indeed we have two solution to handle strict validation.

The runtime approach is more elegant when running the full fledged toolkit. I like just as much the model-only approach as it is opt-in and allows for stricter jsonschema compliance when dealing with generated models only.

casualjim:
I kind of want to revert this. Because it seems like an ad-hoc addition that duplicates functionality that already exists in a verbose way. At the same time the interplay with additional properties, additional items and tuples is not clear. it's also not clear how it behaves with discriminated types.

fredbi:
@casualjim additionalItems is not affected (the template name is misleading, and this has been so for quite a long time).

There might indeed be some limitations with this simple addons regarding complex structures (allOf, etc), but overall I think it is a good addition.

I think of the models feature as kind of standalone and not necessarily coupled with the use of the go-openapi runtime, so I am not seeing this as duplicate functionality. I would not revert this, but to the contrary, maybe someday extend it to support additionalItems etc.

PR #2046, opened 2019-08-17
Add deprecated property support for route

MetalRex101:


MetalRex101:
It looks like i'm a bit late.
https://github.com/go-swagger/go-swagger/pull/2045

casualjim:
you have better tests imo so there is still value in this one

PR #2048, opened 2019-08-17
update builder template to pass along oauth2 scheme name

### LINKED ISSUE #2041: Pass OAuth2 scheme name to handler and add it on the request context
## Problem statement

When setting a custom BearerAuthentication function on a generated API spec, the function is always passed `name` as `""` which isn't very useful…

```go
api.BearerAuthenticator = func(name string, auth security.ScopedTokenAuthentication) runtime.Authenticator {
  if name == "myscheme" … // it never will
```

This is because the generated code for `AuthenticatorsFor` is passing `scheme.Name` as the first parameter. But this cannot be provided on an oauth2 scheme as implemented by go-swagger.

The generated code looks like:
```go
func (o *TestAPI) AuthenticatorsFor(schemes map[string]spec.SecurityScheme) map[string]runtime.Authenticator {
    result := make(map[string]runtime.Authenticator)
    for name, scheme := range schemes {
        switch name {
            case "auth_jwt":
                result[name] = o.BearerAuthenticator(scheme.Name, o.AuthJwtAuth)
        }
    }
    return result
}
```

To fix it, either the scheme.Name needs setting somewhere, or else the generated code for AuthenticatorsFor needs to just pass `name` instead.

## Swagger specification

```yaml
paths:
  /admin/function:
    post:
      security:
        - auth_jwt: [ admin ]
securityDefinitions:
  auth_jwt:
    type: oauth2
    authorizationUrl: http://example.com/dummy
    tokenUrl: http://example.com/dummy
    flow: accessCode
    scopes:
      admin: Admin
    description: Auth JWT
```

Trying to add `name: auth_jwt` to this results in:
```
The swagger spec at "swagger.yml" is invalid against swagger specification 2.0. see errors :
- "securityDefinitions.auth_jwt" must validate one and only one schema (oneOf). Found none valid
- securityDefinitions.auth_jwt.name in body is a forbidden property
```

## Steps to reproduce

Generate swagger definitions with oauth2 security.

## Environment
swagger version: v0.19.0
go version: 1.12.7
OS:  --

casualjim:
address some gocritic warnings

fixes #2041 


generator/templates/server/builder.gotmpl
```diff
@@ -220,7 +220,7 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}API) AuthenticatorsFor(schemes map
         {{if .IsAPIKeyAuth}}result[name] = {{.ReceiverName}}.APIKeyAuthenticator(scheme.Name, scheme.In, {{ if not ( eq .Principal "interface{}" ) }}func(token string) (interface{}, error) {
           return {{ end }}{{.ReceiverName}}.{{ pascalize .ID }}Auth{{ if not ( eq .Principal "interface{}" ) }}(token)
         }{{ end }}){{end}}
-        {{if .IsOAuth2}}result[name] = {{.ReceiverName}}.BearerAuthenticator(scheme.Name, {{ if not ( eq .Principal "interface{}" ) }}func(token string, scopes []string) (interface{}, error) {
+        {{if .IsOAuth2}}result[name] = {{.ReceiverName}}.BearerAuthenticator(name, {{ if not ( eq .Principal "interface{}" ) }}func(token string, scopes []string) (interface{}, error) {
```
fredbi: Whenever a spec only specifies a BearerAuthenticator scheme, then `scheme` is not used and the generated code fails to compile

fredbi: Here is my suggestion:
````golang
for name, scheme := range schemes {
    switch name {
      {{range .SecurityDefinitions}}
      case "{{.ID}}":
        {{if .IsBasicAuth}}_ = scheme
        result[name] = {{.ReceiverName}}.BasicAuthenticator({{ if not ( eq .Principal "interface{}" ) }}func(username, password string) (interface{}, error) {
          return {{ end }}{{.ReceiverName}}.{{ pascalize .ID }}Auth{{ if not ( eq .Principal "interface{}" ) }}(username, password)
        }{{ end }}){{end}}
        {{if .IsAPIKeyAuth}}result[name] = {{.ReceiverName}}.APIKeyAuthenticator(scheme.Name, scheme.In, {{ if not ( eq .Principal "interface{}" ) }}func(token string) (interface{}, error) {
          return {{ end }}{{.ReceiverName}}.{{ pascalize .ID }}Auth{{ if not ( eq .Principal "interface{}" ) }}(token)
        }{{ end }}){{end}}
        {{if .IsOAuth2}}_ = scheme
        result[name] = {{.ReceiverName}}.BearerAuthenticator(name, {{ if not ( eq .Principal "interface{}" ) }}func(token string, scopes []string) (interface{}, error) {
          return {{ end }}{{.ReceiverName}}.{{ pascalize .ID }}Auth{{ if not ( eq .Principal "interface{}" ) }}(token, scopes)
        }{{ end }}){{end}}
      {{end}}
    }
  }
```

fredbi: Or perhaps cleaner:
```golang
for name := range schemes {
    switch name {
      {{range .SecurityDefinitions}}
      case "{{.ID}}":
        {{if .IsBasicAuth}}
        result[name] = {{.ReceiverName}}.BasicAuthenticator({{ if not ( eq .Principal "interface{}" ) }}func(username, password string) (interface{}, error) {
          return {{ end }}{{.ReceiverName}}.{{ pascalize .ID }}Auth{{ if not ( eq .Principal "interface{}" ) }}(username, password)
        }{{ end }}){{end}}
        {{if .IsAPIKeyAuth}}scheme := schemes[name]
        result[name] = {{.ReceiverName}}.APIKeyAuthenticator(scheme.Name, scheme.In, {{ if not ( eq .Principal "interface{}" ) }}func(token string) (interface{}, error) {
          return {{ end }}{{.ReceiverName}}.{{ pascalize .ID }}Auth{{ if not ( eq .Principal "interface{}" ) }}(token)
        }{{ end }}){{end}}
        {{if .IsOAuth2}}
        result[name] = {{.ReceiverName}}.BearerAuthenticator(name, {{ if not ( eq .Principal "interface{}" ) }}func(token string, scopes []string) (interface{}, error) {
          return {{ end }}{{.ReceiverName}}.{{ pascalize .ID }}Auth{{ if not ( eq .Principal "interface{}" ) }}(token, scopes)
        }{{ end }}){{end}}
      {{end}}
    }
  }
```

casualjim:
I've addressed your comment

PR #2067, opened 2019-09-25
Add default flag support  `import "flags"`

hisunwei:
Hi  friends,   

For some modules in my golang project are using  `import "flag"`  as flags strategy,   but it was not compatibility with `go-flags|pflag`.
So, I add some code to support `import "flag"`  to make it work.
The usage like :

```
--flag-strategy=[go-flags|pflag|flag]    the strategy to provide flags for the server (default: go-flags)
```
hope this is useful to you and be adopted, thanks.


casualjim:
Thanks, can you run `go generate ./generator` so that the template changes are picked up in the binary

hisunwei:
> Thanks, can you run `go generate ./generator` so that the template changes are picked up in the binary

yeah,  binddata.go had been generated at modTime= 1568810519.  
And I tried the command again, there is no more changes need to commit.

PR #2076, opened 2019-10-06
added contrib to the docker image

ronnie-stratoscale:
This will allow using the --template-dir parameter when using the
go-swagger container


ronnie-stratoscale:
Thanks @casualjim. When will you release a new minor version?
Specifically how can I get a docker image with this change?

Issue #1606, opened 2018-07-06
Private templates


akhobov:
## Problem statement

There is a possibility to provide custom templates to the go-swagger tool to customise generated code. This functionality can be useful to patch/workaround some issues in the generated code too. However not all templates can be overloaded as majority of them marked as private.

Do you think it would be good to lift that limitation so all templates can be customised? This will give API developers a possibility to workaround some issues while waiting for a new release that fix the original issues. Obviously they would evaluate the risks and remove that customisation during the upgrade to the next version.

fredbi:
you can lift the constraint yourself by removing the "locked" attribute in `template_repo.go`

akhobov:
Yes, I can. However that would require a custom build that is not a very handy options for a group of developers. That's the reason we are sticking to released versions rather than master - to ensure a homogeneous development environment.

I'm not insisting on that change. Just asking if that seems a good idea from your point of view.

fredbi:
Some templates are locked because they are deeply intertwined with the rest of the generation code.
So customization here equates to a major fork.
A better strategy in the case of bugs in this case is to stick on master rather than releases and participate the contribution effort rather than waiting for a release.

fredbi:
CI testing on master have been largely extended and there is no real quality benefit to expect from releases besides repo tagging.

akhobov:
That's not the quality I'm talking about. Imaging there is a new dev joins your team that needs to install the tool or one of the developers decided to update the tooling while the rest continue to use their old versions. The generated code would be different depending on who run the generator. With tags there is some level of certainty that all devs are using the same version and as such they all can get the same result out of the go-swagger generator.

fredbi:
I am just like you guys.
I have a specific branch on my fork to stick my other devs.
This branch is regularly sync'ed with master, or sometimes completely reworked when I need something special ahead of master (or against @casualjim's wishes... 🤣 ).

I m using `dep` and `source` in `Gopkg.toml` to control how dependent packages depend on all this.

Another strategy I put in place for some features (for instance, I needed a custom swagger serve command) is to bake my own docker image from a forked branch and distribute this independently.

akhobov:
Got you. Thanks a lot!

Anyway, if you think that idea is not good enough you can close this issue and we would workaround it in a way similar to what you described even though it is not our favourite solution :)

fredbi:
I think the discussion thread interesting and it is okay to keep it open and collect the opinions from other people.

[...]

PR #2082, opened 2019-10-15
Add flag to allow overwriting protected templates

dominikznidar:
Gives option to the user willing to take the risk to overwrite templates
protected by go-swagger by passing --allow-template-override flag.

Related to #1606


casualjim:
You can use a config file to get this result without changing this.
the docs with some examples are here: https://goswagger.io/use/template_layout.html

dominikznidar:
Thanks @casualjim, didn't notice that feature. 

Tried it out and noticed, I still need to use this newly added flag to be able to use my onw `objectvalidator` template.

Issue #1879, opened 2019-02-11
Generate map[string]json.RawMessage model


sidh:
What is the correct way to generate map[string]json.RawMessage model?

x-go-type does not work in this case:
```
type: object
additionalProperties:
  type: object
  x-go-type:
  import:
    alias: "json"
    package: "encoding/json"
  type: "RawMessage"
```
but does work in this
```
type: object
additionalProperties:
  allOf:
  - type: object
    x-go-type:
    import:
      alias: "json"
      package: "encoding/json"
    type: "RawMessage"
```
but generates an unneeded type that holds json.RawMessage.

fredbi:
Could you please post the generated model and point out what is wrong with it? Thanks.

sidh:
The desired type is map[string]json.RawMessage.

For the first case it generates map[string]interface{}.

For the second case it generates map[string]MyTypeAnon with
```
type MyTypeAnon struct {
  json.RawMessage
}
```

fredbi:
First case looks like a bug. Second case is expected, since the `allOf` makes a new type.

sidh:
Agree on the second case. First case is probably a bug, yes.

Also it seems x-go-type does not work with properties of an object:
```
MyType:
  type: object
  properties:
    prop:
      type: object
      x-go-type:
        import:
          alias: "json"
          package: "encoding/json"
        type: "RawMessage"
```

results in
```
type MyType struct {
  Prop interface{}
}
```
instead of

```
type MyType struct {
  Prop json.RawMessage
}
```

fredbi:
Yes. I just checked in [here](https://github.com/go-swagger/go-swagger/blob/f68196c53821055a3765d1594722e4da46898ade/generator/model.go#L216): x-go-type extension is currently only accounted for definitions, not nested structures. So this is not a bug, but a (limited) feature... :)

I don't think this feature was created in the first place for this use case of hijacking type mappings _inside_ definitions.

A hacky way to go could be to change the type mapping, like defined [here](https://github.com/go-swagger/go-swagger/blob/f68196c53821055a3765d1594722e4da46898ade/generator/formats.go#L141) and define or redefine some format to map as `json.RawMessage`.

sidh:
> A hacky way to go could be to change the type mapping, like defined [here](https://github.com/go-swagger/go-swagger/blob/f68196c53821055a3765d1594722e4da46898ade/generator/formats.go#L141) and define or redefine some format to map as `json.RawMessage`.

[...]

PR #2107, opened 2019-11-13
codescan: Natively support json.RawMessage

### LINKED ISSUE #1622: generate spec: json.RawMessage doesn't represent the intent.
## Problem statement

When generating from source, models with a `json.RawMessage`, it is represented as an array of uint8 values (which is the internal representation of course). Just as a `interface{}` generates a `"type": "object"` it would be nice if `json.RawMessage` did the same.

## Swagger specification

The following model is generated:
```
    "RawMessage": {
      "description": "It implements Marshaler and Unmarshaler and can\nbe used to delay JSON decoding or precompute a JSON encoding.",
      "type": "array",
      "title": "RawMessage is a raw encoded JSON value.",
      "items": {
        "type": "integer",
        "format": "uint8"
      },
      "x-go-package": "encoding/json"
    }
```
## Steps to reproduce

Generate a model with a `json.RawMessage` field.

Command used here: `swagger generate spec -m -i swagger.yaml -o ./swagger.json`

## Environment
swagger version: SHA-1: 8914a0f9bc4c11970552f133f8c9af106c1ac511 (latest master)
go version: 1.10.3
OS: windows/amd64

aeneasr:
Before this patch, `encoding/json.RawMessage` would be interpreted as []uint8, which `encoding/json` is incapable of decoding. This patch addresses that by checking if a type's name is `encoding/json.RawMessage` and if so, sets `type: object`.

In #1622 it was suggested to use `type: string` with `format: binary` which is however not the correct use, because this is not an serialized JSON object but a deserialized JSON object.

If my understanding is correct, `type: string` with `fromat: binary` would expect something like:

```
{ "foo": "[\"bar\"]" }
```

for

```
type f struct { Foo json.RawMessage `json:"foo"` }
```

whereas `json.RawMessage` actually implies:

```
{ "foo": ["bar"] }
```

I've tried to mess around with `x-go-type`

```go
			tgt.AddExtension("x-go-type", map[string]interface{}{
				"import": map[string]interface{}{
					"alias":   "json",
					"package": "encoding/json",
				},
				"type": "RawMessage",
			})
```

but that yielded no result when generating the client code but I believe it to be the right extension here, it does seem however that its applicability is limited as discussed in: https://github.com/go-swagger/go-swagger/issues/1879#issuecomment-462347371

I do understand that `json.RawMessage` can be any defined JSON type, and that this is just a workaround for now, but it's better than what we have so far. I think a `oneOf` would be the appropriate setting here but I'm not sure how to implement that in this particular location. Happy for any pointers though.

Closes #1622 

Related #1879

### DIFF
--- a/codescan/schema.go
+++ b/codescan/schema.go
@@ -263,6 +263,10 @@ func (s *schemaBuilder) buildFromType(tpe types.Type, tgt swaggerTypable) error 
 			tgt.Typed("string", "date-time")
 			return nil
 		}
+		if pkg.PkgPath == "encoding/json" && tio.Name() == "RawMessage" {
+			tgt.Typed("object", "")
+			return nil
+		}
 		cmt, hasComments := s.ctx.FindComments(pkg, tio.Name())
 		if !hasComments {
 			cmt = new(ast.CommentGroup)


aeneasr:
Happy to sign the DCO once I have green light that you'd actually consider merging this.

casualjim:
Thanks. I wonder if that x-go-type definition can't be expanded to support more RawMessage objects from different packages.
Anyway no objection to this PR from me

aeneasr:
> Thanks. I wonder if that x-go-type definition can't be expanded to support more RawMessage objects from different packages.

Maybe, but that's out of my ballpark because I believe this touches lots of internals. Would it be ok with you if this is merged as a workaround with potential downstream fixes to e.g. `x-go-type`?

aeneasr:
DCO is now signed!

fredbi:
Thanks @aeneasr

aeneasr:
@casualjim would it be possible to release this at some point? thank you!

PR #2112, opened 2019-11-19
Fix for path parameters parsing

Emreu:
Fixed parameters parsing for swagger:route annotation.
Previously invalid specification were generated due to parameters defined for path were setting "schema" field (only "type" field should be set). Now "schema" is dropped for in:path parameters.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/go-swagger/go-swagger
+module github.com/emreu/go-swagger
 
 go 1.24.0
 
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/go-swagger/go-swagger
+module github.com/emreu/go-swagger
```
mfranczy: probably by accident, you pushed go.mod from your fork

fredbi:

PR #2114, opened 2019-11-20
CI refactoring

fredbi:
* CI: codegen scripts replaced by a more flexible and readable go executor
* CI: added docker v1.41 spec to canary specs to generate from
* CI: refactored circleCI config with circleci 2.1 commands

> CI: reviewed known codegen failures and provided some slightly different
> "fixed" versions that pass codegen, besides the original version.
> This will allow for a better follow-up when figuring out the impact of some bug fixes.


fredbi:
@casualjim I've looked into fossa to figure out what's wrong with licensing but couldn't really figure out. Any piece of advice on why the license check is failing?

casualjim:
I wonder if this can be moved into an acuatl _test.go file outside of ./hack perhaps with a build tag

fredbi:
Putting this as WIP: I am investigating on how come this addendum makes the license check to break

fredbi:
@casualjim  

**Re location / tag**

The test program is already tagged `// +build ignore`.  We may indeed move it later to the main test suite, that's a good idea.
I'd suggest we see first how it performs for a little while. I has already been useful to me,
since I had to dig again about all the little nitpicks that go wrong with canary specs (to be fixed).
I'd also like to update all our canaries with recent versions of the spec. I've already been testing (with the same test program) the current kubernetes API, but did not land it with this PR.

**Re FOSSA**

Again a nitpick, and most likely a bug in FOSSA CI.

Looks to me that FOSSA is reporting a false negative. I don't know why this particular PR hits this, but I could correct this easily on the fossa for my fork.

Basically, this complains about [this gofontwoff package](https://github.com/shurcooL/gofontwoff) being GPL, which (i) is not the case and (ii) is shipped by Google as part of golang, (iii) anyway is stated by the license policy as a warning
```
Flagged: GPL-3.0-only in gofontwoff   
```

The color package I have imported is not the culprit (testing by removing it altogether: hits the same issue). Further it has no real constraint with its license (i.e "Do What The F. You Want license"...).

This kind of alert can be configured in fossa, but I just don't have access to it. Marking `gofontwoff` as ok makes the run pass.

fredbi:
Thanks Ivan

PR #2117, opened 2019-11-22
Support AllowTemplateOverride in `generate client`

aeneasr:


aeneasr:
Thank you for the quick merge!

PR #2118, opened 2019-11-22
Add .Context to location template context

aeneasr:
This patch adds `.Context` (set using `data`) in GenOpts.location(),
giving more context when rendering the, for example, file_name.

This allows to use configurations such as:

```
layout:
  models:
    - name: definition
      source: asset:model
      target: "{{ joinFilePath .Target .ModelPackage }}"
      file_name: "{{ (snakize (pascalize .Context.OriginalName )) }}.go"
```

This patch is needed because having two models / structs with the same
name in different packages, causes the output files to overwrite
one another. With this patch, it is now possible to prevent that
by using, for example, `.Context.OriginalName` in the `file_name`
template.

Signed-off-by: aeneasr <aeneas@ory.sh>


aeneasr:
Thank you for the quick review! Would love to see this on master if possible :)

PR #2140, opened 2019-12-20
Fix overwriting security definitions on call Server.SetAPI()

h0mjam:
On call Server.SetAPI() trigger internal function configureAPI() which is overwrite above defined (into api object) secirutyDefinitions


casualjim:
can you run `go generate ./generator` so that your template changes are picked up by the build?

h0mjam:
Done `go generate ./generator`

h0mjam:
@casualjim, hi! Can review PR?

generator/bindata.go
```diff
@@ -18,7 +18,7 @@
 // templates/schematype.gotmpl (854B)
 // templates/schemavalidator.gotmpl (31.005kB)
 // templates/server/builder.gotmpl (17.451kB)
-// templates/server/configureapi.gotmpl (6.265kB)
+// templates/server/configureapi.gotmpl (6.421kB)
```
fredbi: @h0mjam hey, thanks for this update.
I see a lot of differences with the generated bindata.
We are using kevin's burke fork github.com/kevinburke/go-bindata
Can you confirm you used that one? Thanks.

https://github.com/go-swagger/go-swagger/blob/master/docs/guidelines/README.md#update-templates

go.sum
```diff
@@ -144,6 +144,8 @@ github.com/hashicorp/hcl v1.0.0/go.mod h1:E5yfLk+7swimpb2L/Alb/PJmXilQ/rhwaUYs4T
 github.com/jessevdk/go-flags v1.4.0 h1:4IU2WS7AumrZ/40jfhf4QVDMsQwqA7VEHozFRrGARJA=
 github.com/jessevdk/go-flags v1.4.0/go.mod h1:4FA24M0QyGHXBuZZK/XkWh8h0e1EYbRYJSGM75WSRxI=
 github.com/jonboulle/clockwork v0.1.0/go.mod h1:Ii8DK3G1RaLaWxj9trq07+26W01tbo22gdxWY5EU2bo=
+github.com/jteeuwen/go-bindata v3.0.7+incompatible h1:91Uy4d9SYVr1kyTJ15wJsog+esAZZl7JmEfTkwmhJts=
```
fredbi: I get it that you've built locally this and it added to the the go.mod, but this is not a build dependency. Please remove this or just run `go mod tidy`

h0mjam: Clean after run `go mod tidy`

generator/bindata.go
h0mjam: I'am install `go-bindata` from brew. go-bindata should be github.com/kevinburke/go-bindata. Can do re-generate if needed.

fredbi:
@h0mjam  we are just a commit signature away... Hang on, this all looks good to me.

PR #1751, opened 2018-10-11
feature: x-omitempty extension for general types

zhuangqh:
### Ⅰ. Describe what this PR did

support `x-omityempty` extension for general type. 
Now we can use `x-omityempty` to control presence of omitempty tag to be used by JSON Marshaler.

By default, `x-omitempty` is false for array type, true for other.

### Ⅱ. Does this pull request fix one issue?

#1623 

### III. Describe how to verify it

I created a test for this feature.

```yaml
---
swagger: '2.0'
info:
  version: 0.0.0
  title: Breaking API as described by christianklotz
paths:
  /stores:
    get:
      operationId: getFoo
      responses:
        200:
          description: 'ok'
          schema:
            type: array
            items:
              $ref: '#/definitions/Foo'
definitions:
  Foo:
    type: object
    properties:
      arrayHasOmitEmptyTrue:
        type: array
        x-omitempty: true
        items:
          type: string
          enum:
          - some
          - values
      arrayHasOmitEmptyFalse:
        type: array
        x-omitempty: false
        items:
          type: string
          enum:
          - some
          - values
      arrayNoOmitEmpty:
        type: array
        items:
          type: string
          enum:
          - some
          - values
      generalHasOmitEmptyTrue:
        type: string
        x-omitempty: true
      generalHasOmitEmptyFalse:
        type: string
        x-omitempty: false
      generalNoOmitEmpty:
        type: string
      refHasOmitEmptyTrue:
        x-omitempty: true
        $ref: '#/definitions/bar'
      refHasOmitEmptyFalse:
        x-omitempty: false
        $ref: '#/definitions/bar'
      refNoOmitEmpty:
        $ref: '#/definitions/bar'
  bar:
    type: object
    x-nullable: false
    properties:
      prop:
        type: string
```

should generate model like

```go
type Foo struct {
			            	
	// array has omit empty false
	ArrayHasOmitEmptyFalse []string `json:"arrayHasOmitEmptyFalse"`
			            	
	// array has omit empty true
	ArrayHasOmitEmptyTrue []string `json:"arrayHasOmitEmptyTrue,omitempty"`
			            	
	// array no omit empty
	ArrayNoOmitEmpty []string `json:"arrayNoOmitEmpty"`
			            	
	// general has omit empty false
	GeneralHasOmitEmptyFalse string `json:"generalHasOmitEmptyFalse"`
			            	
	// general has omit empty true
[...]

Issue #1623, opened 2018-07-18
Proposal: x-omitempty for general types


zhuangqh:
## Problem statement

Currently, we use `x-omitempty` to exclude null array value. related PR #1189 
Maybe `x-omitempty` can support general type to control the `omityempty` tag.

```yaml
Foo:
  type: "object"
  properties:
    hasOmitEmptyTrue:
      type: string
      x-omitempty: true
    hasOmitEmptyFalse:
      type: string
      x-omitempty: false
```
produce
```go
type Foo struct {
	// has omit empty false
	HasOmitEmptyFalse string `json:"hasOmitEmptyFalse"`
	// has omit empty true
	HasOmitEmptyTrue string `json:"hasOmitEmptyTrue,omitempty"`
}
```

## Environment
swagger version: x.x.x  
go version: x.x.x  
OS:

fredbi:
Closed with #1751

PR #2148, opened 2019-12-26
Fixed x-omitempty in the case of $ref'ed schemas

### LINKED ISSUE #2116: x-omitempty handling is incorrect with $ref
## Problem statement

The `x-omitempty` property is incorrectly processed in the presence of `$ref`.

When the referenced definition contains `x-omitempty`, it appears to be ignored.

When the reference has a sibling `x-omitempty` key, the `x-omitempty` is honoured but should not have been.

By comparison, `x-nullable` seems to work as expected in similar circumstances.

## Swagger specification
```yaml
swagger: "2.0"
info:
  title: ""
  version: ""
paths:
  /:
    get:
      description: ""
      responses:
        200:
          description: ""
definitions:

  Case-1--Fail--omitempty-false-not-hoisted-by-ref:
    type: object
    properties:
      Body:
        $ref: '#/definitions/Object-with-omitempty-false'

  Case-2--Fail--omitempty-false-not-overridden-by-ref-sibling:
    type: object
    properties:
      Body:
        $ref: '#/definitions/Object-with-omitempty-true'
        x-omitempty: false

  Case-3--Pass--object-nullable-false-hoisted-by-ref:
    type: object
    properties:
      Body:
        $ref: '#/definitions/Object-with-nullable-false'

  Case-4--Pass--object-nullable-false-overriden-by-ref-sibling:
    type: object
    properties:
      Body:
        $ref: '#/definitions/Object-with-nullable-true'
        x-nullable: false

  Object-with-omitempty-true:
    type: object
    properties:
      Data:
        type: object

  Object-with-omitempty-false:
    type: object
    properties:
      Data:
        type: object
    x-omitempty: false

  Object-with-nullable-true:
    type: object
    properties:
      Data:
        type: object
    x-nullable: true

  Object-with-nullable-false:
    type: object
    properties:
      Data:
        type: object
    x-nullable: false
```

## Steps to reproduce
`swagger generate model`

```go
// Case1FailOmitemptyFalseNotHoistedByRef case 1 fail omitempty false not hoisted by ref
// swagger:model Case-1--Fail--omitempty-false-not-hoisted-by-ref
type Case1FailOmitemptyFalseNotHoistedByRef struct {

        // body
        Body *ObjectWithOmitemptyFalse `json:"Body,omitempty"`
}

// Case2FailOmitemptyFalseNotOverriddenByRefSibling case 2 fail omitempty false not overridden by ref sibling
// swagger:model Case-2--Fail--omitempty-false-not-overridden-by-ref-sibling
type Case2FailOmitemptyFalseNotOverriddenByRefSibling struct {

        // body
        Body *ObjectWithOmitemptyTrue `json:"Body"`
}
```

Given that case 1 fails, it's actually rather convenient for our real use-case that case 2 "works": it allows us ultimately to get the `omitempty` behaviour we want.

## Environment
swagger version: master v0.21.0-13-g2600b738
go version: 1.13.4 
OS: linux/amd64

fredbi:
**ATTENTION: there is a subtle change in here that SHOULD be discussed before blindly merging!!!**

* fixes #2116
* tests are embedded with existing testcase for #1623 (added by #1751)
* ensure arrays _of primitives_ are not omitempty-ed by default, including arrays declared as properties
* arrays as aliased types used as properties remain empty-omitted by default (this may be altered with the x-omitempty extension)

@zhuangqh this brings some slight changes to your PR #1751. Visible outcome is:
- extensions declared as siblings to $ref are ignored
- properties which _are_ arrays are not always empty-omitted by default: to avoid introducing breaking changes I adopted the rule array of primitive vs aliased array

@zhuangqh since you brought in that change in the first place, I'd like to have your opinion on the second point

### DIFF
--- a/fixtures/bugs/2116/.gitignore
+++ b/fixtures/bugs/2116/.gitignore
@@ -1,0 +1,3 @@
+gen-*
+!gen-fixtures.sh
+*.log
--- a/fixtures/bugs/2116/fixture-2116.yaml
+++ b/fixtures/bugs/2116/fixture-2116.yaml
@@ -10,7 +10,6 @@ paths:
         200:
           description: ""
 definitions:
-
   Case-1--Fail--omitempty-false-not-hoisted-by-ref:
     type: object
     properties:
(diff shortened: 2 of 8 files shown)


fredbi:
codegen now encounters out of memory failures from time to time. Will have to split that in 2 workflows.

fredbi:
@casualjim I don't feel at ease with the smarter-and-half change here. I am going to add another change here to change only the $ref behavior and not regular array properties. Please do not merge yet.

PR #2151, opened 2019-12-26
Client template interfaces

### LINKED ISSUE #1138: Could Client be a concrete implementation of an interface?
This would allow for easy mocking out of clients.

https://github.com/go-swagger/go-swagger/blob/f8fdd34ee024def2304ce2417b6bab6046b8cb02/generator/templates/client/client.gotmpl#L33

Or is there a better way to do the mocking?

fredbi:
Trying to complete #1981 

- fixes #1138


fredbi:
Supersedes #1981 

**NOTE**: TODO regen examples

rokf:
I'm sorry that I didn't put more effort into finishing what I started. I got a bit confused by the pipeline failures since I'm not too familiar with the codebase.

I can help with the review process though, as I've ended up using a slightly modified version of these templates here.

generator/templates/client/client.gotmpl
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 // New creates a new {{ humanize .Name }} API client.
-func New(transport runtime.ClientTransport, formats strfmt.Registry) *Client {
+func New(transport runtime.ClientTransport, formats strfmt.Registry) ClientService {
```
rokf: I've ended up using a more dynamic interface name instead of `ClientService`.

```
{{ pascalize .Name }}Client
```

This way there is less ambiguity when there are multiple clients in the codebase.

fredbi: @rokf I am really not sure this is a good idea: when organizing operations with tags, we end up generating `{{ pascalize .Name }}` packages. IMHO, the best way to avoid ambiguity is to generate client with an appropriate package target, so we use it like `target.ClientService`.

PR #2158, opened 2019-12-30
CI codegen

fredbi:
* CI codegen: fixed codegen failures on fixtures not producing models
* CI codegen: added filter to exclude specs from test job


fredbi:
PR actually passes. Codecov got the maths wrong once again

PR #2159, opened 2019-12-30
fix: missing adding postDecls in response map field

### LINKED ISSUE #2145: model will not be generate if it's used as a map's value type in response body
## Problem statement
if we have a swagger:response struct's body is a map[string]XXX, XXX's swagger:model will not be generated

## Swagger specification
using v0.21.0, not sure about other versions

## Steps to reproduce

```
// swagger:response getProductsResponse
type GetIAPProductResponse struct {
	// in:body
	Body map[string]Product `json:"body"`
}

// swagger:model
type Product struct {
	ID string
}
```
after swagger generate spec, will found that Product will not be generate

```
"getProductsResponse": {
      "description": "a productID - productData dictionary",
      "schema": {
        "type": "object",
        "additionalProperties": {
          "$ref": "#/definitions/Product"
        }
      }
    },
```
will be not able to find `"#/definitions/Product"`

## Environment
swagger version: 0.21.0
go version: 1.13.3
OS:   Mac OS X 10.13.6

autumn31:
I think this it the root cause of https://github.com/go-swagger/go-swagger/issues/2145

* fixes #2145


codescan/responses.go
```diff
@@ -185,6 +185,7 @@ func (r *responseBuilder) buildFromField(fld *types.Var, tpe types.Type, typable
 		if err := sb.buildFromType(ftpe.Elem(), schemaTypable{schema, typable.Level() + 1}); err != nil {
 			return err
 		}
+		r.postDecls = append(r.postDecls, sb.postDecls...)
```
fredbi: Thank you for digging this out.
Would it be possible to assert the fix with some unit test case?

autumn31: I just add some test code in `codescan/responses_test.go` and some file in fixtures/goparsing for test case . 
but I have no idea why after submitting this commit will cause somewhere fails in `go test -v -timeout 60m hack/codegen_nonreg_test.go -args -fixture-file codegen-fixtures.yaml -skip-models -skip-full-flatten`

fredbi:
yeah the CI failure is my bad.

autumn31:
ok, just let me know if there anything I need to do . 
On the other hand, just curious what does `go test -v -timeout 60m hack/codegen_nonreg_test.go -args -fixture-file codegen-fixtures.yaml -skip-models -skip-full-flatten` do or test? do I need to add some test case here?

PR #2165, opened 2020-01-07
ignore not exported params fields

### LINKED ISSUE #2164: Swagger adds to a spec not exported parameters fields
## Problem statement
Swagger adds not exported parameters fields to spec, I believe that should not happen.

## Swagger specification
2.0

## Steps to reproduce
Example structure with not exported fields which are later added to spec

```
// Example
// swagger:parameters example
type Example struct {
	// in: query
	Foo  string
	bar   string
}
```  

as a result in a spec, you will find a `bar` param which should not be there.

## Environment
swagger version: master branch  
go version: 1.13.5  
OS: Fedora 31

mfranczy:
fixes #2164

### DIFF
--- a/codescan/parameters.go
+++ b/codescan/parameters.go
@@ -286,6 +286,11 @@ func (p *parameterBuilder) buildFromStruct(decl *entityDecl, tpe *types.Struct, 
 			continue
 		}
 
+		if !fld.Exported() {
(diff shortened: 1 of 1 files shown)


mfranczy:
@fredbi do I need to add something more to the PR that `codecov/patch` can pass?

Issue #2167, opened 2020-01-07
Enable x-go-name in parameters


maxatome:
https://github.com/go-swagger/go-swagger/blob/master/generator/templates/server/urlbuilder.gotmpl#L109-L114

What do you think of allowing `x-go-name` in parameters:
```
  {{ range .PathParams }}
  {{- $goName := index .Extensions "x-go-name" }}
  {{ if $goName }}{{ $goName }}{{ else }}{{ pascalize .ID }}{{ end }} {{.GoType}}
  {{- end }}
  {{ range .QueryParams }}
  {{- $goName := index .Extensions "x-go-name" }}
  {{ if $goName }}{{ $goName }}{{ else }}{{ pascalize .ID }}{{ end }} {{ if and (not .IsArray) .IsNullable }}*{{ end }}{{.GoType}}
  {{- end }}
```
Not really needed on `PathParams`, but to be consistent why not.

On `QueryParams` it allows to expose params like `string` for example (compilation error, as `String` method is defined too).

maxatome:
Shorter version:
```
  {{ range .PathParams }}
  {{ or (index .Extensions "x-go-name") (pascalize .ID) }} {{.GoType}}
  {{- end }}
  {{ range .QueryParams }}
  {{ or (index .Extensions "x-go-name") (pascalize .ID) }} {{ if and (not .IsArray) .IsNullable }}*{{ end }}{{.GoType}}
  {{- end }}
```

fredbi:
@maxatome how would this work as far as parameter structures and validation code are generated for client and server? could you please detail the intent in just changing the URL builder?

maxatome:
@fredbi I didn't use client generation, that is why I forgot it, sorry.
Look at #2168, I have just added the client part in generator/templates/client/parameter.gotmpl

fredbi:
I've looked at the PR. I think it is breaking the generator. Part of it is because of the existing mess in templates (switching from `.ID` to `.Name` without reason, btw this would deserve a fix too...). Part of this is that when following a template-based approach, you are compelled to figure out every impact in all templates (client, server, validator, urlbuilder). I would recommend to apply the naming change in `operation.go` instead, so it propagates to every templates.

maxatome:
@fredbi merged #2168, thanks!

PR #2168, opened 2020-01-07
feat(parameters): query & path parameters now accept x-go-name

maxatome:
Related to #2167

### DIFF
--- a/fixtures/enhancements/2167/swagger-error.yml
+++ b/fixtures/enhancements/2167/swagger-error.yml
@@ -1,0 +1,25 @@
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: Example
+schemes:
+  - http
+produces:
+  - text/plain
+paths:
+  /test/{test_name}:
+    get:
+      operationId: xGoNameInParams
+      summary: Says hello
+      parameters:
+        - in: path
+          name: test_name
+          x-go-name:
+            - bad
+          required: true
+          type: string
+      responses:
+        200:
+          description: A hello world string
+          schema:
+            type: string
--- a/fixtures/enhancements/2167/swagger.yml
+++ b/fixtures/enhancements/2167/swagger.yml
@@ -1,0 +1,37 @@
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: Example
+schemes:
+  - http
+produces:
+  - text/plain
+paths:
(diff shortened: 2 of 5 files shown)


generator/templates/client/parameter.gotmpl
```diff
@@ -91,7 +91,7 @@ type {{ pascalize .Name }}Params struct {
   {{ blockcomment .Description }}
 
   {{ end }}*/
-  {{ pascalize .ID }} {{ if and (not .IsArray) (not .IsMap) (not .HasDiscriminator) (not .IsInterface) (not .IsStream) (or .IsNullable  ) }}*{{ end }}{{ if not .IsFileParam }}{{ .GoType }}{{ else }}runtime.NamedReadCloser{{end}}
+  {{ or (index .Extensions "x-go-name") (pascalize .ID) }} {{ if and (not .IsArray) (not .IsMap) (not .HasDiscriminator) (not .IsInterface) (not .IsStream) (or .IsNullable  ) }}*{{ end }}{{ if not .IsFileParam }}{{ .GoType }}{{ else }}runtime.NamedReadCloser{{end}}
```
fredbi: @maxatome, there is also a templates/server/parameter.gotmpl

Overall, I am not convinced of the template-based approach to alias parameters. Have you thought about changing it here instead? https://github.com/go-swagger/go-swagger/blob/1c98855b472d8782c366459428ec2a5e8339ccf2/generator/operation.go#L741

maxatome: You are right it is a way simpler :)
Done.

fredbi:
@maxatome did you test it with client generation, I mean does your spec produce some code that builds from both client and server? Hint: you can add this to our CI spec build here: https://github.com/go-swagger/go-swagger/blob/master/hack/codegen-fixtures.yaml (note: since you have a "bad spec", you should specify only the one that is supposed to work, with its full path).

maxatome:
@fredbi yes it works for client generation too.
Note I added unit tests to check that case as well as server one.

fredbi:
I've seen the unit test, I am talking about making sure (manually, I am not asking for more automated testing here) that the client compiles properly.

maxatome:
In the `client/xxx_parameters.go` files, if we have:
- `name: foo`
- `x-go-name: bar`

We have the methods `WithBar(foo string)` and `SetBar(foo string)`.

Issue #2054, opened 2019-08-29
extension fields in info.contact do not propagate to embedded spec


mrwonko:
## Problem statement

I need to add an extension field `info.contact.x-teams` to the spec served by the generated server, but the extension field is dropped during generation. I don't need to access it in my code, I just need it to get passed on so tools parsing our swagger specs can make use of it.

## Swagger specification

```yaml
swagger: "2.0"
info:
  title: "go-swagger extension field bug repro"
  version: 0.1-bugrepro
  contact:
    email: "this-is-retained@example.com"
    x-teams: "I'd love for this field to show up in in the embedded spec"
  x-ok: "here extension fields work"
paths:
  "/foo":
    get:
      responses:
        200:
          description: OK
```

## Steps to reproduce

```sh
swagger generate server --spec swagger.yaml
```

Resulting embedded_spec:

```js
{
  "swagger": "2.0",
  "info": {
    "title": "go-swagger extension field bug repro",
    "contact": {
      "email": "this-is-retained@example.com"
    },
    "version": "0.1-bugrepro",
    "x-ok": "here extension fields work"
  },
  "paths": {
    "/foo": {
      "get": {
        "responses": {
          "200": {
            "description": "OK"
          }
        }
      }
    }
  }
}
```

## Environment
swagger version: v0.20.1
go version: go1.12.9 darwin/amd64  
OS: macOS Mojave 10.14.6

fredbi:
Good catch go-openapi/spec should define both contacts and license to be vendor extensible

fredbi:
If anyone is interested in helping out here, here are some guidelines

https://github.com/go-openapi/spec/blob/8557d72e4f077c2dbe1e48df09e596b6fb9b7991/contact_info.go#L20 should be extended with this https://github.com/go-openapi/spec/blob/8557d72e4f077c2dbe1e48df09e596b6fb9b7991/info.go#L73.

For licensing: https://github.com/go-openapi/spec/blob/8557d72e4f077c2dbe1e48df09e596b6fb9b7991/license.go#L20

On top of that, the spec mixin tool at github.com/go-openapi/analysis needs to know about this too: 
https://github.com/go-openapi/analysis/blob/9864a87593df5ab83eac00921906932d1d77fae2/mixin.go#L332

kul-amr:
Hey - I am new to OSS contribution. Can I work on this task if no one else is attempting ?

fredbi:
[...]

PR #2171, opened 2020-01-09
changes for #2054 - extension fields in info.contact do not propagate…

kul-amr:
This is a bug fix for  go-swagger/go-swagger#2054. This fix will include extension/custom fields from Contact and License to the spec served by the generated server.


fredbi:
@kul-amr thanks for your contribution. I think we have fixed our CI issues: you may rebase and PGP-sign you commit. This should be good to go.

casualjim:
looks like those fields are missing from the spec repo

kul-amr:
@casualjim I had added the fields in spec and have created a pull request https://github.com/go-openapi/spec/pull/113

PR #2190, opened 2020-01-23
Fix client generation's mapvalidator with additionalProperties

hmalphettes:
The value expression must use the key on the iterated collection

Code generated without this patch:

```go
func (m *customIntegrations) validateDocuments(formats strfmt.Registry) error {

	if swag.IsZero(m.Documents()) { // not required
		return nil
	}

	for k := range m.Documents() {

		if err := validate.Required("documents"+"."+k, "body", m.Documents[k]); err != nil {
			return err
		}
		if val, ok := m.Documents[k]; ok {
			if err := val.Validate(formats); err != nil {
				return err
			}
		}

	}

	return nil
}
```

This patch fixes the problematic `m.Documents[k]`
I hope this helps.


casualjim:
I think something went wrong generating the bindata. We use this tool: https://github.com/kevinburke/go-bindata
Can you try to run `go generate ./generator` with that tool first on your path? Currently the tests fail because it can't find files in `/Users/hugues`

hmalphettes:
@casualjim Thanks a lot for looking into this. Apologies for not doing this myself.
I have followed your instructions and forced pushed after checking that my own file system does not appear anymore.

Thanks for your attention!

casualjim:
@fredbi can you see why that test failed? I can't find what the failure is in this run.

fredbi:
@casualjim ok done: codegen CI sometimes hit the memory limit whenever it is picking  the largest tests to run in parallel. Usually, running again the CI workflow (sometimes several times...) is good enough.
The hint to understand this is when one of the go command reports a 127 error code and there is no other output about the specific failure. I am trying to address that by throttling the degree of parallel tests during codegen.

fredbi:
@hmalphettes could you please PGP-sign you commits?

hmalphettes:
> @hmalphettes could you please PGP-sign you commits?
Done

I ran the test command that fails from my machine: `go test -v -timeout 60m hack/codegen_nonreg_test.go -args -fixture-file codegen-fixtures.yaml -skip-models -skip-full-flatten`
Test 1277 that is reported as failed by circleci is green on my machine.

fredbi:
@hmalphettes yes it is a transitory memory shortage issue on the CI, which makes test fail from time to time. I am rerunning the CI workflow. I think PR #2193 is addressing this issue

hmalphettes:
@fredbi @casualjim thanks for reviewing this patch so quickly and many thanks for your work on go-swagger in general!

PR #2197, opened 2020-01-27
added CLI option to generate client together with server

fredbi:
Usage: trying to reduce the number of generations needed to explore all use cases during CI jobs


youyuanwu:
Generate client with server could be helpful. Is this feature enabled? Why is this closed?

casualjim:
because this is easily achieved by crafting 2 `//go:generate` lines that share the models folder, that gives you the same result without more code options to support.

youyuanwu:
Ok. So this will generate model twice? I was doing generating client skip model and specify the model import path. I can definitely use this, but a separate cli option can be easier to use.

Issue #1679, opened 2018-09-01
"swagger generate model" does nothing if no paths are defined


atombender:
## Problem statement

If a spec doesn't define any paths, then you cannot generate any model code from it.

This makes `go-swagger` apparently useless for projects that only want to leverage its support for schema validation and serialization/deserialization. (Or at least I can't figure out how to use it in "schema only" mode.)

(If there are any other Go tools that generate model/serialization code from OpenAPI, please let me know.)

## Swagger specification

```yaml
swagger: "2.0"
info:
  title: Something
  version: 0.1.0
paths: {}
definitions:
  myConfig:
    type: object
    properties:
      name:
        type: string
```

## Steps to reproduce

```shell
swagger generate model -f ./swagger.yml
2018/09/01 13:46:56 validating spec .../swagger.yml
2018/09/01 13:46:56 preprocessing spec with option:  minimal flattening
no operations were selected
```

## Environment

swagger version: 0.16.0
go version: 1.11
OS: macOS 10.13.6

fredbi:
As a workaround, it is sufficient to declare just one dummy path operation to generate the whole model.

fredbi:
That's right that generating model only should support incomplete swagger specs. Indeed what it really needs is just the list of definitions, but that would not be a valid swagger spec.

A PR to get a more tolerant generate model CLI here could be useful.

sachja:
+1 . This should work like google protobuf where we can just have object definitions and the model can be generated. 
Is there any progress for this request.

fredbi:
Since #2204, model generation supports incomplete specs with only definitions.There is a CLI option to activate this and bypass the full swagger spec validation.

PR #2204, opened 2020-02-02
add option in model generation to output all definitions regardless of usage

dan-j:
Attempt at solving #1679 

I have manually tested this will output all the expected go code for definitions defined in a swagger spec where `paths: {}`.

This is my first contribution, so apologies if I've missed anything. I'm guessing I might need to add some tests, any pointers into that would be appreciated.


dan-j:
Created a new PR on a non-master branch from my fork

PR #2205, opened 2020-02-02
add option in model generation to output all definitions regardless of usage

dan-j:
Attempt at solving #1679

I have manually tested this will output all the expected go code for definitions defined in a swagger spec where `paths: {}`.

This is my first contribution, so apologies if I've missed anything. I'm guessing I might need to add some tests, any pointers into that would be appreciated.

Signed-off-by: dan-j <5727701+dan-j@users.noreply.github.com>


dan-j:
FYI in case anybody noticed my previous MR.. It was on `dan-j:master`, and the CI checks failed but appeared to not be related to my changes. I noticed other PRs which pass were created from a branch different from `master` so I've recreated it on a different branch

PR #2209, opened 2020-02-03
fix go-swagger#2066 issue

mfranczy:
@fredbi @casualjim instead of copying the map  before iteration I removed the place where `Models` is being modified, I think that was wrong, at this stage model should be known and embedded should not be added as a "standalone" model. Does that make sens to you? I updated related tests too.

Please take a look, thanks.

EDIT: I changed approach, to be sure that all models are processed (with extra ones), I added a new map `ExtraModels` which is joined later to `Models`. `Models` cannot be updated during iteration.

### DIFF
--- a/codescan/application.go
+++ b/codescan/application.go
@@ -291,7 +291,7 @@ func (s *scanCtx) FindModel(pkgPath, name string) (*entityDecl, bool) {
 		}
 	}
 	if decl, found := s.FindDecl(pkgPath, name); found {
-		s.app.Models[decl.Ident] = decl
+		s.app.ExtraModels[decl.Ident] = decl
 		return decl, true
 	}
 	return nil, false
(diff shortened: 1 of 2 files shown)


codescan/schema_test.go
```diff
@@ -793,6 +793,7 @@ func TestEmbeddedAllOf(t *testing.T) {
 	models := make(map[string]spec.Schema)
 	require.NoError(t, prs.Build(models))
 	schema := models["AllOfModel"]
+	require.Len(t, models, 1)
```
casualjim: This is odd I think because we do have a case for allOf where we do actually want extra models to be generated and end up as separate definitions

codescan/application.go
```diff
@@ -291,7 +291,7 @@ func (s *scanCtx) FindModel(pkgPath, name string) (*entityDecl, bool) {
 		}
 	}
 	if decl, found := s.FindDecl(pkgPath, name); found {
-		s.app.Models[decl.Ident] = decl
```
mfranczy: this gives inconsistent result

> If an entry is created during iteration, that entry may or may not be produced during the iteration.

Issue #2234, opened 2020-02-20
Add middleware router or a kind of path matching


ilyakaznacheev:
## Problem statement

The generated server is nice, but internal routing is hidden. Without any routing or way to interact with a path, it becomes painful to deal with middleware which should apply to only one path or a group of paths by a prefix.

It would be really great to add the out-of-the-box feature to add HTTP middleware to certain routes or groups, which will make it much easier for everyday use.

It also would be nice to make it available not inside `configure_***.go`, but also outside the package using global methods `restapi.Server`.

ilyakaznacheev:
It would be even more useful if it can use swagger tags as an alternative to the path.

fredbi:
closed with #2236

PR #2235, opened 2020-02-24
Add middleware setter for a route (path prefix)

ilyakaznacheev:
This change adds a method `api.AddMiddlewareFor` that allows binding an HTTP middleware to the actual method:path handler.

Related to #2234 

- [x] template
- [x] bindata
- [x] documentation
- [x] tests ok

Now there is no simple way to add routed middleware, i.e. a middleware that will run for a certain path and method only. It is still possible to do that in `setupMiddlewares` function, but it's difficult to do that for an average user (in comparison with easy ways to do that in libraries like Gorilla or Chi).

I'm adding a method that will simply wrap an existing handler into a middleware if such a combination of an HTTP method and path exists.
It also runs `api.Init()` to prevent empty handler map to be accessed and avoid complex state dependency.

This method can be used inside of `configure_xxx.go` for those who use the generated server as an entry point, and also outside of this file, who use go-swagger as a pluggable API server. Many projects do it this way, and it is even more difficult to setup middlewares from the outside. So would be great to have such a nice and clean opportunity to do that the way other routing libraries do that.

It would be a really big help for those who want to add middlewares as simple and as fast as possible without tricks. It also will allow adding existing middlewares from external libraries (like auth, monitoring, logging, etc.) in only one line of code like that:

```go
api.AddMiddlewareFor("POST", "/user", jwtMiddleware)
```


ilyakaznacheev:
Closed because I fucked up with signing

Issue #353, opened 2016-03-03
Generated models annotated with json when the media type is application/xml


stoyanr:
We are using a swagger model in which the produced and consumed media type for all operations is "application/xml". However, in all generated model structs, fields are still annotated with `json:`, while in this case they should be annotated with `xml:` instead (or in addition). Simply replacing `json:` with `xml:` results in an (almost) usable model.

This is related to #183. Currently, by default there are no also XML consumers and producers in the transport, but they are also rather trivial to add.

fredbi:
parent: #183

fredbi:
Resolved with #2241

Issue #183, opened 2015-12-29
Add support for XML marshalling


casualjim:
Currently the support for xml is underwhelming, it should at least know how to serialize things to XML

casualjim:
This ticket is labeled with raspberry pi 3 bounty. Solve the ticket and I'll send you a pi

PR #2241, opened 2020-03-10
Basic support for XML model generation

nubenum:
This is related to #183 and #353. It enables the addition of xml struct tags to generated fields in order to be able to unmarshal and (theoretically) marshal XML.

Now, XML struct tags are added everywhere (not only for fields where xml.name is explicitly set, like before). I don't think there is a straightforward way to ensure that a given schema definition will never be used to transmit XML? Consequently, this change breaks lots of tests, which I made green
replacing `` `json:\\"([^,"]+)([^"]*)\\"` `` with `` `json:\\"$1$2\\" xml:\\"$1\\"` ``
in the right (TM) places.

Searching through the codebase with the above regex yields more results that I haven't replaced, particularly `json:\"-\"`. This is because I left the insertion of the XMLName as is, i.e. it is only used in structfield.gotmpl and not in schemabody.gotmpl and allofserializer.gotmpl. As far as I can tell, this means that currently no inline nested schema definitions can be used with XML.

Before changing this, one should IMHO think about refactoring the templates to a) remove redundancy of json struct tag insertion and b) possibly make json and xml struct tag insertion work the same way, i.e. either both concatenating the tags beforehand or both constructing them in the templates. For the same reason, omitempty is currently not set for XML.

I added a simple test for the model generation with XML attributes and custom element names. Retrieving an XML response works for me now. I did not try sending an XML request body.

I'm not at all familiar with the codebase, so please excuse major bloopers and review thoroughly, if it's even worth the effort for such a hacky feature.

**TL;DR:**
added/improved support for:
* xml struct tag generation without setting explicit swagger xml.name
* setting custom swagger xml.name
* setting swagger xml.attribute
* receiving XML
* \+ sending XML
* \+ omitempty

missing support:
* inline nested schema definitions
* ~sending XML (?)~
* namespaces
* wrapping arrays
* ~omitempty~
* additional properties
* ...

See also: https://swagger.io/docs/specification/data-models/representing-xml/


nubenum:
For future reference and since it is not (yet) documented anywhere, my setup to consume an XML API:
```
import (
	"github.com/go-openapi/runtime"
	httptransport "github.com/go-openapi/runtime/client"
        "github.com/go-openapi/strfmt"
	apiclient "[...]/client"
	"os"
)
r := httptransport.New(apiclient.DefaultHost, apiclient.DefaultBasePath, apiclient.DefaultSchemes)
r.DefaultAuthentication = httptransport.BearerToken(os.Getenv("ACCESS_TOKEN"))
r.DefaultMediaType = runtime.XMLMime
r.Consumers = map[string]runtime.Consumer{
	runtime.XMLMime: runtime.XMLConsumer(),
}
r.Producers = map[string]runtime.Producer{
	"application/xhtml+xml": runtime.XMLProducer(),
}
client := apiclient.New(r, strfmt.Default)
res, err := client.Operations.MyGreatEndpoint()
```

casualjim:
the integration test fails: 
```
--- FAIL: TestCodegen/fixtures/codegen/xml-model.yml (0.00s)
        --- FAIL: TestCodegen/fixtures/codegen/xml-model.yml/expand_spec_run (0.48s)
        --- FAIL: TestCodegen/fixtures/codegen/xml-model.yml/minimal_flatten_spec_run (0.54s)
```
you can run those tests locally with:
```sh
go test -v -timeout 20m -parallel 3 hack/codegen_nonreg_test.go -args -fixture-file codegen-fixtures.yaml -skip-models -skip-full-flatten
```

generator/moreschemavalidation_fixtures_test.go
```diff
@@ -39,19 +39,19 @@ func initFixture2116() {
 	flattenRun := f.AddRun(false).WithMinimalFlatten(true)
 
 	flattenRun.AddExpectations("case1_fail_omitempty_false_not_hoisted_by_ref.go", []string{
-		"Body *ObjectWithOmitemptyFalse `json:\"Body\"`",
+		"Body *ObjectWithOmitemptyFalse `json:\"Body\" xml:\"Body\"`",
```
fredbi: @nubenum you would save yourself a lot of pain updating all the model tests by making the systematic addition of XML tags opt-in. I _do_ think this is a nice addition, but making it appear everywhere ,every time is simply causing a lot of impacts.

generator/moreschemavalidation_fixtures_test.go
```diff
@@ -75,28 +75,28 @@ func initFixture2116() {
...
 	}, todo, noLines, noLines)
 
 	flattenRun.AddExpectations("object_with_nullable_true.go", []string{
-		"Data interface{} `json:\"Data,omitempty\"`",
+		"Data interface{} `json:\"Data,omitempty\" xml:\"Data\"`",
 	}, todo, noLines, noLines)
 
 	flattenRun.AddExpectations("object_with_omitempty_false.go", []string{
-		"Data interface{} `json:\"Data,omitempty\"`",
+		"Data interface{} `json:\"Data,omitempty\" xml:\"Data\"`",
```
fredbi: Why don't we have support for `omitempty` in the case of XML serialization?

generator/moreschemavalidation_fixtures_test.go
nubenum: This is a good idea. However, I have decided to reuse the --default-produces and --default-consumes flags for that. If either of them is set to xml, xml struct tags will be added. What do you think, do you prefer a separate flag `--with-xml`?

generator/moreschemavalidation_fixtures_test.go
nubenum: I have now added omitempty support. As stated above, the reasons for me to not add it in the first place were that
* I proposed to first think about refactoring the json struct tag templates. I have now decided to put the entire logic for xml struct tags into the buildXMLNameWithTags function. This is a deliberate deviation from the json struct tags implementation.
* omitempty is only relevant for serializing, not for deserializing (cf. also https://golang.org/pkg/encoding/xml/#Marshal). I have now additionally manually tested the generation of a server which produces xml. Am I missing something or aren't there any integration tests that actually test serializing and deserializing requests and responses?

nubenum:
@casualjim Yes, unfortunately I missed this. When running like this locally, quite a few other tests fail for me. Hope it'll work now. Is there any getting started documentation on how to setup the local environment so that all tests run?

@fredbi Added omitempty and enabling xml struct tags generation via --default-(consumes|produces). We'll need to add documentation for that, no matter how we're doing it.

nubenum:
No problem, I wasn't prepared that it would already have been merged :)

Auto-detection would be nice, but as far I saw, it will be difficult. I think we do not have the information handy whether some endpoint somewhere that uses the model is producing/consuming XML.

My main open points are:
* Integration tests of actual (de-)serialization of requests/resonses
* inline nested schema definitions (other gotmpls)
* documentation. We should definitely add some, otherwise, this is of no value. Do you have a proposition for a place where it fits?

casualjim:
You can add the docs here maybe: https://github.com/go-swagger/go-swagger/blob/master/docs/generate/server.md

Currently we use the mime types in the accept headers to work out which output the client wants and then use the appropriate serializer for it. For the input we use the content type header to do that.

I don't think we actually have an integration test for the json version either, anything there is of course welcome.

fredbi:
@nubenum yes I've merged this because I had the feeling it is already useful as it is, in spite of not fully covering the topic.

I've created this issue https://github.com/go-swagger/go-swagger/issues/2256, so we may follow up how we move further on with enhancing XML support.

Issue #303, opened 2016-02-19
option for case insensitivity for enum query params


jawspeak:
Our swagger API is case sensitive for enums. E.g. "DESC" is ok but "desc" or "deSC" will cause 400s. This is needlessly strict and we should support case insensitive enums. We'd like to allow it to be case insensitive (these are get params). Since the validation code is deep in go-swagger generated code, have you considered this as a go-swagger feature?

thanks!

casualjim:
We could add an extension like: x-strict and that would then make the string comparison use equalfold. 

I do think both options should be supported because I think that typically enums are an exhaustive list of strict values; but at the same time I think you have a valid use case.

fredbi:
Interested by this. However this only applies to string values. Enum is more general than that.
Will do, for models:
-x-go-enum-ci : CI validation on this specific enum (ignored if values not string)
--with-enum-ci : CLI option to apply x-go-enum-ci as a default for the whole model

Future generalization to params remain possible.

jawspeak:
https://stackoverflow.com/questions/60772786/case-insensitive-string-parameter-in-schema-of-openapi

fredbi:
Closed with #2264

PR #2264, opened 2020-04-14
Add support of case-insensitive enums (#303)

pheepi:
Enumerations was case-sensitive only which is usually too strict in many projects. Swagger 2.0 does not support any mechanism to distinguish enum case-sensitivity (or more precisely, case folding). This feature adds support to parameters and definitions. Every type may set case-sensitivity with extension `x-go-enum-ci`. If the options is not set, the previous behavior is kept. Moreover, there is command line option `--with-enum-ci` that switches all enumerations to be case-insensitive by default (for operations only). The feature makes sense for strings only and restricted in `validate.EnumCase()` function.

- Replace validation function `Enum` -> `EnumCase`
- Add `x-go-enum-ci` vendor extension
- Add `--with-enum-ci` command line operation option
- Add test YAML specification for #303
- Add tests of extension in parameters and definitions
- Fix all tests invalidated by the change
- Update go-openapi/validate v0.19.7 -> v0.19.8


pheepi:
> I think you still need to run `go generate ./generator` to update the templates in the binary because I don't see it in the list.
> 
> Other than that this PR looks good to me

Aha, it generates `bindata.go`. I thought it will run automatically, would not be better to remove it from repo since it is generated file? I was confused that `go generate` did not work to me on the first attempt, because I have not installed `go-bindata`. It might be reasonable to add it to `go.mod`. The file changed a lot and I do not know if the shuLhan's version I have used is the right one.

Still it cannot compile now, because it depends on PR go-openapi/validate#125.

casualjim:
I've merged and tagged the other pr.

We use: https://github.com/kevinburke/go-bindata
In the build environment we run this script to get all the tools: https://github.com/go-swagger/go-swagger/blob/master/hack/devtools.sh

Issue #2256, opened 2020-04-07
Improve support for XML in codegen


fredbi:
## Problem statement

#2241 brought some basic support for XML tagging and marshaling in generated code.

This is nice. Here are some improvements identified for this feature:

*    Integration tests of actual (de-)serialization of requests/responses
*    inline nested schema definitions (other gotmpl's)
*    documentation. We should definitely add some, otherwise, this is of no value. 
*    auto-detecting XML in all produce/consume rather than relying on defaults

Documentation could start here:  https://github.com/go-swagger/go-swagger/blob/master/docs/generate/server.md

We might want to split this request for enhancement in several smaller ones, if need be.

## Environment
swagger version: x.x.x  
go version: x.x.x  
OS:

Issue #2243, opened 2020-03-11
Performing basic auth from client with go-swagger 0.22


NiharS:
## Problem statement
I'm trying to use an existing swagger spec to interact with a server API. Right now, I can CURL the endpoint and it'll return properly. The server is set up to accept the user/pass combo of an api key and empty string (i.e. curl -u "${API_KEY}:" ...). I am trying now to use the generated client functions in Go to do the same thing.

Originally I followed this doc: https://github.com/go-swagger/go-swagger/blob/master/docs/generate/client.md only to find that it seems outdated; while `httptransport.BasicAuth` still exists, there's no function within the generated client.Operations to load these parameters (no such .All method, and no "overloaded" methods that accept additional parameters aside from the <operationName>Params). I tried overriding my transport object's DefaultAuthentication to the result of that function, but my server says the auth is invalid.

Instead, I've also tried setting the HttpClient to one with a different roundtripper that wrote the username and password, but at this point my server would no longer even respond to these requests and the context deadline would be exceeded.

Sorry if I'm missing something obvious, but how can I use the client code to perform basic auth? I cannot find any other documentation or examples detailing how to do this, I'm worried the docs in the link above are out of date.

## Environment
swagger version: 2.0
go-swagger version: 0.22.0
go version: 1.13.1
OS:   Darwin

nubenum:
With some trial and error, I got it working for me with Bearer Auth; it should be similar with Basic Auth:
```go
import (
  "os"

  "github.com/go-openapi/strfmt"
  "github.com/go-openapi/runtime"

  apiclient "github.com/myproject/client"
  httptransport "github.com/go-openapi/runtime/client"
)

func main() {
  r := httptransport.New(apiclient.DefaultHost, apiclient.DefaultBasePath, apiclient.DefaultSchemes)
  r.DefaultAuthentication = httptransport.BasicAuth(os.Getenv("API_USER"), os.Getenv("API_PASSWORD"))

  client := apiclient.New(r, strfmt.Default)

  resp, err := client.Operations.MyGreatEndpoint()
  if err != nil {
    log.Fatal(err)
  }
  fmt.Printf("%#v\n", resp.Payload)
}
```
Maybe it helps. If that is the case, we should indeed update the documentation.

sbgreene1307:
Can someone please update the docs on this? Chased my tail with the outdated docs, glad there is an example here at least. Thanks for the example @nubenum

bharatmicrosystems:
[...]

PR #2265, opened 2020-04-15
Improvements for XML support

nubenum:
For starters, at least some documentation, see #2256.

I decided to primarily put it into the client.md, because I had an example ready. And who would voluntarily _provide_ an XML API anyways ;)

Maybe the generate client documentation needs an update in general, also see #2243. For me it was trial and error as well, maybe someone can shed some light on that.


movergan:
OMG! That's what I was looking for!

PR #2282, opened 2020-04-28
Fix multiline descriptions for the stratoscale template

JoakimSoderberg:
Before this change, if a description contained multiple lines, the generated code
would be invalid, since only the first row would be marked as a comment.

Changed the comment to use a multiline `/* */` comment instead.

Also made it so that the `summary` is checked and used preferably.
Since in the Swagger 2.0 it's recommended to be less than 120 characters.

Rather than the description which might be quite long.


JoakimSoderberg:
@casualjim the **GolangCI** is not getting started. Not sure what to do? Are you waiting for that before merge?

PR #2312, opened 2020-06-12
feature: strict responders option

onokonem:


examples/todo-list-strict/restapi/operations/todos/update_one_responses.go
```diff
@@ -0,0 +1,140 @@
...
+	return &UpdateOneNotImplementedResponder{
+		middleware.NotImplemented(
+			"operation authentication.UpdateOne has not yet been implemented",
+		),
+	}
+}
+
+type UpdateOneResponder interface {
+	middleware.Responder
+	UpdateOneResponder()
```
casualjim, line 139: What is the intent with this? Create a sum type?

onokonem: the main idea is to disallow to return a wrong responder from a handler.

we have special types generated by go-swagger intended to be used to return something from the method, so let's use them

casualjim: Ok in that case it might make sense to use a more unique name and perhaps unexport it (for example prefix it with the package name or snakeized full package path).

Right now if something has the signature UpdateOneResponder as a method it can be used as a responder without any guarantee that it is actually one of the generated ones. 

One more suggestion would be to have some helper to wrap a middleware responder with the updateoneresponder so that people can still provide their own but intentionally

casualjim:
I approved but I want to give @fredbi a chance to add his comments, so i'll keep it open for another day or so.

onokonem:
> I approved but I want to give @fredbi a chance to add his comments, so i'll keep it open for another day or so.

no rush :)

It took me more than a year to finally prepare it, so one more day does not matter I guess

examples/todo-list-strict/restapi/operations/todos/update_one_responses.go
onokonem: > Ok in that case it might make sense to use a more unique name and perhaps unexport it (for example prefix it with the package name or snakeized full package path).

It is unique already, isn' it? I did introduce these types, they were there already. I've just changed the Handler code to accept the limited set of types only as of the responders.

> Right now if something has the signature UpdateOneResponder as a method it can be used as a responder without any guarantee that it is actually one of the generated ones.

I do not think it could happen accidentally, do you? And we are not trying to stop the programmer from doing something intentionally, right?

> One more suggestion would be to have some helper to wrap a middleware responder with the updateoneresponder so that people can still provide their own but intentionally

2 things crossing my mind:

1. We have all the methods returning anything described in the spec and so the client can/will expect to receive. The main idea behind this patch is to stop my colleagues from returning custom answers. It was easier to return something custom rather the generated one, it is not any more so problem solved.
2. It is still easy to create a custom time conforms the requirements, but not too easy

fredbi:
@onokonem please rebase and regenerate the bindata assets.

onokonem:
I've rebased my branch from upstream master

PR #2313, opened 2020-06-13
update to errors that carry the value

casualjim:


v-antech:
This commit does not work with the latest errors dependency, as `errors.Required()` accepts 2 and not 3 arguments.

casualjim:
update your dependencies. The change is made across the board

PR #2326, opened 2020-06-29
re-enacted CI-based linting

### LINKED ISSUE #2257: CI: Remove golangci dependency
## Problem statement
As per Denis' announcement on March 14th, golangci is ceasing its operation as a CI provider on April 15th. The golangci-lint tool remains available as open source.

We need to:
* [ ] remove the golangci hook in github
* [ ] add a lint step in CI, running golangci-lint installed locally

## Environment
swagger version: x.x.x  
go version: x.x.x  
OS:

fredbi:
* fixes #2257

**NOTE**:
golangci hosted service has closed. We have to run it from CI.
I haven't tried github's super-linter yet: just wanted to lint our golang code again and not to
re-lint the entire repo with Dockerfiles, sh scripts etc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,7 +137,7 @@ jobs:
             - go-mod-v1-{{ checksum "go.sum" }}
       - run:
           name: Install golangci-lint
-          command: go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.24.0
+          command: go get github.com/golangci/golangci-lint/cmd/golangci-lint@latest
       - run:
           name: Run unit tests
           command: golangci-lint run --timeout 2m --new-from-rev master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,10 +67,12 @@ workflows:
 
   build_and_test:
     jobs:
+      - lint
       - build
 
       - test:
           requires:
+            - lint
             - build
           filters:
             tags:
@@ -78,6 +80,7 @@ workflows:
 
       - codegen_tests:
           requires:
+            - lint
             - build
           filters:
             tags:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -123,6 +127,22 @@ workflows:
...
+    docker:
+      - image: goswagger/builder:debian
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - go-mod-v1-{{ checksum "go.sum" }}
+      - run:
+          name: Install golangci-lint
+          command: go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.24.0
```
casualjim, line 140: can you make the tag: `@latest`

fredbi: I have pinned this on purpose. Lately, on some other repos of mine, more recent versions were broken.
Let me push some more commits downgrading from latest up to the most recent version that works.

casualjim: if we don't specify `-u` and use `GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@latest`
There should be no issue because it will just resolve to the highest tag version.

fredbi:
@casualjim you were right, it works with latest. Not so long ago, I have had broken linter... weird.

.circleci/config.yml
fredbi: With go1.14, "on" should be the default. Is there still a need to enforce GO111MODULE="on" nowadays?
Thinking about the impact of "-u"...

casualjim: correct on the env var.
the impact of `-u` is that it ignores the content of the go.mod file of the tagged version and updates the dependencies of that module. So that's when things break because they are not using the correct versions.

fredbi:
@casualjim  I think we are good to go with this one. I am going to merge it.

PR #2348, opened 2020-07-10
Improve client response error message when status code not recognized

minitauros:
I was debugging a failing test and I saw this error message and thought "huh, did I not set up my mocks correctly? what's going on here?". It actually took me a while before checking the generated swagger code and realizing that all I did wrong is have my mock return the wrong response status code.


casualjim:
can you run `go generate ./generator` otherwise the changes to the template won't get picked up

minitauros:
My bad, must have missed some instructions.

Ran it and added it as a new commit.

casualjim:
Github won't let me merge this because of merge conflicts in bindata.go, could you pull master first?

minitauros:
Pulled master, rebased, force pushed.

Apologies for the force push since you can't see the diff. Not sure what kind of reviewing strategy you prefer.

casualjim:
That's fine, we only care about the final changeset. Thanks!

PR #2367, opened 2020-07-28
Fixed diff writing to file called stdout instead of os.StdOut

wjase:


wjase:
Not sure why DCO check is failing - think i have signed commits??

casualjim:
I disabled DCO, but the build is also failing. 

There are 2 sign things for git. One is -S for a gpg signature and then there is -s which adds your email as the person who signed off on it. Anyway I disabled so you can just ignore that one. see here: https://github.com/go-swagger/go-swagger/pull/2367/checks?check_run_id=920673580

wjase:
there seems to be a slack auth issue unrelated to my change

->Error sending Slack notification: Error sending Slack message: invalid_token

Issue #936, opened 2017-02-17
Add validation for a 'readOnly' property


posener:
## Problem statement

Sending a request with property which is defined as 'readOnly' in the schema should return a 400.
Currently it is accepted.

## Swagger specification

readOnly: boolean:

Relevant only for Schema "properties" definitions. Declares the property as "read only". This means that it MAY be sent as part of a response but MUST NOT be sent as part of the request. Properties marked as readOnly being true SHOULD NOT be in the required list of the defined schema. Default value is false.

## Steps to reproduce

1. Create a POST verb on path / with body with schema that has a readOnly property
2. Send a request that contains this property.
3. You won't get an 400 error

## Environment

swagger version: 0.8.0
go version: go1.7.4
OS: linux/amd64

dbaumgarten:
+1
I also would really like to have this feature. An api I am building has some properties in its models like createdAt or modifiedAt, that are declared as readOnly. Currently I have to "manually" remove these fields from requests, what is quite annoying, and kind of negates the whole purpose of auto-generating code from swagger.

I tried to find the relevant places, where this had to be implemented, but I failed. Although I think I am relatively proficient in go, I have absolutely no idea, what exactly is going on in the go-swagger code.

fredbi:
With go-openapi/validate#50 the spec validate will issue a warning on this situation.
I understand this doesn't fully address the issue, but it's at least a start...

fredbi:
This one is difficult because it makes the validation _contextual_: an instance of a model (or simple schema) may be valid or not depending on who sent this data.

To my knowledge, this is now the **ONLY** swagger 2.0 feature that is not supported by go-swagger.

For simple params and headers, we might find a straightforward way to generate this additional validation, since they only exist in the context of an operation and (at the moment) we never validate responses or response headers. This validation would return an error is the value is not IsZero().

For models representing schemas, the situation is more difficult: the `runtime.Validatable` interface says if a model is valid or invalid, full stop. It is not contextual.

One idea would be to extend the validation capabilities with an additional interface (say, `runtime.ValidatableParameter`, or `ValidatableFor`...) which would carry on the extra check in this specific context (JSON-Schema draft 7 introduces the concept of "owning authority" to describe the newly introduced `readOnly` and `writeOnly`). Or add methods to set a validation context.

I think this one should be discussed to figure out a sound design.

gregmarr:
[...]

PR #2369, opened 2020-08-02
Feature/ReadOnly-Validation Support

youyuanwu:
Support swagger 2.0 readonly validate.

This is WIP. To complete this feature, all `Validate(formats strfmt.Registry)` signature needs to be changed to `Validate(ctx context.Context, formats strfmt.Registry)`, and some interface definitions of this signature needs to be changed. Maintainers please comment on if this switch is acceptable, and on backward compatibility.

Feature Logic:
When validate signature is changed to `Validate(ctx context.Context, formats strfmt.Registry)`, ctx can be used to pass in value of operation type Request or Response. and validate.ReadOnly(ctx,formats) can return true or false based on the context value.
Without ctx, there was no way to do this, and this is why this feature is missing in go-swagger for so long.

TODOs:
- Integrate with go-openapi validate pkg (WIP in https://github.com/go-openapi/validate/pull/127)
- write tests
- Figure out validate interface signature and backward compatibility
- Implement ReadOnly check for special cases (resolve all WIP comments in code)

Reviewers:
Essential changes are in the gotmpl files. Tests changed by using string replacement to update to new validate signature. 

Future:
The new Validate signature paves way for additional features in swagger3.0 like WriteOnly.


youyuanwu:
@fredbi  We should discuss in https://github.com/go-swagger/go-swagger/issues/936 .

youyuanwu:
Updated PR:
* ContextValidate(ctx,format) is generated for each object (we can optimize this), and is independent of previous validate(format).
* So for each model, Validate function does what it did before. ContextValidate validates the model based on the context the model is used. i.e. if it is in a request or response.

Difficulty: 
* When model1 has model2 as a ref property, I could not find a way in the generator code do pass the gen-context from model2 to model1. If model2 has a readonly field, model1 needs the gen-context to determine if it needs a contextvalidate function generated. For now I implemented all models has contextvalidate generated, (most of them are empty methods since readonly is rare).  

TODO (I will do it if this approach is approved by mainainer):
* Fix tests by implementing context validate for additional properties etc.

casualjim:
I think this approach looks good, keep up the good work :+1:

youyuanwu:
All Tests are fixed. The codegen_tests failed because compiler is flagging variable not used, which is caused by the validate.ReadOnly is still commented.

If https://github.com/go-openapi/validate/pull/127 is merged, I will uncomment genreated readOnly validation, and this PR is ready for merge.

jerome-laforge:
Still it possible to change  `ContextValidate` by `ValidateWithContext`.
If I take the example of `http.NewRequest(method, url string, body io.Reader) (*Request, error)` has the same function with context named `http.NewRequestWithContext(ctx context.Context, method, url string, body io.Reader) (*Request, error)`

youyuanwu:
> Still it possible to change `ContextValidate` by `ValidateWithContext`.
> If I take the example of `http.NewRequest(method, url string, body io.Reader) (*Request, error)` has the same function with context named `http.NewRequestWithContext(ctx context.Context, method, url string, body io.Reader) (*Request, error)`

The original name before `ContextValidate` was `ValidateContext`. The internal new variable created in model.go is `HasContextValidation` and is a mirror of `HasValidation`. `ContextValidate` seems clear enough.

generator/templates/schemaembedded.gotmpl
```diff
@@ -10,4 +10,13 @@ func ({{.ReceiverName }} {{ if or .IsTuple .IsComplexObject }}*{{ end }}{{ if .D
   }
   return nil
 }
+
+func ({{.ReceiverName }} {{ if or .IsTuple .IsComplexObject }}*{{ end }}{{ if .Discriminates }}{{ camelize .Name }}{{ else if .IsExported }}{{ pascalize .Name }}{{ else }}{{ .Name }}{{ end }}) ContextValidate(ctx context.Context, formats strfmt.Registry) error {
+  var f interface{} = {{ .ReceiverName }}.{{ dropPackage .ElemType.GoType }}
+  // WIP: change to context validatable
+  if v, ok := f.(runtime.Validatable) ; ok {
```
youyuanwu, line 17: We probably need to ContextValidatable signature in runtime pkg.

generator/templates/schemapolymorphic.gotmpl
```diff
@@ -8,6 +8,8 @@
         the definition of the base type has a Validate() func.
     */}}
         runtime.Validatable
+        // WIP: add to interface in runtime pkg
+        ContextValidate(context.Context, strfmt.Registry) error
```
youyuanwu, line 12: This should go in runtime pkg

casualjim: Yeah you're right it can go in there. Most of them are here: https://github.com/go-openapi/runtime/blob/master/interfaces.go

youyuanwu: https://github.com/go-openapi/runtime/pull/185 added PR here.

youyuanwu: fixed.

generator/templates/schemaembedded.gotmpl
youyuanwu: fixed

youyuanwu:
@casualjim We need a proper tag for runtime pkg.
Let me know if there is more that I need to fix in this PR. I think it is ready to merge.

casualjim (approved):

Issue #2309, opened 2020-06-10
No Validate method for server parameter structs


mwieser:
## Problem statement

When generating models with `swagger generate model` or `swagger generate server` the resulting models from the swagger definitions all implement a validation method like `Validate(formats strfmt.Registry) error`. However, the generated models for parameters do not include this functionality.

Validations for the parameter are currently generated but only used while binding the param struct. In order to use the validations standalone, without having to bind them with the generated method (for example when binding with a different framework), I propose to add a `Validate(formats strfmt.Registry) error` method to https://github.com/go-swagger/go-swagger/blob/548768a026b86c7f9836e1c0dc76cbdce3718767/generator/templates/server/parameter.gotmpl#L489 to be able to use the parameter validation separately.

Is there any interest in adding this functionality?

## Environment
swagger version: 0.23.0  
go version: 1.14.2  
OS: Debian

fredbi:
The `runtime.Validatable` interface was primarily defined for genuine for swagger schemas.
That is true that the validation of `swagger simple types` is a pain in the generated code.
This is mostly because the unmarshaling of these parameters is more complex than just reading from a json body.
Refactoring the `parameter.gotmpl` to have some reusable part (could be used, for example, to validate params on the client before sending a request) is a good idea. Feel free to post a PR along these lines.

mwieser:
I am currently working on a PR for this and I was wondering if the creation of a new array/map is necessary in the validation methods. At the moment when the validation takes place a new array/map is build and then reassigned to the property. Can this be omitted? 


```
// validateSimpleArrayWithValidationBody validates an inline body parameter
func (o *GetSimpleArrayWithValidationParams) validateSimpleArrayWithValidationBody(formats strfmt.Registry) error {
	simpleArrayWithValidationIC := o.SimpleArrayWithValidation

	var simpleArrayWithValidationIR []int64
	for i, simpleArrayWithValidationIV := range simpleArrayWithValidationIC {
		simpleArrayWithValidationI := simpleArrayWithValidationIV

		if err := validate.MaximumInt(fmt.Sprintf("%s.%v", "simpleArrayWithValidation", i), "body", int64(simpleArrayWithValidationI), 12, false); err != nil {
			return err
		}

		simpleArrayWithValidationIR = append(simpleArrayWithValidationIR, simpleArrayWithValidationI)
	}

	o.SimpleArrayWithValidation = simpleArrayWithValidationIR
	return nil
}
```

[...]

PR #2387, opened 2020-08-21
Feature runtime.Validateable for for parameters

mwieser:
Regarding https://github.com/go-swagger/go-swagger/issues/2309. Add a Validate method to the parameter generation, which reuses the validate methods of the properties if present. Refactored the `parameter.gotml` to have reusable code for binding and validation code generation. Added the `GenParameterValidateOnly` and `GenItemsValidateOnly` structs for this to pass a `GenParameter/GenItems` to a template to only generate the code necessary for validation.

This removes reassigning array/slice properties in the validation methods. As mentioned in https://github.com/go-swagger/go-swagger/issues/2309, it seems that the reassignment of the property is not necessary in the validation methods and I therefore removed it to remove side effects from the validation methods. In case the binding in the validate methods is needed, the step would need to be added again.

The reason for this PR is to validate parameter structs without having to bind them. This can be helpful when validating the params on a client before sending them. Furthermore the method exists for models so it would be more unified.


casualjim:
Hi, 

I'm sorry I missed this PR. Would it be possible to bring it up to date so I can merge it?

PR #2389, opened 2020-08-24
Pass examples to the generated models

SVilgelm:
In our workflow we are using the `swagger.yaml` as a source to generate the models then we generate the `swagger.json` to use in some other services.
There is a problem, we have some examples in the `swagger.yaml`, but the generated models don't have those examples, and the generated `swagger.json` from the code also don't have those examples.
This PR fixes the problem.


SVilgelm:
@fredbi @casualjim please review this

Issue #1480, opened 2018-04-16
Missing strfmt import


adamflott:
## Problem statement

Generated responses code has a missing import statement for 

       strfmt "github.com/go-openapi/strfmt"

When used in a type like:

        XRateLimitReset strfmt.DateTime `json:"X-Rate-Limit-Reset"`

## Swagger specification

    swagger: "2.0"

    consumes:
    - application/json
    produces:
    - application/json

    info:
      title: importbug
      description: ...
      version: 0.0.1

    schemes:
    - http

    parameters:
      requestId:
        name: X-Request-Id
        description: A unique UUID for the request
        in: header
        type: string
        required: false
        minLength: 1
      urlPurgeBody:
        name: body
        in: body
        schema:
          $ref: "#/definitions/urlPurgeData"
      tagPurgeBody:
        name: body
        in: body
        schema:
          $ref: "#/definitions/tagPurgeData"
      cpcodePurgeBody:
        name: body
        in: body
        schema:
          $ref: "#/definitions/cpcodePurgeData"
      purgeType:
        name: purgetype
        description: ...
        in: path
        type: string
        required: true
        minLength: 1
        enum:
          - delete
          - invalidate
      purgeClient:
        name: purgeclient
        description: ...
        in: path
        type: string
        required: true
        minLength: 1
        enum:
          - a
          - b
          - c
          - d
      purgeEnvironment:
        name: purgeenvironment
        description: Enumerate supported Purge environments
        in: path
        type: string
        required: true
        minLength: 1
        enum:
          - staging
          - production

    paths:
      /{purgeclient}/v3/{purgetype}/url:
        post:
          tags:
            - Purge
          operationId: v3Url
          parameters:
            - $ref: "#/parameters/requestId"
            - $ref: "#/parameters/purgeType"
            - $ref: "#/parameters/purgeClient"
            - $ref: "#/parameters/urlPurgeBody"
          responses:
            201:
              $ref: '#/responses/Standard201Response'
            403:
              $ref: '#/responses/Standard403Response'
            429:
              $ref: '#/responses/Standard429Response'
            507:
              $ref: '#/responses/Standard507Response'
            default:
              $ref: '#/responses/StandardErrorResponse'

      /{purgeclient}/v3/{purgetype}/url/{purgeenvironment}:
        post:
          tags:
            - Purge
          operationId: v3UrlWithEnv
          parameters:
            - $ref: "#/parameters/requestId"
[...]

PR #2411, opened 2020-10-06
Fix generation of server responses when response has date-time format

bvwells:
Signed-off-by: Ben Wells <b.v.wells@gmail.com>

Fix generation of server responses when response has date-time format.


bvwells:
I believe that this PR fixes #1480. Server generation fails a response contains a date-time format. The example here illustrates the problem https://github.com/bvwells/go-swagger-issue.

Hope this is useful.

bvwells:
Grrr, looks like this doesn't help after all. :-(

PR #2427, opened 2020-11-16
updated go-openapi dependencies

fredbi:
Update is broken (go-openapi/analysis broken after breaking change in go-openapi/spec@v0.19.9)

This PR integrates the fix (dry run with analysis replaced on wip branch)

The CI remains broken on windows (see https://github.com/go-swagger/go-swagger/issues/2428). I am working on that one this week.


eimlav:
Hi @fredbi . Is the addition of the empty `description` field in the responses valid as part of the specification or is this a bug introduced by updating the dependencies? Running `swagger generate` now adds this field to responses in our swagger.json as you have in the fixtures. However, we are seeing `swagger validate` fail due to this:

```
- "paths./api/test.post.responses.200" must validate one and only one schema (oneOf). Found none valid
- paths./api/test.post.responses.200.$ref in body is a forbidden property
...
```

I could not find anything in the swagger spec docs that indicates it is valid.

fredbi:
@eimlav yes, this field (which is required by the spec) is now no more `omitempty`-ied
This comes from an update in go-openapi/spec
A crap, we have this validation issue now. Could you please file in an issue about that. I'll try to figure out a fix.

eimlav:
Sure @fredbi, I've created an issue https://github.com/go-swagger/go-swagger/issues/2429 . Thanks for your help! :)

PR #2438, opened 2020-11-22
Added more control over external types

### LINKED ISSUE #2385: x-go-type for a custom type in models/
## Problem statement

Before #2340 I was able to declare a type in my spec that was already present in `models/` (= not auto-generated). Here `MyType` used in a `Zzz` object type:

```
type Zzz struct {
        // meta
        Meta MyType `json:"meta,omitempty"`
}
```

Now it generates bad go code as the "import.package" is empty (then gofmt glues field name and the type together):

```
type Zzz struct {
        // meta
        Meta.MyType `json:"meta,omitempty"`
}
```

Is there a way to overcome this behavior? (excluding the move of custom types outside models/, of course)

## Swagger specification

```
---
swagger: "2.0"
info:
  version: "1.0"
  title: "Test API"
paths:
  /test:
    get:
      responses:
        default:
          description: xxx
          schema:
            $ref: "#/definitions/Zzz"

definitions:
  Zzz:
    type: object
    properties:
      meta:
        $ref: '#/definitions/MyType'

  MyType:
    type: object
    x-go-type:
      type: MyType
```

## Steps to reproduce

Using v0.25.0:
```
swagger generate server -f api.yaml
```

## Environment
swagger version: 0.25.0  
go version: 1.15
OS:   Linux x86_64

fredbi:
This PR brings more granular support to use external types for every schema: in definitions,
properties, anonymous parameter or response schemas.

We can generate collections of external types such as objects, slices, maps or tuples.

> NOTE:
> * embedded external types must still be defined at the top level (definitions).
> * simple types are not supported (query param, header, etc)

* fixes #2385 (defaults to package location for generated models)

* added support for inlined external type declaration, not only for top level definitions
* added support for "novalidation" hint to prevent external types to call validations (by default, they do)
* fixed some nullability control issues over external types
* added package alias deconfliction for external types (when the alias is not specified explicitly)

* added full example with comments and generated code
* added markdown doc

* fixed broken codegen with type binary (io.Reader): was breaking on the ContextValidate methods. This was breaking the new tests.


fredbi:
@maxatome  this should address issue #2385  you reported some time ago, and then some.

PR #2458, opened 2020-12-14
ensure custom tag is applied to object in generated model (#2457)

### LINKED ISSUE #2457: custom tags are not added to instances of objects when specified
## Problem statement

I'm trying to use the `x-go-custom-tag` extension to add a tag to an object in my schema, but when a model is generated that has an instance of that object in it, the instance does not have any custom tags applied. 

When applying the `x-go-custom-tag` as seen in the sample spec below, this is the code generated:
```
...

// ObjWithCustomTag obj with custom tag
//
// swagger:model ObjWithCustomTag
type ObjWithCustomTag struct {

    // custom tag
    CustomTag *ObjWithCustomTagCustomTag `json:"customTag,omitempty"`

}

...
```

### Expected Result
```
...

// ObjWithCustomTag obj with custom tag
//
// swagger:model ObjWithCustomTag
type ObjWithCustomTag struct {

    // custom tag
    CustomTag *ObjWithCustomTagCustomTag `json:"customTag,omitempty" myobjtag:"foo,bar"`

}

...
```

## Swagger specification

```
ObjWithCustomTag:
  type: object
  properties:
    customTag:
      type: object
      x-go-custom-tag: myobjtag:"foo,bar"
      properties: 
        something:
          type: string
```

## Steps to reproduce

`swagger generate model -M ObjWithCustomTag -f spec.yml -t src/generated/`

## Environment
swagger version: 0.25.0
go version: 1.15.6
OS: macOS 11.0.1

bobcallaway:
* object properties now have custom tags appended to generated models
* fixes #2457 

Signed-off-by: Bob Callaway <bcallawa@redhat.com>


generator/types.go
```diff
@@ -934,6 +934,7 @@ func (t *typeResolver) ResolveSchema(schema *spec.Schema, isAnonymous, isRequire
 			break
 		}
 		result.HasDiscriminator = schema.Discriminator != ""
+		result.Extensions = schema.Extensions
```
fredbi, line 937: This is probably tampering with a lot of other extensions. Have you considered adding the x-go-tag extension in the setExtension func (called in a defer statement, so it is not overwritten by ref resolution)?

generator/model.go
```diff
@@ -804,6 +804,11 @@ func (sg *schemaGenContext) buildProperties() error {
 			return err
 		}
 
+		// set custom serializer tag
```
fredbi, line 807: I fail to see why the statement doing the same thing at the end of the loop is not enough. My hunch is that only the type resolver is failing and that piece wouldn't be necessary if the extension was properly resolved for the property (as you have intended to do in types.go)

generator/types.go
fredbi: Based on your original idea (which is good) I have a tried a version that pass the unit tests and seems to do the job, at least on a basic test case.

```go
+++ b/generator/types.go
@@ -1244,6 +1244,13 @@ func (rt *resolvedType) setExtensions(schema *spec.Schema, origType string) {
        rt.IsEnumCI = hasEnumCI(schema.Extensions)
        rt.setIsEmptyOmitted(schema, origType)
        rt.setIsJSONString(schema, origType)
+
+       if customTag, found := schema.Extensions[xGoCustomTag]; found {
+               if rt.Extensions == nil {
+                       rt.Extensions = make(spec.Extensions)
+               }
+               rt.Extensions[xGoCustomTag] = customTag
+       }
 }
 
 func (rt *resolvedType) setIsEmptyOmitted(schema *spec.Schema, tpe string) {
```

```go
+++ b/generator/model.go
@@ -917,7 +917,7 @@ func (sg *schemaGenContext) buildProperties() error {
                emprop.GenSchema.Extensions = emprop.Schema.Extensions
 
                // set custom serializer tag
-               if customTag, found := emprop.Schema.Extensions[xGoCustomTag]; found {
+               if customTag, found := tpe.Extensions[xGoCustomTag]; found {
                        emprop.GenSchema.CustomTag = customTag.(string)
                }
                sg.GenSchema.Properties = append(sg.GenSchema.Properties, emprop.GenSchema)
@@ -1649,7 +1649,7 @@ func (sg *schemaGenContext) buildAdditionalItems() error {
 }
 
```

fredbi: We have to be more surgical regarding bubbling up extensions, because of the x-nullable.

PR #2464, opened 2020-12-26
Ensure enum variant name is unique

mraerino:
Sometimes enum variants have special characters that are significant to distinguish them.
Example: `2.4Ghz` vs `24Ghz` which might be valid enum variants for an API dealing with wifi frequencies.

This attempts to avoid const name clashes when generating those enum variants.

Let me know if you'd like a different approach for this. The main goal here was to be as least disruptive as possible.
That's why i decided to not change `pascalize` to do this but instead introduced a new function that is only called for enum variant names.


fredbi (changes requested):
LGTM. Could you please add some tests?

mraerino:
This looks like it's passing now and has 100% coverage on my diff.

It likely is a breaking change as you can see on the changes to existing tests. How should that be handled?

fredbi:
@mraerino the breaking change is IMHO acceptable. Merging this. Thank you

fredbi (approved):

PR #2475, opened 2021-01-07
add specific version install info

### LINKED ISSUE #2474: Missing info on how to build specific version from source
## Problem statement
Just a nit. There's no info in docs how to build a specific version from source and why tagged build shows as `dev` version.

## Steps to reproduce
```
dir=$(mktemp -d)
git clone https://github.com/go-swagger/go-swagger "$dir" 
cd "$dir"
go checkout v0.25.0
go install ./cmd/swagger

swagger version
version: dev
```

## Environment
swagger version: 0.25.0  
go version: 1.15.6
OS: Linux

blami:
Add information how to install specific version from source and how to set `-ldflags` so swagger version reports that version and commit id instead of just `dev`.

Fixes #2474

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -88,6 +88,16 @@ cd "$dir"
 go install ./cmd/swagger
 ```
 
+To install a specific version from source an appropriate tag needs to be checked out first (e.g. `v0.25.0`). Additional `-ldflags` are just to make `swagger version` command print the version and commit id instead of `dev`.
+
+```
+dir=$(mktemp -d)
+git clone https://github.com/go-swagger/go-swagger "$dir" 
+cd "$dir"
+go checkout v0.25.0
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -88,6 +88,16 @@ cd "$dir"
 go install ./cmd/swagger
 ```
 
+To install a specific version from source an appropriate tag needs to be checked out first (e.g. `v0.25.0`). Additional `-ldfalgs` are just to make `swagger version` command print the version and commit id instead of `dev`.
```
gregmarr, line 91: typo: ldfalgs

blami: thanks for catching it. fixed.

gregmarr: You're welcome.

fredbi (approved):

fredbi:
Thanks @gregmarr for the review. Your help is very much appreciated!

@blami I think this should also work now with `go get github.com/go-swagger/go-swagger/cmd/swagger@v0.25.0`

PR #2500, opened 2021-01-25
WIP: add client options.

### LINKED ISSUE #1924: Can't configure content type in generated client
## Problem statement

The generated client only allows you to set `params` and there isn't a way to override the content type used if the endpoint accepts multiple content types. The runtime library itself just uses the first content type in the list of `ConsumesMediaTypes` https://github.com/go-openapi/runtime/blob/3d6572c5a80826d5e7f6362d4c681875207c3ef6/client/runtime.go#L333-L341.

## Suggestion

Instead of hardcoding the `ClientOperation` https://github.com/go-swagger/go-swagger/blob/312366608bbf17dd219190b66ab63bdc8b4d0db0/examples/todo-list/client/todos/todos_client.go#L36-L48 we could have some kind of function that produces the `DefaultAddOneOperation` and in addition to `AddOne` have an `AddOneWithOperation` that takes in a `ClientOperation` as an argument.

That way, a user can create a default operation, modify it, and then pass it in to the client. What do you think?

Or, since it's generated code, the existing `AddOne` can remain unmodified, and we can just generate the additional helper functions.

## Environment
swagger version: 0.18.0

gmidorii:
fixes #1924


gmidorii:
Since the edit of commit failed, we will try again.

Issue #397, opened 2016-03-08
edit but don't overwrite configure_xxx.go


casualjim:
currently keeping configure_xxx.go in sync is a manual exercise which is often times confusing.
It would be better if missing things could be added. 
Deleted operations and such should also be reflected in the configure_xxx.go file

Also track a list of generated files or use a .swagger.go extension so that we can delete all ephemeral files without worrying and we don't leave stale code lying around.

Updates should be ignored though, hopefully they'll fail compilation.

ontucker:
I'm still getting used to how things in go-swagger are organized, but my desire is to have the API implementation live in an entirely different module so that I don't have to edit any of the files that live amongst the generated code.  I am currently modifying configure_xxx.go to just do a one-line call out to a server which exports a .ConfigureAPI method and then implements all the XYZHandlerFuncs itself and hooks them to 'api'.  However, it might be cleaner if the generated API defined an interface that contains all the handlers I need to provide and I could just pass something which implements that into operations.NewXXXAPI from main.

AnsonT:
I started putting operation implementations under restapi/implementations. It would be great if configure_***.go file automatically binds to implementations.xxx methods when models are updated. The generator can add stub implementation if the implementation files do not exist, but leave existing ones in place.

fredbi:
I had not seen this issue and just realized I've made the same on my own custom templates.

This could indeed be added as a featured CLI option (something like `--auto-wire-implementation`) but we should be wary of 2 things:
- the configurable implementation target package should be pre-existing, or the goimport won't work at generation time (this makes yet another `GenOpt` / CLI opt, like `--implementation="mypackage"`)
- the target handler is not arbitrarily and should be named after the operationID

In my own `configure_api.gotmpl`, this gives something like:
```golang
...
  // Registers server shutdown method
  api.ServerShutdown = func() {  }

  // Registers API operations
  {{- range .Operations}}
  api.{{if ne .Package $package}}{{pascalize .Package}}{{end}}{{pascalize .Name}}Handler = {{.Package}}.{{pascalize .Name}}HandlerFunc(implementation.{{pascalize .Name}})

  {{ end }}

  return setupGlobalMiddleware(api.Serve(setupMiddlewares))
...
```

This kind of feature is actually a direct application of using custom templates. Adding it as a feature should be rather straightforward. Feel free to submit a PR on this.

gregmarr:
Yeah, that would work perfectly for my server, which uses the pattern:

[...]

PR #2510, opened 2021-02-01
Auto wire api and backend

youyuanwu:
Solving issue: https://github.com/go-swagger/go-swagger/issues/397

Configure.go can be regenerated, if cli option`--implementation-package` specified, then using another template.

The solution is suggested in the issue above. The implementation package must pre-exist and have certain functions implemented.

This is WIP, any suggestions are welcomed.


generator/templates/server/autoconfigureapi.gotmpl
```diff
@@ -0,0 +1,154 @@
...
+  {{ imports .Imports }}
+)
+
+{{ with .GenOpts }}
+//go:generate swagger generate server --target {{ .TargetPath }} --name {{ .Name }} --spec {{ .SpecPath }}
+{{- if .APIPackage }}{{ if ne .APIPackage "operations" }} --api-package {{ .APIPackage }}{{ end }}{{ end }}
+{{- if .ModelPackage }}{{ if ne .ModelPackage "models" }} --model-package {{ .ModelPackage }}{{ end }}{{ end }}
+{{- if .ServerPackage }}{{ if ne .ServerPackage "restapi"}} --server-package {{ .ServerPackage }}{{ end }}{{ end }}
+{{- if .ClientPackage }}{{ if ne .ClientPackage "client" }} --client-package {{ .ClientPackage }}{{ end }}{{ end }}
+{{- if .ImplementationPackage }} --implementation-package {{ .ImplementationPackage }}{{ end }}
```
youyuanwu, line 31: Missing --auto-configure flag.

youyuanwu: Not needed since I removed the flag and if impl-pkg is non-empty, auto config template is chosen.

generator/templates/server/autoconfigureapi.gotmpl
```diff
@@ -0,0 +1,149 @@
...
+  api.{{ pascalize .Name }}Producer = {{ .Implementation }}
+    {{- else }}
+  api.{{ pascalize .Name }}Producer = runtime.ProducerFunc(func(w io.Writer, data interface{}) error {
+    return Impl.{{ pascalize .Name }}Produce(w, target)
+  })
+    {{- end }}
+  {{- end}}
+  {{ range .SecurityDefinitions }}
+    {{- if .IsBasicAuth }}
+  // Applies when the Authorization header is set with the Basic scheme
```
fredbi, line 83: This is good, although I'd have lower requirements on the API implementation regarding the presence of all technical methods (Auth, ConfigureFlags, etc.).

This is how I would do it. Maybe this is something to consider for a later improvement. Feel free to comment on these ideas.

```go
type Implements{{ API name }}API interface {
  {{ ... all API handlers }} // attention! there is a CLI opt that changes the signature and requires specialized responders instead of the generic middleware.Responder
}

type Authable{{ API name}} interface {
   ... //  {{ ...}}Auth methods
}

type ConfigurableAPI interface {
... // ConfigureFlags, ConfigureServer, SetupMiddleware, etc...
   CustomConfigure(ImplementsSimpleToDoListAPIAPI)
}

func configureAPI(injected ImplementsSimpleToDoListAPIAPI) http.Handler {
 ...
 if authable, ok := injected.(Authable{{ API name}}) ; ok {
  api.{{ .ID }}Auth = authable.{{ .ID }}Auth
}
... etc you probably got the gist of it

```
This comment is not a blocker: we may proceed like this and improve it with another PR.

fredbi (approved):

generator/templates/server/autoconfigureapi.gotmpl
youyuanwu: I have some pending changes to displaying all the functions required for implementation obj in interfaces. I will push soon.

I thought about using interface assertion, and make this auto_configure_xxx.go easily compile, but the user should catch the unimplemented error in compile time, instead of runtime. If the interface assertion failed (in your authable) above, it is seems more appropriate to panic, instead of assigning a unimplemented runtime error.

youyuanwu: I added interfaces in the auto-configure template. If we think that type assertion on interfaces are preferred, I can add it in another PR.

fredbi: Just leave it as you first intended. Let's see how people react to this.

fredbi:
Thank you  @youyuanwu

Issue #1176, opened 2017-09-18
some codepaths in server generation assume "./src" while other assume "./" 


notasecret:
## Problem statement

Try to generate go server code, and it fails, complaining about paths.

```
Jons-MacBook-Pro-2:xxx-backend jonmclachlan$ alias swagger="docker run --rm -it -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger"
Jons-MacBook-Pro-2:xxx-backend jonmclachlan$ swagger version
version: dev
commit: 8110e1bfee9ca99f346cdad6146dfe05161cb905
Jons-MacBook-Pro-2:xxx-backend jonmclachlan$ ls src/
morningTulips	xxx		xxx-server
Jons-MacBook-Pro-2:xxx-backend jonmclachlan$ echo $GOPATH
/Users/jonmclachlan/xxx/xxx-backend
Jons-MacBook-Pro-2:xxx-backend jonmclachlan$ swagger generate server -f swagger/swagger.yaml -A xxx -t xxx
2017/09/18 01:49:40 building a plan for generation
2017/09/18 01:49:40 lstat /Users/jonmclachlan/xxx/xxx-backend/xxx: no such file or directory
Jons-MacBook-Pro-2:xxx-backend jonmclachlan$ swagger generate server -f swagger/swagger.yaml -A xxx -t src/xxx
2017/09/18 01:49:51 building a plan for generation
2017/09/18 01:49:51 target must reside inside a location in the $GOPATH/src
Jons-MacBook-Pro-2:xxx-backend jonmclachlan$ 
```

This used to work.

## Environment
swagger version: 8110e1bfee9ca99f346cdad6146dfe05161cb905
go version: whatever is in your latest docker img

notasecret:
this is the commit:

commit fb51ecafe32dd2e9ea3cd697d30044fd2ee514a9
Author: Tushar <taggarwal@vmware.com>
Date:   Mon Aug 28 11:36:46 2017 -0700

    Corrected Path generation for Base Imports

EleanorRigby:
@notasecret : Is your target path inside GOPATH? Are there any symlinks that you are using in your path? 
There was a bug due to which all the paths for code generation worked. Now the behavior is corrected.

notasecret:
Is your target path inside GOPATH? 
Yes, as you can verify above.

Are there any symlinks that you are using in your path?
No sym link funny business.

[...]

PR #2512, opened 2021-02-05
Update install.md

ashishtiwari1993:
For step:
```bash
alias swagger='docker run --rm -it  --user $(id -u):$(id -g) -e GOPATH=$HOME/go:/go -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger'
swagger version
```
It gets failed when GOPATH is not set to home folder. It gives below error:

```sh
2021/02/05 12:37:01 lstat /home/ashish/go: no such file or directory
```
Replacing `$HOME/go`  with `$(which go)` so that it can take GOPATH dynamically. 

Solving issue: https://github.com/go-swagger/go-swagger/issues/1176

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -19,7 +19,7 @@ docker pull quay.io/goswagger/swagger
 #### For Mac And Linux users:
 
 ```bash
-alias swagger='docker run --rm -it  --user $(id -u):$(id -g) -e GOPATH=$HOME/go:/go -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger'
+alias swagger='docker run --rm -it  --user $(id -u):$(id -g) -e GOPATH=$(which go):/go -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger'
 swagger version
 ```
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -19,7 +19,7 @@ docker pull quay.io/goswagger/swagger
 #### For Mac And Linux users:
 
 ```bash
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -19,7 +19,7 @@ docker pull quay.io/goswagger/swagger
 #### For Mac And Linux users:
 
 ```bash
-alias swagger='docker run --rm -it  --user $(id -u):$(id -g) -e GOPATH=$HOME/go:/go -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger'
+alias swagger='docker run --rm -it  --user $(id -u):$(id -g) -e GOPATH=$(which go):/go -v $HOME:$HOME -w $(pwd) quay.io/goswagger/swagger'
```
casualjim, line 22: I think it would be better to use `$(go env GOPATH)`

casualjim (approved):

casualjim (changes requested):
The which command will result in a path like /etc/go/bin/go while we want it to be /etc/go

docs/install.md
ashishtiwari1993: Perfect!! Should i am raising new PR with this option.

ashishtiwari1993: Thanks for this suggestion @casualjim . I thought we need `bin` here. Made new commit with suggested changes: https://github.com/go-swagger/go-swagger/commit/ebc939f1d8b31af7c60feeb36bc07a2b40778d86

casualjim (approved):

PR #2516, opened 2021-02-10
add option to use opentracing

zeitlinger:
https://github.com/go-openapi/runtime/releases/tag/v0.19.26
has added support for OpenTracing (I added it)

and now I'm trying to add the support for the generated client.

I got a bit stuck - not sure if I added the option in the right places.


fredbi:
@zeitlinger we don't need to change the generated client. You can already inject a configured runtime client with the generated client.

zeitlinger:
yes, that actually works :smile:

PR #2554, opened 2021-03-27
refactor CLI generate, support datetime, support multi success resp

youyuanwu:
* Create shared CLI templates for model and anonymous model in operation pkg.
* Support multi success response.
* Support datetime format.
* Add generate tests


generator/operation.go
```diff
@@ -1130,6 +1130,13 @@ func (b *codeGenOpBuilder) buildOperationSchema(schemaPath, containerName, schem
 		isInterface := schema.IsInterface
 		hasValidations := schema.HasValidations
 
+		// TODO: remove this and find a better way to get package name for anonymous models
```
youyuanwu, line 1133: @casualjim In generated cli pkg, I need to render operationName.ModelStruct{} for anonymous schema in parameter. I could not find a way to access the operationName in ExtraSchemas of type []GenSchema. This code here is a hack. Is there a easy way to get the operationName?

casualjim: You mean to get the package name right? I think this is the right way actually

casualjim (approved):

generator/operation.go
youyuanwu: Yes. I need to set the package name recursively for all schemas, right now it is only setting the top level schema. For example pkg is not filled for the array-items0 type generated for array items.

PR #2565, opened 2021-04-21
Export enum slice variables in package models to allow access via reflect

kingnidge:
We would find it extremely useful if generated slice variables for package 'models' representing our defined swagger enums could be exported.  Some of our enums are quite large and require iteration for further processing.

Thank you.

### DIFF
--- a/generator/templates/schemavalidator.gotmpl
+++ b/generator/templates/schemavalidator.gotmpl
@@ -771,19 +771,19 @@ const (
     {{ end }}
 
 // for schema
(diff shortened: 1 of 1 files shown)


youyuanwu:
If the private enum array is exported, user can potentially(or unconsciously) modify the enum array and the validation can break. I am not sure if this is the right way to expose enum array.

I recommend adding an exported function returning enum array of concrete type instead of interface{} in models pkg, but that needs more work.

fredbi:
@youyuanwu  agree: this should be exported as a function.

PR #2593, opened 2021-06-29
Support nested body params for CompositeError

### LINKED ISSUE #2594: Support nested body params for CompositeError
## Problem statement

If you create a spec that has a complex nested body parameter.  The validation logic will return `errors.CompositeError` instead of `errors.Validation`.  As a result, the nested error messages don't get adjusted with the nested field name making it impossible to differentiate from errors that are returned from the root object and those returned by the nested object.  

Here's an example of the generated validation logic.

```go
func (m *CreateRequest) validatePlayer(formats strfmt.Registry) error {
	if m.Player != nil {
		if err := m.Player.Validate(formats); err != nil {
			if ve, ok := err.(*errors.Validation); ok {
				return ve.ValidateName("player")
			}
			return err
		}
	}

	return nil
}

func (m *Player) Validate(formats strfmt.Registry) error {
	var res []error

	if err := m.validateLastUpdated(formats); err != nil {
		res = append(res, err)
	}

	if len(res) > 0 {
		return errors.CompositeValidationError(res...)
	}
	return nil
}
``` 

If the parent object has a property with the same name as the child (`LastUpdated` in the example below), it will return an error saying `lastUpdated in body must be of type date-time` or similar for both an error on `CreateRequest.LastUpdated` and `CreateRequest.Player.LastUpdated`.

The generated validation logic should be updated to support CompositeErrors.  There are a few ways to accomplish this:
1. Add some inline logic for calling `ValidateName` on each `errors.Validation` in a `CompositeError`
2. Update `errors.CompositeError` to have an implementation of `ValidateName` and call it.
3. Create a `ValidationNamer` interface that both `CompositeError` and `Validation` implement and then call that here (simplifies the logic but still needs #2 above implemented).

## Swagger specification
```json
{
  "swagger": "2.0",
  "info": {
    "title": "Sample",
    "version": "v1"
  },
  "paths": {},
  "definitions": {
    "CreateRequest": {
      "properties": {
        "lastUpdated": {
          "format": "date-time",
          "type": "string"
        }
        "player": {
          "$ref": "#/definitions/Player"
        }
      },
      "type": "object"
    },
    "Player": {
      "properties": {
        "lastUpdated": {
          "format": "date-time",
          "type": "string"
        }
      },
      "type": "object"
    }
  },
  "x-components": {}
}
```

## Steps to reproduce
1. Create a spec with a nested param that returns a `CompositeError` from its `Validate` method.
2. Generate client code and verify the validation logic for nested properties.

veleek:
Update templates that have nested body param logic to call ValidateName API on CompositeError.
See https://github.com/go-openapi/errors/pull/30 for addition of ValidateName API to CompositeError which needs to be committed prior to this.

Closes #2594

### DIFF
--- a/generator/templates/schemavalidator.gotmpl
+++ b/generator/templates/schemavalidator.gotmpl
@@ -72,6 +72,8 @@
       if err := {{.ValueExpression }}.ContextValidate(ctx, formats); err != nil {
         if ve, ok := err.(*errors.Validation); ok {
           return ve.ValidateName({{ if .Path }}{{ .Path }}{{ else }}""{{ end }})
+        } else if ce, ok := err.(*errors.CompositeError); ok {
+          return ce.ValidateName({{ if .Path }}{{ .Path }}{{ else }}""{{ end }})
         }
         return err
       }
@@ -137,6 +139,8 @@
       if err := {{.ValueExpression }}.Validate(formats); err != nil {
         if ve, ok := err.(*errors.Validation); ok {
           return ve.ValidateName({{ if .Path }}{{ .Path }}{{ else }}""{{ end }})
+        } else if ce, ok := err.(*errors.CompositeError); ok {
+          return ce.ValidateName({{ if .Path }}{{ .Path }}{{ else }}""{{ end }})
         }
         return err
       }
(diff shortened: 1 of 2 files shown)


veleek:
Whoops this PR should have been opened as a draft.

veleek:
Renamed branch according to contribution guidelines, closing this in favor of #2595

Issue #2602, opened 2021-07-10
Error: "source formatting on generated source "autoconfigure" failed" on server generate


mattalexx:
## Problem statement

Happens if StrictResponders and ImplementationPackage are both set for a server generation.

> Error: "source formatting on generated source "autoconfigure" failed"

## Swagger specification

```yaml
swagger: '2.0'
info:
  description: Test for impl package
  version: 0.0.1
  title: Test for impl package
schemes:
  - http
paths:
  /ping:
    get:
      summary: Ping
      description: Ping
      consumes:
        - application/json
      produces:
        - application/json
      responses:
        '200':
          description: successful operation
          schema:
            $ref: '#/definitions/pong'

definitions:
  pong:
    type: object
    properties:
      pong:
        type: string
```

## Steps to reproduce

See the unit test.

Version 0.27.0

youyuanwu:
Should be fixed by above PR.

PR #2603, opened 2021-07-10
Removed bracket that was causeing an error if StrictResponders and Im…

mattalexx:
…plementationPackage were both set for a server generation


mattalexx:
This addresses https://github.com/go-swagger/go-swagger/issues/2602

youyuanwu (approved):
I have seen CI build broken for the same error on multiple PRs, will investigate.

Change looks good.

casualjim (approved):

PR #2605, opened 2021-07-11
Fix go sum error in CI build

youyuanwu:
CI build has been failing. Trying to reproduce.
WIP

go version used in CI is:
`go1.16.5 linux/amd64` 
There is a known bug in this version that results in error like this:
 `missing go.sum entry for module providing package golang.org/x/net/idna`
See golang bug here: https://github.com/golang/go/issues/44129

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,9 @@ commands:
         default: "./dist/swagger-musl"
     steps:
       - run:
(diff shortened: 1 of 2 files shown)


youyuanwu:
Fixed by https://github.com/go-swagger/go-swagger/pull/2606. Fixing go version to 1.15

PR #2659, opened 2021-11-13
codescan: use TypeSpec.Doc over GenDecl.Doc

crazytyper:
This merge requests improves the command `generate spec` as follows:

```go
type (
   // Do not ignore this comment block
  Foo struct {
  }
)
```

```go
// Do not use this comment block for type Foo and Bar
type (
   // Use this comment block for type Foo instead
  Foo struct {
  }
  // Use this comment block for type Bar instead
  Bar struct {
  }
)
```


casualjim (changes requested):
Can you add a test?

crazytyper:
> Can you add a test?

I have modified existing tests to fail without the proposed changes.

casualjim (approved):

PR #2680, opened 2021-12-08
enhancement: Add `.Pointer()` method to enums.

thetorpedodog:
Following the example of protobuf generated code, this change creates
a `.Pointer()` method on enums that creates a new, non-aliased pointer
to an enum value.

This means that you can call:

    enumPtr := SomeEnumConst.Pointer()

to get a pointer to an enum value.

For reference, see:
https://developers.google.com/protocol-buffers/docs/reference/go-generated#enum

---

I'm not sure what the best way to test this is; the way the existing test cases work is fairly opaque to me. Any pointers (heh)?


casualjim:
can you run `go generate ./generator` ?

thetorpedodog:
Done.

casualjim (approved):

thetorpedodog:
Is there anything else I need to do here?

casualjim:
I fixed the build step on master, so if you update then the builds will hopefully pass

PR #2685, opened 2021-12-16
Add support for nested map property names in errors.

### LINKED ISSUE #2684: Support nested body params in map objects.
## Problem statement
Nested properties that generate map objects don't create error messages that describe the full path to that object.  Instead the error message just refers to the property as though it's a top-level property.

Example CreateRequest JSON:
```json
{
  "data": {
    "gameType": {
      "value": "Something longer than 2048chars...",
    },
  }
}
```

Example error:
> value in body should be at most 2048 chars long.

This is related to #2595 which added support for paths for nested non-map properties of objects.

## Swagger specification

```json
  ...
  "components": {
    "schemas": {
      "DataObject": {
        "type": "object",
        "nullable": true,
        "properties": {
          "value": {
            "type": "string",
            "nullable": true,
            "maxLength": 2048,
          },
        },
        "additionalProperties": false
      },
      "CreateRequest": {
        "type": "object",
        "properties": {
          "data": {
            "type": "object",
            "additionalProperties": {
              "$ref": "#/components/schemas/DataObject"
            },
            "nullable": true,
          }
        },
        ...
  }
```

## Steps to reproduce
1. Create a spec with a map param that returns `CompositeError` from its `Validate` method.
2. Generate client code and verify the validation logic for nested properties returns errors

veleek:
For the following JSON where the `value` property is limited to 2048 chars:

```json
{
  "data": {
    "gameType": {
      "value": "Something longer than 2048chars...",
    },
  }
}
```

Example error before change:
> value in body should be at most 2048 chars long.

Example error after change:
> data.gameType.value in body should be at most 2048 chars long.

Fixes #2684


veleek:
I have no idea why the circleci build is failing... it seems unrelated to this change.

veleek:
FYI @casualjim, this is essentially building on #2684.

casualjim (approved):

veleek:
I believe I forgot to verify the tests last time around.  I'll try to take a look at them tomorrow.

casualjim:
I fixed the build on master, if you update your base branch the tests should be passing

PR #2708, opened 2022-02-05
add an interface to deal with client responses based on the status code

### LINKED ISSUE #2706: Call for transparent HTTP status codes in generated struct
## Problem statement

Seeking guidance.  Say various responses are modelled by the swagger spec (see below) and there are many such endpoints and all I want do to is to categorise them by server errors or client errors.  The HTTP spec makes this easy: server errors have status codes 500-599; and client errors have status codes 400-499.  The generated response structs (e.g. GetBadRequest, GetTooManyRequests, GetInternalServerError, etc.) hide or obfuscate the HTTP status code and I am reduced to using a priori knowledge to compute the distinction.  E.g.

```
func IsClientError(err error) bool {
	switch err := p.cause.(type) {
	case
		*operations.GetBadRequest,
		*operations.GetTooManyRequests,
		*operations.PostBadRequest,
		*operations.PostTooManyRequests,
		*operations.EtcBadRequest,
		*operations.EtcTooManyRequests:
		return true
	case *runtime.APIError:
		return err.Code >= 400 && err.Code < 500
	}
	return false
}
```

I'm aware that [APIError](https://pkg.go.dev/github.com/go-openapi/runtime#APIError) has a Code field.  But what I'm seeking is a common interface across all HTTP error values that might be produced by generated client operation interfaces.  E.g.

```
type HTTPResponse interface {
	StatusMessage() string
	StatusCode()    int
}
```

I realise this may be a can of worms... but it's very difficult to maintain functions such as the above IsClientError.  I wasn't able to find anything that converts, say, *operations.GetBadRequest to *runtime.APIError.  I do see that a const is generated:

```
// GetBadRequestCode is the HTTP code returned for type GetBadRequest
const GetBadRequestCode int = 400
```

But I don't see how that improves the situation.

Is this a common scenario?  Am I missing a trick?

## Swagger specification

```
paths:
  /:
    get:
      operationId: Get
      responses:
        '200':
          description: OK
        '400':
          description: Bad Request
        '429':
          description: Too Many Requests
        '500':
          description: Internal Server Error
```

## Environment
swagger version: v0.25.0
go version: 1.17.5

casualjim:
Adds the same method sets to runtime.APIError and the generated client responses. There is also a convenience interface `runtime.ClientResponseStatus` to runtime 

```go

// IsSuccess returns true when this elapse o k response returns a 2xx status code
func (o *APIError) IsSuccess() bool {
	return o.Code/100 == 2
}

// IsRedirect returns true when this elapse o k response returns a 3xx status code
func (o *APIError) IsRedirect() bool {
	return o.Code/100 == 3
}

// IsClientError returns true when this elapse o k response returns a 4xx status code
func (o *APIError) IsClientError() bool {
	return o.Code/100 == 4
}

// IsServerError returns true when this elapse o k response returns a 5xx status code
func (o *APIError) IsServerError() bool {
	return o.Code/100 == 5
}

// IsCode returns true when this elapse o k response returns a 4xx status code
func (o *APIError) IsCode(code int) bool {
	return o.Code == code
}

```

So you should be able to use it as:

```go
func IsClientError(err error) bool {
	if rerr, ok := err.(runtime.ClientResponseStatus); ok {
            return rerr.IsClientError() 
        }
	return false
}
```

closes #2706

### DIFF
--- a/generator/templates/client/response.gotmpl
+++ b/generator/templates/client/response.gotmpl
@@ -71,16 +71,64 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}) Code() int {
 }
   {{- end }}
 
+// IsSuccess returns true when this {{ humanize .Name }} response returns a 2xx status code
+func ({{ .ReceiverName }} *{{ pascalize .Name }}) IsSuccess() bool {
+  {{- if eq .Code -1 }}
+  return {{ .ReceiverName }}._statusCode/100 == 2
+  {{- else }}
+  return {{ and (ge .Code 200) (lt .Code 300) }}
+  {{- end }}
+}
+
+// IsRedirect returns true when this {{ humanize .Name }} response returns a 3xx status code
+func ({{ .ReceiverName }} *{{ pascalize .Name }}) IsRedirect() bool {
+  {{- if eq .Code -1 }}
+  return {{ .ReceiverName }}._statusCode/100 == 3
+  {{- else }}
+  return {{ and (ge .Code 300) (lt .Code 400) }}
+  {{- end }}
+}
+
+// IsClientError returns true when this {{ humanize .Name }} response returns a 4xx status code
+func ({{ .ReceiverName }} *{{ pascalize .Name }}) IsClientError() bool {
+  {{- if eq .Code -1 }}
+  return {{ .ReceiverName }}._statusCode/100 == 4
+  {{- else }}
+  return {{ and (ge .Code 400) (lt .Code 500) }}
+  {{- end }}
+}
+
+// IsServerError returns true when this {{ humanize .Name }} response returns a 5xx status code
+func ({{ .ReceiverName }} *{{ pascalize .Name }}) IsServerError() bool {
+  {{- if eq .Code -1 }}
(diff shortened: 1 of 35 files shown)


generator/templates/client/response.gotmpl
```diff
@@ -71,16 +71,64 @@ func ({{ .ReceiverName }} *{{ pascalize .Name }}) Code() int {
...
+func ({{ .ReceiverName }} *{{ pascalize .Name }}) IsServerError() bool {
+  {{- if eq .Code -1 }}
+  return {{ .ReceiverName }}._statusCode/100 == 5
+  {{- else }}
+  return {{ and (ge .Code 500) (lt .Code 600) }}
+  {{- end }}
+}
+
+// IsCode returns true when this {{ humanize .Name }} response returns a 4xx status code
+func ({{ .ReceiverName }} *{{ pascalize .Name }}) IsCode(code int) bool {
```
pwmcintyre, line 111: I think this might be a typo?🤔

```suggestion
// IsCode returns true when this {{ humanize .Name }} response a status code equal to that given
func ({{ .ReceiverName }} *{{ pascalize .Name }}) IsCode(code int) bool {
```

Issue #1083, opened 2017-06-29
escaped parameters fail to generate the correct url path when a base path is present.


james-lawrence:
If a path parameter needs to be escaped the runtime code generates an invalid path.

BasePath = "/foo"
Path = "/v1/{param}"

I suggest adding a transport to the HTTP client that dumps the request.
```
import openapiclient "github.com/go-openapi/runtime/client"
c := generatedClient.New(
		openapiclient.NewWithClient("localhost", "/foo", []string{"http"}, http.DefaultClient),
		nil,
)

// all of the following fail.
// passing in a raw string will invoke an invalid path: /foo/v1/param/value
c.ResourceMethod((&ResourceMethodParameters{}).WithParam("param/value"))
// passing in an escaped string will invoke an invalid path: /foo/v1/param/value
c.ResourceMethod((&ResourceMethodParameters{}).WithParam(url.PathEscape("param/value")))
```
These fail because of the runtime/client's interaction with URL.
The http client uses URL.EscapedPath() to determine what path to use.
```
func (u *URL) EscapedPath() string {
	if u.RawPath != "" && validEncodedPath(u.RawPath) {
		p, err := unescape(u.RawPath, encodePath)
		if err == nil && p == u.Path {
			return u.RawPath
		}
	}
	if u.Path == "*" {
		return "*" // don't escape (Issue 11202)
	}
	return escape(u.Path, encodePath)
}
```

client.Runtime.Submit has the following code:
```
func (r *Runtime) Submit(operation *runtime.ClientOperation) (interface{}, error) {
        // ...
	if req.URL.Path != "" && req.URL.Path != "/" && req.URL.Path[len(req.URL.Path)-1] == '/' {
		reinstateSlash = true
	}
	req.URL.Path = path.Join(r.BasePath, req.URL.Path)
	if reinstateSlash {
		req.URL.Path = req.URL.Path + "/"
	}
        // ...
}
```

setting `req.URL.Path = path.Join(r.BasePath, req.URL.Path)` causes url.EscapedPath() to return
req.URL.Path unescaped, since escape(u.Path, encodePath) only escapes the ?, it assumes the fragments have already been properly escaped.

the fix is to properly set both the RawPath and the Path in the submit function.
```
	req.URL.Path = path.Join(r.BasePath, req.URL.Path)
        req.URL.RawPath = path.Join(r.BasePath, req.URL.RawPath)
```

[...]

PR #2713, opened 2022-02-13
support for passing basepath as an argument

svyotov:
Not all projects use as a base path `/`, and this limitations makes generation of projects with go-swagger difficult to maintain.
Please consider this proposal as a solution to the issue described above.

- [x] This is an extension and it is backwards compatible, does not break old way of work, extends it.

Possibly related to https://github.com/go-swagger/go-swagger/issues/1083

### DIFF
--- a/examples/cli/README.md
+++ b/examples/cli/README.md
@@ -56,12 +56,13 @@ $ ./examples/cli/cmd/todoctl/todoctl --help
 $ ./examples/cli/cmd/todoctl/todoctl todos updateOne --help
 ```
 ### Use config file to store common flag values
-`hostname`, `scheme`, and auth tokens can be read from a config file, so that you do not need to enter it every time via command line.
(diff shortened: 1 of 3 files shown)


youyuanwu:
If you follow https://swagger.io/docs/specification/2-0/api-host-and-base-path/ and specify base path in your spec, the generated Default base path should use the one in the spec. This new CLI option added by you is a override of the base path in spec.

svyotov:
> If you follow https://swagger.io/docs/specification/2-0/api-host-and-base-path/ and specify base path in your spec, the generated Default base path should use the one in the spec. This new CLI option added by you is a override of the base path in spec.

Yes, because in some cases the base path is dynamic, it is not hardcoded. This PR still keeps the old behavior, but allows overrides when needed - for servers for example ran behind a load balancer someone might need to provide a custom base path to get the deployment done cleanly.

casualjim:
Thanks for not changing existing behavior

PR #2727, opened 2022-02-23
support `MarshalText` in custom type

emilgpa:
This PR gives supports `MarshalText` in custom type.
The problem is that possibly generates breaking change. So the solution here is:
- Merge this PR in a major version.
- Or to use a flag to enable `MarshalText` for custom type. Something like ` swagger generate spec -o ./swagger.json --marshalText`

Personally, the second option can be good enough.

This code https://go.dev/play/p/tI99ntNYt62 generate this schema https://www.codepile.net/pile/KOGVJqym

I would like to create a couple of tests to verify that everything is correct but I don't have time to see how they are done in depth, could someone give me some clues where to start?

EDIT: I forgot: this generates the format of field to `uuid` if the custom type is named as `uuid`, `UUID` or `Uuid` (L259).


casualjim:
Any chance you could make a test for this?  I rarely touch this codebase and I don't want to break stuff in the future

casualjim (approved):

casualjim:
Excellent, thanks

PR #2738, opened 2022-03-10
allow trimPrefix and trimSuffix in templates

kevinbarbour:
In order to better manipulate strings in custom templates.

fixes #2737 

Signed-off-by: Kevin Barbour <kevinbarbourd@gmail.com>

PR #2741, opened 2022-03-18
Add sprig library for template functions.

### LINKED ISSUE #2737: Add trimPrefix and trimSuffix to template FuncMap
## Feature Request

When using custom templates for code generation I would like to be able to manipulate strings a bit better - specifically with the strings functions `trimPrefix` and `trimSuffix`.

This allows for some useful modification of names defined in Swagger spec that I would like to portray differently in a generated CLI.

kevinbarbour:
Add sprig library for template functions.

Fixes #2737
Alternative to #2738

Removes duplicates from the funcMap that exist already in sprig.
Note that this does leave the `hasPrefix` function overriding the one in
sprig. This is because the sprig function expects arguments of the form
`(prefix, string)` and the existing go-swagger function expects
`(string, prefix)`.

As to not cause a breaking change to the existing templating this was
left as-is.

Signed-off-by: Kevin Barbour <kevinbarbourd@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/go-swagger/go-swagger
 go 1.17
 
 require (
+	github.com/Masterminds/sprig v2.22.0+incompatible
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d // indirect
 	github.com/coreos/go-oidc v2.2.1+incompatible
 	github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -49,17 +49,26 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require github.com/Masterminds/sprig v2.22.0+incompatible
```
youyuanwu, line 52: why this is a separate line? Can it be moved into the groups?

youyuanwu (approved):

go.mod
kevinbarbour: Not sure, just is how `go mod tidy` inserted it.

I put it into the first group and ran the tidy again and now it's good.

PR #2745, opened 2022-03-23
Add support for template plugins

kevinbarbour:
This PR adds the ability to use a [go plugin](https://pkg.go.dev/plugin) to add custom functions to the templating library when using custom templates.

In #2741 I added the sprig library to provide some templating functionality we needed for a project, but it may be valuable to have the ability for users to inject their own template functions without modifications to the go-swagger codebase.

Haven't checked that this compiles on Windows yet, and I'd like to write some tests around the plugin functionality itself but wanted to gauge interest and see if something like this could be a valuable contribution to the project. 

Note that plugins are not supported on Windows hence the splitting of the code into files with `windows` and `!windows` build tags.

### DIFF
--- a/cmd/swagger/commands/generate/shared.go
+++ b/cmd/swagger/commands/generate/shared.go
@@ -112,7 +112,7 @@ func (w WithShared) getConfigFile() string {
 	return string(w.Shared.ConfigFile)
 }
 
-type sharedOptions struct {
+type sharedOptionsCommon struct {
 	Spec                  flags.Filename `long:"spec" short:"f" description:"the spec file to use (default swagger.{json,yml,yaml})" group:"shared"`
 	Target                flags.Filename `long:"target" short:"t" default:"./" description:"the base directory for generating the files" group:"shared"`
 	Template              string         `long:"template" description:"load contributed templates" choice:"stratoscale" group:"shared"`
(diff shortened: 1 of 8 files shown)


casualjim:
I think this is an interesting idea, I always wanted to do plugins for adding extra string formats.

What kind of uses would this unlock beyond what you can achieve with the template config file? I'm not against it, just curious what you have in mind

kevinbarbour:
As far as I'm aware the template config file doesn't let you alter the way in which templates are processed, it just lets you configure which templates are used, the output filename, etc. But if I am missing something let me know.

With this plugin setup you are able to add additional functions to the funcmap for template processing. For a silly example, let's say you needed to use a `rot13` function in your templates.

You can write the following plugin:

```go
package main

import "text/template"

func rot13(s string) string {
	// implement me
	return s
}

func AddFuncs(f template.FuncMap) {
	f["rot13"] = rot13
}
```

Build it as a plugin, which will create `plugin.so`:
`go build -buildmode=plugin plugin.go`

Use your plugin function within one of your custom templates:
`{{ rot13 .Item }}`

And then run swagger generate using the plugin:
`swagger generate cli -t gen --spec=spec.json -T templates -p plugin.so`


I do not actually have a specific use-case that needs this at the moment, but the idea is that in the future if you need some additional templating functionality you don't need to add it to the go-swagger codebase like I did with #2741, you can just pull in your own plugins and modify them as needed.

kevinbarbour:
Fixed up an issue with tests and got the windows build fixed. The failing AppVeyor tests are due to an unrelated issue (#2762).

@casualjim Curious if you have any further feedback on this, and my explained use case above.

casualjim:
This looks good to me

This model could also be used for strfmt extensions which is an often requested feature

kevinbarbour:
> This looks good to me
> 
> This model could also be used for strfmt extensions which is an often requested feature

Thanks Ivan! Should be ready to merge now. 

I will look into the strfmt use case as well.

PR #2751, opened 2022-03-30
1957: Support for extensions within swagger:route

### LINKED ISSUE #1957: Support for extensions within swagger:route
## Problem statement
Vendor extensions are currently supported only as part of swagger:meta in go-swagger. However, extensions are allowed as part of paths too as referenced here: https://swagger.io/docs/specification/2-0/swagger-extensions/

As an example, the AWS API Gateway injects extensions at each path.
```
paths:
  /alerts:
    get:
      description: |-
        List alerts.
      operationId: ListAlerts
      responses:
        "200":
          $ref: '#/responses/ListAlertsResponse'
        default:
          description: Error
          schema:
            $ref: '#/definitions/Error'
      security:
      - CongnitoAuthorizer: []
      summary: List Alerts
      tags:
      - Alerts
      x-amazon-apigateway-integration:
        httpMethod: GET
        passthroughBehavior: when_no_match
        responses:
          default:
            statusCode: "200"
        type: http_proxy
        uri: https://myproxy.com/alerts
```

## Environment
swagger version: 0.19.0
go version: 1.11.6  
OS:   macOs

Huckletoon:
## Summary ##
Adding basic functionality to support the definition of vendor extensions on a per-path basis as requested in issue #1957. This would be defined by a developer within a `swagger:route` block of their codebase, and the corresponding swagger definition would be created when running `swagger generate spec`. These changes currently only support go1.11+.

These changes are making the following assumptions:
- the value of a vendor extension can be a single string, a list/array of strings, or objects
- a list cannot contain a mix of strings and objects
- annotations defining extensions should be indented properly
- each extension key must start with `x-`, e.g. `x-some-extension`
- objects defined in an extension may contain multiple other nested objects

The new extension parser reads through the `Extension` annotation block within a `swagger:route` annotation, checking for what kind of data each line may be and builds out as many `map[string]interface{}`s as necessary until reaching the end of the block. 
- Simple key:value extensions will be `map[string]string`
- Lists/Arrays will be `map[string]*[]string`
- Objects will be `map[string]interface{}` that can contain any of the above types, including more objects

#### Input: ####
```go
// swagger:route GET /pets pets users listPets
//
// Lists pets filtered by some parameters.
//
//     Extensions:
//       x-example-flag: true
//       x-some-expected-list:
//         - dog
//         - cat
//         - bird
//       x-env-data:
//         userId:
//           prod: some-user-id
//           dev: some-test-id
```
#### Output: ####
```yaml
paths:
  "/pets":
    get:
      operationId: listPets
      summary: Lists pets filtered by some parameters.
      tags:
      - pets
      - users
      extensions:
        x-example-flag: true
        x-some-list:
        - dog
        - cat
        - bird
        x-env-data:
          userId:
            prod: some-user-id
            dev: some-test-id
```

Fixes #1957

### DIFF
--- a/codescan/parser.go
+++ b/codescan/parser.go
@@ -1683,7 +1683,7 @@ func (ss *setOpExtensions) Parse(lines []string) error {
 		if _, ok := ext.Root.(map[string]string); ok {
 			exts.AddExtension(ext.Extension, ext.Root.(map[string]string)[ext.Extension])
 		} else if _, ok := ext.Root.(map[string]*[]string); ok {
-			exts.AddExtension(ext.Extension, ext.Root.(map[string]*[]string)[ext.Extension])
+			exts.AddExtension(ext.Extension, *(ext.Root.(map[string]*[]string)[ext.Extension]))
 		} else if _, ok := ext.Root.(map[string]interface{}); ok {
 			exts.AddExtension(ext.Extension, ext.Root.(map[string]interface{})[ext.Extension])
 		} else {
--- a/codescan/parser.go
+++ b/codescan/parser.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"go/ast"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


casualjim (approved):

casualjim:
Thanks. nice  piece of code.

PR #2772, opened 2022-05-11
Add details on how to install using go install

vidhill:
I was adding instructions on how to install the command line tool to a new project and I was unsure if the cli tool could be installed using `go install`

I worked it out, so adding details on how to install using `go` here


vidhill:
Not sure why Circle Ci is failing, as I only updated a markdown file..

fredbi (approved):

Issue #2758, opened 2022-04-12
[CLI] Generating an empty object using `swagger:type object` does not work as expected.


adam-debkowski:
## Problem statement

I'm trying to generate an OpenAPI spec from code that skips the `additionalProperties` when generating spec based on a `map[string]interface{}` type.
So I want a definition of:
```go
type CustomMetadata map[string]interface{}
```
To generate:
```yml
  CustomMetadata:
    type: object
```

As far as I observed, the `additionalProperties` property comes from using a `map[string]interface{}`, since when using a simple `interface{}`, it generates the expected output. A definition:
```go
type CustomMetadata interface{}
```
Generates:
```yml
  CustomMetadata:
    type: object
```

I have tried to achieve that using `swagger:format`. I expected a definition of:
```go
// swagger:type object
type CustomMetadata map[string]interface{}
```
To generate:
```yml
  CustomMetadata:
    type: object
```
so without the `additionalProperties` field. But it is not the case and the `// swagger:type object` seems to be ignored. The same happens with a type definition of `type CustomMetadata interface{}`: no matter the `swagger:type`, the type definitions based on a `map` or `interface` always generate their respective OpenAPI specs disregarding a `swagger:type`.

I've also checked other types, and so for example:
```go
// swagger:type object
type CustomMetadata string
```
Generates:
```yml
  CustomMetadata: {}
```

This is still not ideal, since as far as I know, the `{}` in Swagger 2.0 means "any type", and not an object specifically. But it shows that the `swagger:type` is taken into account in this case.

The current solution is to use `interface{}` instead of `map[string]interface{}` which generates expected output, but is not ideal since it requires a lot of changes in the actual application code.

## Swagger specification

```yml
definitions:
  CustomMetadata:
    additionalProperties:
      type: object
    type: object
info:
  description: Test description
  title: Test
  version: 1.2.3
paths:
  /pets:
    get:
      operationId: listPets
      responses:
        "200":
          $ref: '#/responses/someResponse'
responses:
  someResponse:
    description: ""
    schema:
      properties:
        CustomMetadata:
          $ref: '#/definitions/CustomMetadata'
      type: object
swagger: "2.0"
```

## Steps to reproduce

```go
// Test
//
// Test description
//
//     Version: 1.2.3
//
// swagger:meta
package main

// swagger:route GET /pets listPets
//
//     Responses:
//       200: someResponse

// swagger:response someResponse
type SomeResponse struct {
	// in: body
	Body struct {
[...]

PR #2776, opened 2022-05-21
fixbug: `interface{}` in a `map` as value should be generated/typed to `any`

emilgpa:
```go
package main

// swagger:model Example
type Example struct {
	MyField map[string]interface{} `json:"my_field,omitempty"`
}
```
should to generate:

```
{
  "swagger": "2.0",
  "paths": {},
  "definitions": {
    "Example": {
      "type": "object",
      "properties": {
        "my_field": {
          "type": "object",
          "additionalProperties": {},
          "x-go-name": "MyField"
        }
      }
    }
  }
}
```

but currently generates:

```
{
  "swagger": "2.0",
  "paths": {},
  "definitions": {
    "Example": {
      "type": "object",
      "properties": {
        "my_field": {
          "type": "object",
          "additionalProperties": {
            "type": "object"
          }
          "x-go-name": "MyField"
        }
      }
    }
  }
}
```

`interface{}` as value in a `map` should be mapped to `any`. This PR fixes this! 

this issue is more described in: #2758

### DIFF
--- a/codescan/application_test.go
+++ b/codescan/application_test.go
@@ -47,7 +47,7 @@ func loadClassificationPkgsCtx(t testing.TB, extra ...string) *scanCtx {
 func TestApplication_LoadCode(t *testing.T) {
 	sctx := loadClassificationPkgsCtx(t)
 	require.NotNil(t, sctx)
-	require.Len(t, sctx.app.Models, 31)
+	require.Len(t, sctx.app.Models, 32)
 	require.Len(t, sctx.app.Meta, 1)
 	require.Len(t, sctx.app.Routes, 7)
 	require.Empty(t, sctx.app.Operations)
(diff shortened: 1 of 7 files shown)


casualjim:
I don't think this PR fixes #2758 

As far as I can tell this PR just removes the default of type: object to be any type of json value.  But it doesn't address the real issue that is mentioned which is respecting `swagger:type`

casualjim (changes requested):
Are you going to work on supporting `swagger:type`

emilgpa:
@casualjim Yes, you're right. It is also related to `swagger:type`. Although I understand from the code that `swagger:type object` doesn't exist. Could that be the reason the guy was trying to use it to avoid the `additionalProperties` generated by `interface{}`?. Anyway, i included in the PR support `swagger:type object` (it did not exists).

casualjim (approved):

PR #2779, opened 2022-05-26
Bugfix: err shadows declaration

epanipinto-jc:
Fixes error generated from linting by `go vet`.

```
Starting go vet...
/rest/server.go:192:7: declaration of "err" shadows declaration at line 154
/rest/server.go:222:7: declaration of "err" shadows declaration at line 154
```

### DIFF
--- a/generator/templates/server/server.gotmpl
+++ b/generator/templates/server/server.gotmpl
@@ -336,7 +336,7 @@ func (s *Server) Serve() (err error) {
 		s.Logf("Serving {{ humanize .Name }} at unix://%s", s.SocketPath)
 		go func(l net.Listener){
 		  defer wg.Done()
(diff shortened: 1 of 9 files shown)


casualjim (changes requested):
I think it makes more sense to use another name than to refer to err variable that is shared in different go routines

generator/templates/server/server.gotmpl
```diff
@@ -366,7 +366,7 @@ func (s *Server) Serve() (err error) {
 		s.Logf("Serving {{ humanize .Name }} at http://%s", s.httpServerL.Addr())
 		go func(l net.Listener) {
 			defer wg.Done()
-			if err := httpServer.Serve(l); err != nil && err != http.ErrServerClosed {
+			if err = httpServer.Serve(l); err != nil && err != http.ErrServerClosed {
```
casualjim, line 369: it's better to rename because this is used in a different go routine

PR #2806, opened 2022-07-29
add option to select custom struct tag for spec-generation

fufuceng:
Hi All,

First, thanks for this library. It makes many things easier to implement.

I'm using go-swagger to generate swagger files from the code for almost 3 years. I've been searching for a way to provide custom names other than JSON tag and field name for a while. I thought that providing a custom struct tag while generating spec would be awesome.

Example:

Let's say I've struct like

```go

type CreateUserRequest struct {
       XForwardedFor []string `json:"X-Forwarded-For" header:"X-Forwarded-For"`
       FirstName string `json:"firstName"`
}
```
I'm decoding the request according to the `json, header, query` tags. But I have to add the JSON tag to `XForwardedFor` field to show it in the swagger with the correct name. In this scenario, the client can override the XForwardedFor header by sending `X-Forwarded-For` field in the body. I think `--struct-tag` option can be useful in anyways. 

Usage:
`swagger generate spec -o ./docs/swagger.yaml --scan-models -w ./examples/generated --struct-tag swg`

With `--struct-tag swg` struct will look like 

```go
type CreateUserRequest struct {
       XForwardedFor []string `header:"X-Forwarded-For" swg:"X-Forwarded-For"`
       FirstName string `json:"firstName" swg:"firstName"`
}
```

Signed-off-by: furkan <fufuceng@gmail.com>

### DIFF
--- a/cmd/swagger/commands/generate/spec_go111.go
+++ b/cmd/swagger/commands/generate/spec_go111.go
@@ -30,6 +30,7 @@ type SpecFile struct {
 	IncludeTags []string       `long:"include-tag" short:"" description:"include routes having specified tags (can be specified many times)"`
 	ExcludeTags []string       `long:"exclude-tag" short:"" description:"exclude routes having specified tags (can be specified many times)"`
 	ExcludeDeps bool           `long:"exclude-deps" short:"" description:"exclude all dependencies of project"`
+	StructTag   string         `long:"struct-tag" short:"" description:"specify which struct tag will be used" default:"json"`
 }
 
 // Execute runs this command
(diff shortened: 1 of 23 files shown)


mkaykisiz (approved):

youyuanwu:
Have you tried to modify/add tags in generated code using https://github.com/fatih/gomodifytags ?
I would do that instead of change the generator.

mkaykisiz (approved):

casualjim:
It would be good if the PR also included a fallback to the json tag, so look at custom tag first and then fall back to the json tag when it's not present, and perhaps allow for multiple alternative tags to be provided.

That way your use case can be addressed without adding extra swagger tags but rather look up in order: header, query, json for example.

fredbi:
Superseded by https://github.com/go-openapi/codescan/pull/55
Which will ship in the next go-swagger release

PR #2832, opened 2022-09-08
fixes #2831, circular dependency issue

### LINKED ISSUE #2831: Handle circular references
## Problem statement
When you have a circular reference, the swagger DIFF command goes into an infinite loop. I think this is something to be supported because below is perfectly valid in Go

```
type A1 struct {
 id int
 name string
 a1 *A1
}
```

## Steps to reproduce
*spec_1.json*
```
{
  "swagger": "2.0",
  "info": {
    "title": "Swagger Fixture",
    "version": "1.0"
  },
  "paths": {
    "/a/": {
      "get": {
        "parameters": [
          {
            "name": "",
            "in": "body",
            "schema": { "$ref": "#/definitions/A1" }
          }
        ],
        "responses": {
          "200": {
            "description": "200 response",
            "schema": {
              "type": "array",
              "items": { "$ref": "#/definitions/A1" }
            }
          }
        }
      }
    }
  },
  "definitions": {
    "A1": {
      "type": "object",
      "properties": {
        "id": { "type": "integer" },
        "name": { "type": "string" },
        "a1":{ "$ref": "#/definitions/A1" }
      }
    }
  }
}
```
*spec_2.json*
*You can just use the same one above*

Then Run `swagger diff spec_1.json spec_2.json`. It will never stop.

## Environment
swagger version: 2.0  
go version: 1.18
OS:  Mac Monterey

Aven30:


Aven30:
My approach was to track what schemas, based off of location, has already been compared. This  makes it where we can complete the DIFF on a specs that have circular references as expected.

fixtures/diff/refcircular.v2.json
```diff
@@ -0,0 +1,39 @@
...
+          }
+        }
+      }
+    }
+  },
+  "definitions": {
+    "A1": {
+      "type": "object",
+      "properties": {
+        "id": { "type": "integer" },`
```
casualjim, line 33: That ` should not be there

casualjim (changes requested):

Aven30:
Gahh, my bad, removed!

casualjim (approved):

PR #2849, opened 2022-10-09
 support serve multiple sepcs; support auto reloading spec files; support swagger-ui topbar plugin; support server gracefull shutdown;support customizing  documentation site title.

chaosue:
as described in title.
we can use `--spec-file` to specify multiple spec files to serve, `--auto-reload-specs` to enable auto renewing specs periodically, `--site-title` to set the documentation site title.
图片

图片


casualjim:
the swaggerui middlewater in this PR should be put in go-openapi/middleware, it doesn't belong in this code generation tool

PR #2863, opened 2022-11-29
Configurable consumers in stratoscale templates

jhernand:
Currently the configuration code generated by the _stratoscale_ templates doesn't allow changing the consumers before creating the handler. This makes it difficult to use custom consumers. For example, we would like to use a JSON consumer that rejects unknown fields. To make that easier this patch changes the templates so that the _Config_ object that is passed to the _Handler_ function has additional fields to specify those custom consumers. With that in place we will be able to do something like this:

```go
// myConsumer is a consumer function that rejects unknown fields.
func myConsumer() runtime.Consumer {
        return runtime.ConsumerFunc(func(reader io.Reader, data interface{}) error {
                dec := json.NewDecoder(reader)
                dec.UseNumber()
                dec.DisallowUnknownFields()
                return dec.Decode(data)
        })
}

// Create the handler using our custom consumer:
handler, err := restapi.Handler(restapi.Config{
	....
	JSONConsumer: myConsumer(),
})
```

Existing code will not be affected because the default value for this new `JSONConsumer` field will be nil, and that results in the same behaviour than before this change.

### DIFF
--- a/examples/contributed-templates/stratoscale/restapi/configure_petstore.go
+++ b/examples/contributed-templates/stratoscale/restapi/configure_petstore.go
@@ -85,6 +85,9 @@ type Config struct {
 	BasicAuthenticator func(security.UserPassAuthentication) runtime.Authenticator
 	// Authenticator to use for all Basic authentication
 	BearerAuthenticator func(string, security.ScopedTokenAuthentication) runtime.Authenticator
+
+	// JSONConsumer is a json consumer that will replace the default if not nil.
+	JSONConsumer runtime.Consumer
 }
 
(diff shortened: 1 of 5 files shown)


jhernand:
@filanov I think you wrote the first version of these templates, I'd appreciate your review.

jhernand:
@casualjim would you be so kind to approve running the workflows? Would be nice if you can review as well.

casualjim (approved):

jhernand:
Thanks @casualjim. Is there any chance that you can do a new release containing this change?

PR #2870, opened 2022-12-17
Make stratoscale template easier to test

samj1912:
This PR modifies the stratoscale template to make it easier to test.

1. The facade uses the operation group interfaces (API) instead of using the actual client struct. This makes it easy to mock things out during tests.
2. mockery v2 is used instead of mockery v1. Mocks are generated in a mocks package inside the operation group package now instead of the same package. (--keeptree vs --inpackage)
3. Adds the expecter interface to mockery which is preferable and keeps the file casing as underscore to comply with golang file conventions.

### DIFF
--- a/examples/contributed-templates/stratoscale/client/pet/pet_client.go
+++ b/examples/contributed-templates/stratoscale/client/pet/pet_client.go
@@ -13,7 +13,7 @@ import (
 	strfmt "github.com/go-openapi/strfmt"
 )
 
-//go:generate mockery -name API -inpkg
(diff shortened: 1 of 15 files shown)


casualjim (approved):

samj1912:
@casualjim thanks for taking a look! Could you please re-approve, I had to make a couple of small changes. See https://github.com/go-swagger/go-swagger/compare/565fb66f64c1bff624b40d469d24a76e06f45991..779defea9ff30013d74c97f3dbcdb70c2d42e4b9

PR #2879, opened 2023-01-12
Ensure that the stratoscale client is able to handle non success-response types

samj1912:
This change adds 2 things

- Expose a Code method that returns the status code from a response
- Handle non-success response types in the stratoscale template.


casualjim (approved):

samj1912:
@casualjim thanks for approving the PR! I believe the tests should now be passing and this should be good to merge :)

Separately, would it be possible to cut a patch release? There are a bunch of commits on master that are useful bug fixes.

PR #2896, opened 2023-02-01
Indicate consumer parse error and provide server response code and message body

wildsheepz:
When the server returns an error for unexpected communication errors eg. using http schema instead of https, the parser error message could be misleading and not helpful at all.


wildsheepz:
Found a new unrelated issue, will fix before reopeb

PR #2926, opened 2023-04-19
Fixed panic in `AddMiddlewareFor` method if method is not uppercase

iv-menshenin:
Use the same variable to set the handler value by key

### DIFF
--- a/generator/templates/server/builder.gotmpl
+++ b/generator/templates/server/builder.gotmpl
@@ -441,6 +441,6 @@ func ({{.ReceiverName}} *{{ pascalize .Name }}API) AddMiddlewareFor(method, path
   }
   {{.ReceiverName}}.Init()
   if h, ok := {{.ReceiverName}}.handlers[um][path]; ok {
-    {{.ReceiverName}}.handlers[method][path] = builder(h)
+    {{.ReceiverName}}.handlers[um][path] = builder(h)
   }
 }


iv-menshenin:
If you call a method and pass a non uppercase word as an argument to method, it will inevitably panics because we are not using the same (that we checked when getting) value as the key when setting the value.

casualjim (approved):

iv-menshenin:
Why didn't it pass the lint? I don't see any details

iv-menshenin:
@casualjim can you help me with linter? What does it want from me?

casualjim:
Sorry I saw this failure and I fixed the build on master, not sure how that ever went through in the first place. 
I've just rebased your branch, we'll see how it does now

iv-menshenin:
Forgot `signed-off`

iv-menshenin:
Sorry, I forgot my `signed-off` in comment and then couldn't deal with your rebase in my branch =)
Please fix it yourself, or I can repeat my request  later

casualjim:
the signed off I should remove that. the pr was good

iv-menshenin:
> If you call a method and pass a non uppercase word as an argument to method, it will inevitably panics because we are not using the same (that we checked when getting) value as the key when setting the value.

re-created the request

https://github.com/go-swagger/go-swagger/pull/2928

Issue #2932, opened 2023-04-27
[Proposal] Introducing new compatibility level "Warning" for changes which are technically non-breaking but can cause unforeseen behavior change for client


schafle:
## Introducing compatibility level "Warning"
There are certain changes diff tool reports which are technically non-breaking but can cause unforeseen behavior changes in client and should be reported differently. Cases where this is applicable:
- Default value is added/removed/changed
- Extension is added/removed
- An enum value is added

schafle:
I will hold off merging my PR https://github.com/go-swagger/go-swagger/pull/2931 if this proposal is accepted.

casualjim:
I think it's a good idea, you can consider it accepted :+1:

schafle:
Done with #2933. Closing.

PR #2933, opened 2023-04-27
feat(diff): introducing warning compatibility level

schafle:
This PR just adds a new compatibility level warning. Warning should be used for changes which are technically non-breaking but can cause behavior changes on client.

Implements #2932


schafle:
I am open to a better name than "Warning" as "Warning changes" doesn't seem appropriate.

casualjim:
Warn is also what's used in other similar tools, like code linters etc. I think this can be expanded upon for users to allow uplifting warnings to errors

casualjim (approved):

casualjim:
You could maybe rename it to something like: UserImpacting if you prefer but for me Warn is also fine

PR #2947, opened 2023-06-07
Use proper operation name in default client error

### LINKED ISSUE #2945: Default client error lacks context which is the failing operation
## Problem statement

If an API call returns a status code that is not defined in the swagger spec, we get an error message like following:
```
response status code does not match any response statuses defined for this endpoint in the swagger spec (status 503): {}
```
This lacks context which is the failing operation. A program may call many different APIs, then it is hard to tell which is the failing one.
While this context can be added in the application code making the call, it would require changes in all the places the application makes some API calls. It would be more efficient to handle this in a single place. Also the information about the API operation being executed is available in the swagger client.

Ideally the error message would contain the failing request method and complete URL.

## Swagger specification

Any swagger spec without a default response.

## Steps to reproduce

Call an API operation that returns a status code that is not defined in the swagger spec.

petar-dochev-f3:
Fixes #2945

### DIFF
--- a/fixtures/bugs/2945/fixture-2945.yaml
+++ b/fixtures/bugs/2945/fixture-2945.yaml
@@ -1,0 +1,13 @@
+swagger: "2.0"
+info:
+  title: foo api
+  version: "1.0.0"
+paths:
+  /version:
+    get:
+      operationId: getVersion
+      responses:
+        "200":
+          description: Success
+          schema:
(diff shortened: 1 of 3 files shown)


casualjim (approved):

petar-dochev-f3:
Thanks for the merge @casualjim. When is the next release planned?

PR #2955, opened 2023-07-18
Fix #2783

### LINKED ISSUE #2783: Models get mixed when using structs from several packages
## Problem statement

When generating spec from code, using structs with the same name from different packages results in their model definitions getting mixed up. **This is a major problem when integrating third party packages (e.g. Docker and Kubernetes in the same codebase).**

## Swagger specification

```json
{
  "swagger": "2.0",
  "paths": {},
  "definitions": {
    "Test": {
      "type": "object",
      "required": [
        "c"
      ],
      "properties": {
        "c": {
          "type": "string",
          "x-go-name": "C"
        }
      },
      "x-go-package": "test/c"
    },
    "TestResponseBody": {
      "type": "object",
      "required": [
        "test1",
        "test2"
      ],
      "properties": {
        "test1": {
          "$ref": "#/definitions/Test"
        },
        "test2": {
          "$ref": "#/definitions/Test"
        }
      },
      "x-go-package": "test"
    }
  },
  "responses": {
    "TestResponseBody": {
      "description": "",
      "schema": {
        "$ref": "#/definitions/TestResponseBody"
      }
    }
  }
}
```

## Steps to reproduce

In package A:

```go
// swagger:model TestResponseBody
type TestResponseBody struct {
    // required: true
    // in: body
    Test1 b.Test `json:"test1"`
    // required: true
    // in: body
    Test2 c.Test `json:"test2"`
}

// swagger:response TestResponseBody
type TestResponse struct {
    // required: true
    // in: body
    Body TestResponseBody
}
```

In package B:

```go
// swagger:model Test
type Test struct {
    B string `json:"b"`
}
```

In package C:

```go
// swagger:model Test
type Test struct {
    // required: true
    C string `json:"c"`
}
```

This will result in the `Test` model being added only once. Instead, the generation routine should start numbering the models.

## Why naming models explicitly doesn't work

When integrating third party data structures, the comments can't be applied to the codebase. This is the case when integrating Docker and Kubernetes, both have a model called `Volume`, which will result in the following error:

```
Validation failed. Property 'Driver' listed as required but does not exist in '/definitions/Volume'
```

## Environment
swagger version: 0.29.0  
go version: 1.18.2
OS:  linux/amd64

rwwiv:
From https://github.com/go-swagger/go-swagger/pull/2784

> This PR attempts to fix https://github.com/go-swagger/go-swagger/issues/2783 where duplicate struct names were handled incorrectly. This may only be a partial fix, please provide feedback.

Reopening this as-is, tests are currently failing with this change which needs to be fixed before this is ready for review.


casualjim:
thanks, I think this looks good. We do something similar on the codegeneration side

EugeneWImprint:
Hi! Is there any update on this? It seems like go-swagger still has issues handling duplicate struct names across different packages

fredbi:

PR #2960, opened 2023-08-13
Fix description for write timeout default value

aiwasaki126:
The default value for --write-timeout is defined in gotmple (generated server struct part):
https://github.com/go-swagger/go-swagger/blob/6a14cbceb631a433899a34201dc9eca44f46a1b2/generator/templates/server/server.gotmpl#L233


casualjim:
You're changing the default value for it, this is not a fix for the description. You can already set the value when you run the server. 60s is a *very* long time for REST requests

aiwasaki126:
Thank you for your reviewing. I found that the generated server.go has default 60s write timeout in default tag, so I wrote this PR to unify the write timeout description to 60s. However, I should write up an issue referring to that, and I agree "60s is a very long time for REST requests".  

I'll submit a new issue. Thanks!

Issue #2952, opened 2023-06-26
Go-swagger diff runtime error when comparing schema with different response codes


a-betenev:
## Problem statement

Command "go-swagger diff" crashes when comparing two simple Swagger schema files. The output is:

~~~~~
2023/06/26 09:52:14 Run Config:
2023/06/26 09:52:14 Spec1: ref.yaml
2023/06/26 09:52:14 Spec2: new.yaml
2023/06/26 09:52:14 ReportOnlyBreakingChanges (-c) :false
2023/06/26 09:52:14 OutputFormat (-f) :txt
2023/06/26 09:52:14 IgnoreFile (-i) :none specified
2023/06/26 09:52:14 Diff Report Destination (-d) :stdout
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x78571c]
goroutine 1 [running]:
github.com/go-swagger/go-swagger/cmd/swagger/commands/diff.getSchemaDiffNode(0xbeb29d, 0x4, 0xbdb2a0, 0x0, 0xc000406c00)
        /app/cmd/swagger/commands/diff/node.go:68 +0x49c
github.com/go-swagger/go-swagger/cmd/swagger/commands/diff.(*SpecAnalyser).analyseResponseParams(0xc000871b18)
        /app/cmd/swagger/commands/diff/spec_analyser.go:225 +0x28b
github.com/go-swagger/go-swagger/cmd/swagger/commands/diff.(*SpecAnalyser).Analyse(0xc000871b18, 0xc000120200, 0xc000120100, 0x90, 0x0)
        /app/cmd/swagger/commands/diff/spec_analyser.go:57 +0x109
github.com/go-swagger/go-swagger/cmd/swagger/commands/diff.Compare(0xc000120200, 0xc000120100, 0x0, 0x0, 0x0, 0xc0002e4280, 0x0)
        /app/cmd/swagger/commands/diff/reporting.go:22 +0x7e
github.com/go-swagger/go-swagger/cmd/swagger/commands.(*DiffCommand).getDiffs(0xc0002d9560, 0xc0004a3c10, 0x79d0ce, 0xc000513900, 0xc0004a3c30, 0x3)
        /app/cmd/swagger/commands/diff.go:135 +0x111
[...]

PR #2973, opened 2023-09-07
Updated to handle cases where 'body' is not present in the YAML file …

Shimizu1111:
I have addressed GitHub Issue https://github.com/go-swagger/go-swagger/issues/2952 
Updated to handle cases where 'body' is not present in the YAML file without causing an error.


Franogales:
how can I replicate the error?  I usualy validate if the body != nil. why is this needed? what would be the output?

Shimizu1111:
See #2952 for contents and how to reproduce the error.

casualjim (approved):

PR #2991, opened 2023-11-26
setup dependabot

mmorel-35:
This setup dependabot to help keeping github-actions and gomod dependencies up-to-date.
Creates a PR in case a new version is available.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,7 @@ updates:
     directory: /
     schedule:
       interval: weekly
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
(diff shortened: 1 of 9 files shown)


casualjim:
This one I'm not sure about, it would need to be enabled on all the go-openapi repos too and the value of it is limited.
I say the value is limited, because in golang you're never pinning always just hinting at a given minimum version. So if a dependency is a problem people can simply go get it in their own repo to override ours. This really only provides value in major version upgrades or when there are breaking changes that need to be addressed.

At the same time it creates a hamster wheel because those PR's need to be reviewed and merged, in this case weekly. At the moment it's mostly @youyuanwu and me who are doing reviews and merges.

mmorel-35:
I can remove gomod part and just keep github-actions if you prefer ?
Or just close this PR if you don’t want it either ?

PR #3001, opened 2023-12-16
Add missing line from custom server example code

kevinbarbour:


docs/tutorial/custom-server.md
```diff
@@ -144,6 +144,8 @@ Putting that all together, we have the following main function:
 
 ```go
 func main() {
+    var portFlag = flag.Int("port", 3000, "Port to run this service on")
```
youyuanwu, line 147: nit: indent.

kevinbarbour: Thanks - fixed.

youyuanwu:
This is not a code change. ci failure has to do without setup issues. I will merge this and fix ci later.

Issue #3002, opened 2023-12-19
Enum fields are not properly scanned (and documented)


dimovnike:
## Problem statement

Although not documented here https://goswagger.io/use/spec/model.html the codescan attempts to parse the // Enum: comments but the result is not as expected. Also, the generated documentation comment is impossible to parse when enum items are strings and contain spaces or special chars.

## Steps to reproduce

here is a spec that illustrates both problems:

swagger spec:

```yaml
swagger: "2.0"
info:
  version: "1.0.0"
  title: "Test"

paths: {}
definitions:
  test_enum:
    type: object
    properties:
      enum_string_field:
        type: string
        enum: ["1", "with space", 'with "quotes"']
      enum_int_field:
        type: integer
        enum: [-1, 2, 6000]
```
When code is generated using this spec (`go run cmd/swagger/swagger.go generate model -t test/ -f test/swagger.yml`) we get the following result:

```go
// TestEnum test enum
//
// swagger:model test_enum
type TestEnum struct {

	// enum int field
	// Enum: [-1 2 6000]
	EnumIntField int64 `json:"enum_int_field,omitempty"`

	// enum string field
	// Enum: [1 with space with "quotes"]
	EnumStringField string `json:"enum_string_field,omitempty"`
}
```

It is already clear that the EnumStringField items can not be parsed because it is impossible to reconstruct where the spaces were initially. However, here is the generate spec result (`go run cmd/swagger/swagger.go generate spec  -w test/ --scan-models --include=test  -o test/swagger-gen.yml`):

```yaml
definitions:
    test_enum:
        description: TestEnum test enum
        properties:
            enum_int_field:
                description: enum int field
                enum:
                    - '[-1 2 6000]'
                format: int64
                type: integer
                x-go-name: EnumIntField
            enum_string_field:
                description: enum string field
                enum:
                    - '[1 with space with "quotes"]'
                type: string
                x-go-name: EnumStringField
        type: object
        x-go-name: TestEnum
        x-go-package: github.com/go-swagger/go-swagger/test/models
paths: {}
swagger: "2.0"
```

And we see how enum are parsed as a single item.

## Fix proposal.

First we need a better documentation format for the items, and we can use json:

```go
type TestEnum struct {

	// enum int field
	// Enum: [-1,2,6000]
	EnumIntField int64 `json:"enum_int_field,omitempty"`

	// enum string field
	// Enum: ["1","with space","with \"quotes\""]
[...]

PR #3004, opened 2023-12-19
3002 enum parse

dimovnike:
fixes both issues described in https://github.com/go-swagger/go-swagger/issues/3002


codescan/parser.go
```diff
@@ -1467,17 +1467,34 @@ func (ss *setOpResponses) Parse(lines []string) error {
 }
 
 func parseEnum(val string, s *spec.SimpleSchema) []interface{} {
-	list := strings.Split(val, ",")
-	interfaceSlice := make([]interface{}, len(list))
-	for i, d := range list {
-		v, err := parseValueFromSchema(d, s)
```
fredbi, line 1473: How about users who created their docstrings for enum manually, not using json?

fredbi (changes requested):
I agree that making the list of values a JSON array simplifies things a lot.
But how would we support code that has already been written manually (not from codegen) by users of the previous version?
Couldn't we add some JSON-array detection to switch to the new mode?

codescan/parser.go
dimovnike: I can change the implementation so that when JSON decoding fails it will fall back to the old version. I also recommend a deprecation warning when this happens. Are there already such warnings so I can display them in a similar manner? Any recommendation for it?

dimovnike: It sems, I also forgot some debug output, will remove that too.

fredbi: example of  a warning: https://github.com/go-swagger/go-swagger/blob/f1c27cb213c57d0953a020aa87ab87db3f52ddbc/codescan/schema.go#L228

dimovnike: Thanks for the example. The change is done. I also added the old test cases and adjusted the tests.

fredbi (approved):
LGTM. Thanks a lot for contributing.

PR #3018, opened 2023-12-22
chore(ci): linter & ci configuration

fredbi:
* lint: fixed outdated configuration for linter "depguard" (resulting in many alerts on local testing with the latest version of golangci-lint)
* ci: updated github actions
* ci: use latest golangci-lint
* ci: replace "go1.20" by "oldstable", added go1.21 with "stable" - kept go1.19 for now

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,9 +22,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
+      - uses: actions/setup-go@v4
         with:
-          go-version: "^1"
+          go-version: stable
           check-latest: true
           cache: true
       - name: golangci-lint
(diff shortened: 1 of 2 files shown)


fredbi:
Ouch: 31 min run for CI... That's long.

fredbi:
Now running codegen only on 2 versions. If we change the "required" action in branch check rules, I'd rather have "oldstable" and "stable" and not running CI on go1.19 any longer (even tough our go.mod builds with go1.19)

youyuanwu:
If the lowest supported version is 1.19, we probably should run on it.

youyuanwu (approved):

fredbi:
> If the lowest supported version is 1.19, we probably should run on it.

The regular test suite is tested against 1.19, so it builds and test. But for the full codegen suite, a matrix with 3 go versions takes super long to run. I think that we can prioritize there latest (go1.21) over oldest (go1.19), given that our official policy is to support only the 2 latest releases (but yes, go.mod compatibility can lag behind a bit)

PR #3024, opened 2023-12-27
fix(mangling): fixes name mangling in the presence of a special character

### LINKED ISSUE #2821: JSON Schema in swagger API response
I have an API that returns a body that corresponds to a JSON Schema, The JSON schema structure is included in the API spec file. But when I try to generate the client I get the following error:

name field ArraySchema
package field models
creating generated file "array_schema.go" in "models" as definition
executed template asset:model
source formatting failed on template-generated source ("models\\array_schema.go" for definition). Check that your template produces valid code
unformatted generated source "array_schema.go" has been dumped for template debugging purposes. DO NOT build on this source!
source formatting on generated source "definition" failed: models\array_schema.go:118:34: expected ';', found string (and 10 more errors)

## Swagger specification

[ServiceManagementBody.zip](https://github.com/go-swagger/go-swagger/files/9485408/ServiceManagementBody.zip)

## Steps to reproduce

run 
swagger genereate client -f ServiceManagementBody.json -A client

## Environment
swagger version: 2.0
go version:  1.18.4
OS:   Windows 11

fredbi:
This PR fixes a special case when a "$" is inside a name AND we generate polymorphic types.

In this case the unexported name was not properly generated (included a blank space). More generally, swag mangling functions fail to properly handle this extra blank space. Fortunately, ToGoName wasn't affected and this is why this failure occurs in such rare circumstances.

* requires go-openapi/swag#78
* fixes #2821

### DIFF
--- a/generator/template_repo_test.go
+++ b/generator/template_repo_test.go
@@ -672,3 +672,25 @@ func TestGt0(t *testing.T) {
 	require.False(t, gt0(swag.Int64(0)))
 	require.False(t, gt0(nil))
 }
+
+func TestIssue2821(t *testing.T) {
+	tpl := `
+Pascalize={{ pascalize . }}
+Camelize={{ camelize . }}
+`
+
+	require.NoError(t,
+		templates.AddFile("functpl", tpl),
+	)
+
+	compiled, err := templates.Get("functpl")
+	require.NoError(t, err)
+
+	rendered := bytes.NewBuffer(nil)
+	require.NoError(t,
+		compiled.Execute(rendered, "get$ref"),
+	)
+
+	assert.Contains(t, rendered.String(), "Pascalize=GetDollarRef\n")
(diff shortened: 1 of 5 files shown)


generator/template_repo_test.go
```diff
@@ -672,3 +672,33 @@ func TestGt0(t *testing.T) {
...
+
+	require.NoError(t,
+		templates.AddFile("functpl", tpl),
+	)
+
+	compiled, err := templates.Get("functpl")
+	require.NoError(t, err)
+
+	/*
+		opts := opts()
```
youyuanwu, line 690: comments intentionally kept?

fredbi: ah oops :)
good catch sir

fredbi: fixed. Thanks @youyuanwu

PR #3029, opened 2023-12-27
test(client): added dynamic test to assert path params

### LINKED ISSUE #1083: escaped parameters fail to generate the correct url path when a base path is present.

fredbi:
Added a test that constructs a generated client, dynamic untyped server and exercise the stack (in this case, asserting how an escaped path parameter moves across the stack from runtime client to runtime server).

* fixes #1083 (asserts was already working)

### DIFF
--- a/fixtures/bugs/1083/pathparam_test.go
+++ b/fixtures/bugs/1083/pathparam_test.go
@@ -1,0 +1,102 @@
+//go:build ignore
+
+package main
+
+// this test is designed to be run dynamically by go-swagger test suite
+// (generator.generate_test.go), after a test client has been generated.
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/go-openapi/errors"
+	"github.com/go-openapi/loads"
+	"github.com/go-openapi/runtime"
+	rtclient "github.com/go-openapi/runtime/client"
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/go-openapi/runtime/middleware/untyped"
+	"github.com/go-openapi/swag"
+	"github.com/go-swagger/go-swagger/fixtures/bugs/1083/codegen/client"
+	"github.com/go-swagger/go-swagger/fixtures/bugs/1083/codegen/client/pet_operations"
+	"github.com/go-swagger/go-swagger/fixtures/bugs/1083/codegen/models"
+	"github.com/stretchr/testify/require"
+)
+
+type operationHandler struct{}
+
+func (h *operationHandler) Handle(in interface{}) (interface{}, error) {
+	// this handler sends back the input path parameter in Name
+
+	paramsMap, ok := in.(map[string]interface{})
+	if !ok {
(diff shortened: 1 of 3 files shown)


fredbi:
This issue was so old... Couldn't believe it wasn't working... Now some evidence it does.

youyuanwu (approved):

PR #3037, opened 2023-12-31
chore(test): reduced the verbosity of tests

fredbi:
This PR generalizes the discarding of log output.

A few rough edges remain (but were previously there), as it is still possible that the few tests that assert the logged output cross the way of a test that discards it.

At least, running tests now generates less clutter.

### DIFF
--- a/generator/build_test.go
+++ b/generator/build_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	flags "github.com/jessevdk/go-flags"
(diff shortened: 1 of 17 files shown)


generator/build_test.go
```diff
@@ -103,3 +99,27 @@ func goExecInDir(t testing.TB, target string, args ...string) {
 	p, err := cmd.CombinedOutput()
 	require.NoErrorf(t, err, "unexpected error: %s: %v\n%s", cmd.String(), err, string(p))
 }
+
+var lockLogger sync.Mutex
+
+func discardOutput(w ...io.Writer) func() {
```
youyuanwu, line 105: Can this be merged with the one in utils_test.go?

youyuanwu (approved):

generator/build_test.go
fredbi: OK. Since these are 2 distinct packages, then I need to factorize inside some internal module. Like so: https://github.com/go-swagger/go-swagger/pull/3037/commits/c1bca457fbeecff841ad8b91fd559be6579f7c38

youyuanwu (approved):

PR #3038, opened 2023-12-31
chore(go build): removed code that requires go < 1.19 to build

fredbi:
Time to simplify a bit the code base:

* removed all pre-go1.19 specifics
* removed the pre-go1.11 "scan" package


cmd/swagger/commands/generate/spec.go
```diff
@@ -1,20 +1,3 @@
-//go:build !go1.11
-// +build !go1.11
-
-// Copyright 2015 go-swagger maintainers
```
youyuanwu, line 4: Keep licensing?

cmd/swagger/commands/generate/spec.go
```diff
@@ -23,16 +6,17 @@ import (
...
 	"github.com/go-openapi/spec"
-	"github.com/go-swagger/go-swagger/scan"
 	"github.com/jessevdk/go-flags"
 	"gopkg.in/yaml.v3"
 )
 
 // SpecFile command to generate a swagger spec from a go application
 type SpecFile struct {
-	BasePath    string         `long:"base-path" short:"b" description:"the base path to use" default:"."`
+	WorkDir     string         `long:"work-dir" short:"w" description:"the base path to use" default:"."`
```
youyuanwu, line 19: Is this backward compatible?

youyuanwu (approved):

cmd/swagger/commands/generate/spec.go
fredbi: ok. Good point. The newer version omitted the licensing terms. Fixed

cmd/swagger/commands/generate/spec.go
fredbi: No it's not. Options changed when migration rom "scan" to "codescan" 5 years ago. BasePath has not been in use since the post-go1.11 days (5 years ago?). Leaving it without.

PR #3041, opened 2024-01-03
fix(codegen): fixed calling order for WithAutoXOrder

### LINKED ISSUE #2949: Use of --keep-spec-order leads to empty response structs in v0.30.5
## Problem statement

First off, thanks for `go-swagger`, it's a great tool and I'm sure it's not an easy one to maintain!

I recently tried upgrading from `go-swagger` v0.30.4 to v0.30.5 and noticed that some of the generated response structs are now empty.  I've tried to reduce it down to a minimal reproducible example below.

## Swagger specification

```
swagger: '2.0'
info:
  title: Users API
  version: 0.0.1

paths:
  /api/v1/users/:
    get:
      responses:
        200:
          $ref: '#/responses/Users'

responses:
  Users:
    description: List of users
    schema:
      allOf:
        - $ref: '#/definitions/APIResponse'
        - properties:
            data:
              $ref: '#/definitions/Users'

definitions:
  APIResponse:
    type: object
    properties:
      timestamp:
        type: string
  Users:
    type: array
    items:
      $ref: '#/definitions/User'
    x-omitempty: false
  User:
    type: object
    properties:
      id:
        type: integer
      admin:
        type: boolean
      name:
        type: string
```

## Steps to reproduce

Put the above spec in a file `swagger.yaml` and then run the following in the same directory as the `swagger.yaml` file using a v0.30.4 `go-swagger` binary:

```
rm -rf gen/client
mkdir -p gen/client
swagger-v0.30.4 generate client --target gen/client --name Users --spec swagger.yaml --name Users --client-package users --keep-spec-order
```

open the file `gen/client/users/operations/get_api_v1_users_responses.go` and note that the `GetAPIV1UsersOK` struct contains a `Payload` field:

```go
type GetAPIV1UsersOK struct {
        Payload *GetAPIV1UsersOKBody
}
```

Now, run the same command with a v0.30.5 `go-swagger` binary:

```
rm -rf gen/client
mkdir -p gen/client
swagger-v0.30.5 generate client --target gen/client --name Users --spec swagger.yaml --name Users --client-package users --keep-spec-order
```

and open the same `gen/client/users/operations/get_api_v1_users_responses.go` file and note that the `GetAPIV1UsersOK` struct is now empty:

```go
type GetAPIV1UsersOK struct {
}
```

Finally, run the command with a v0.30.5 `go-swagger` binary but omit the `--keep-spec-order` option:

```
rm -rf gen/client
mkdir -p gen/client
swagger-v0.30.5 generate client --target gen/client --name Users --spec swagger.yaml --name Users --client-package users
```

open the file `gen/client/users/operations/get_api_v1_users_responses.go` and note that the `GetAPIV1UsersOK` struct once again has a `Payload` field:

```go
type GetAPIV1UsersOK struct {
        Payload *GetAPIV1UsersOKBody
}
```

I am not by any means a Swagger expert but this behavior was very surprising to me.  Is the spec I'm using invalid or is this a regression of some sort on the part of `go-swagger`?  Please let me know if I can supply any more information that would be helpful.

## Environment
swagger version: 0.30.5
go version: 1.20
OS: CentOS 7

fredbi:
This PR moves the call to WithAutoXOrder _before_ we validate and flatten the spec for codegen purpose.

* fixes #2949

Other changes:
* layout: moved all the spec massaging to a dedicated source
* allowed WithAutoXOrder to report about errors rather than panicking

Overall, WithAutoXOrder is simplistic and could be improved a lot:
* supporting JSON
* supporting remote $refs, circular $refs
* removing the dependency to yaml.v2
* not producing a temporary file that hangs around
* etc

These improvements are beyond the scope of this PR, which is just to fix responses. Perhaps a follow up could bring a more general solution to this problem.


fredbi:
@casualjim  @youyuanwu  I realize that a lot of work about "keeping things ordered" has already been accomplished in swag.

I believe this is the way to solve this issue (and many other unreported ones) for good.
IMHO, the "WithAutoXOrder()" hack in go-swagger should disappear and be replaced by appropriate spec loaders (-> in "loads") that inject the "x-order" when loading each document (local or remote, YAML or JSON).
Working on that now. Looks like there is quite a loaded plate...

PR #3049, opened 2024-01-06
ci: setup ossf scorecard and codql workflows

mmorel-35:
CodeQL and OSSF scorecard are both complementary tools useful to detect vulnerabilities.
They also provide help on how to fix them.
There is also a score to publish on the readme.md

This PR includes the pin of GitHub Action and Docker images .
It also provides dependabot config for them as it is not easy to keep track of those hash manually. 
Also using an updater is preconised by ossf.
I didn't configured it for go mod as it is the object of another PR

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,48 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: '31 21 * * 4'
+
+permissions:
+  contents: read
+
+jobs:
+  analyze:
+    name: Analyze.
+    runs-on: ubuntu-latest
+    timeout-minutes: 360
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@012739e5082ff0c22ca6d6ab32e07c36df03c4a4 # v3.22.12
+      with:
+        languages: ${{ matrix.language }}
+
+
+    # Autobuild attempts to build any compiled languages (C/C++, C#, Go, Java, or Swift).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@012739e5082ff0c22ca6d6ab32e07c36df03c4a4 # v3.22.12
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@012739e5082ff0c22ca6d6ab32e07c36df03c4a4 # v3.22.12
+      with:
+        category: "/language:${{matrix.language}}"
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,55 @@
+name: Scorecard supply-chain security
+
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '44 2 * * 1'
+  push:
+    branches: [ "master" ]
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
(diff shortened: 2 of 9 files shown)


.github/workflows/test.yaml
```diff
@@ -21,14 +21,14 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-go@v4
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
```
fredbi, line 24: why do we need to pin commits like this?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,48 @@
+name: "CodeQL"
```
fredbi, line 1: how is this different from the default settings we may enable for the  repository?

.github/workflows/test.yaml
mmorel-35: There is a [nice article](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) about this. Here are the three main reasons they give

>  They guarantee that checking and deployment are performed with the same software, minimizing deployment risks, simplifying debugging, and enabling reproducibility.

> They can help mitigate compromised dependencies from compromising the project’s security. By evaluating the pinned dependency and being confident that it is not compromised, you can prevent the use of a later version that may be compromised.

> Pinned dependencies are a way to counter dependency confusion (or substitution) attacks. In these attacks, an application uses multiple feeds to acquire software packages (a “hybrid configuration”), and attackers trick the user into using a malicious package from an unexpected feed.

.github/workflows/codeql.yml
mmorel-35: This is maintable . You can also customize it.

.github/workflows/test.yaml
fredbi: I am not referring to the dependencies in the build, but to the pinning of github action versions. We don't want to maintain the pinning of such gory details.  Why is that required?

mmorel-35: I'm referring to GitHub actions. You won't maintain them. Dependabot can

fredbi: I've read the article. I understand better now. My concern is that we have  13+ repos to maintain.

mmorel-35: I guess you're worried about the number of PR you'll have to validate?

You can change the frequency and also group the dependencies to reduce this number.

mmorel-35: Can you imagine give this a try for a certain time. Identify and fix the irritant if there are and roll it back if it's a fail

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,55 @@
+name: Scorecard supply-chain security
```
fredbi, line 1: can we show some kind of badge in the readme for this scorecard?

.github/workflows/test.yaml
fredbi: Yes that could be possible. What is your own feedback about using this technique in other projects?

README.md
```diff
@@ -7,6 +7,7 @@
 [![FOSSA Status](https://app.fossa.io/api/projects/git%2Bgithub.com%2Fgo-swagger%2Fgo-swagger.svg?type=shield)](https://app.fossa.io/projects/git%2Bgithub.com%2Fgo-swagger%2Fgo-swagger?ref=badge_shield)
 [![GolangCI](https://golangci.com/badges/github.com/go-swagger/go-swagger.svg)](https://golangci.com)
 [![Go Report Card](https://goreportcard.com/badge/github.com/go-swagger/go-swagger)](https://goreportcard.com/report/github.com/go-swagger/go-swagger)
+[![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/go-swagger/go-swagger/badge)](https://securityscorecards.dev/viewer/?uri=github.com/go-swagger/go-swagger)
```
mmorel-35, line 10: @fredbi ,
The badge is here ;)

.github/workflows/scorecard.yml
mmorel-35: Done ;)

.github/workflows/test.yaml
mmorel-35: It quite practical and flexible enough to fit with a variety of practice.
Some teams prefer daily checking some other monthly depending on the frequency they publish release or follow another logic. 

Depending on the amount of dependencies you've got in the project it is some time preferable to use renovate which is way more configurable than dependabot.

But for most of the use case dependabot is good enough.

As it creates a PR, it executes your pipeline with all your tests. Most of the time you just click on merge because there is nothing to fix.

You can can even have a workflow that automatically validate those merge requests if it passes the pipeline.

fredbi: ok let's give at try. some other contributors have pushed forward to use dependabot in our other repos.
I am fine with that but I think I'll reduce the frequency or configure some kind . auto merge

fredbi (approved):

casualjim:
can we revert this. I don't agree with the pinning philosophy. And I also think there is limited benefit to dependabot on this repo. In my experience it is very noisy 

I understand the perceived benefits, but I think that's for people who like busy work. 

We do not want to pin, we want to always move with the latest when possible, and fix things as they break. It's just a different way of looking at the problem, but pinning is a major security risk so we don't want to do that.

PR #3055, opened 2024-01-06
Use modules instead of incompatibles

mmorel-35:
Hi @fredbi ,
This uses modules instead of incompatibles versions of oidc and jwt dependencies. 

It also provides gomod config for dependabot  if you want.

### DIFF
--- a/examples/composed-auth/auth/authorizers.go
+++ b/examples/composed-auth/auth/authorizers.go
@@ -6,7 +6,7 @@ import (
 
 	errors "github.com/go-openapi/errors"
 	models "github.com/go-swagger/go-swagger/examples/composed-auth/models"
-	jwt "github.com/golang-jwt/jwt"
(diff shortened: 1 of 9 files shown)


examples/composed-auth/auth/authorizers.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	errors "github.com/go-openapi/errors"
 	models "github.com/go-swagger/go-swagger/examples/composed-auth/models"
-	jwt "github.com/golang-jwt/jwt"
+	jwt "github.com/golang-jwt/jwt/v5"
```
fredbi, line 9: thanks. that's fine. Did you build the examples? Examples are currently not covered by our CI tests.

mmorel-35: I didn’t built them, just updated the doc.

PR #3056, opened 2024-01-06
Revert "ci: setup ossf scorecard and codql workflows"

casualjim:
Reverts go-swagger/go-swagger#3049


fredbi:
@casualjim I am proposing to take care of this by reverting more selectively.

1. tame dependabot to avoid the overwhelming flurry of PRs
    * monthly scans, automatic approve&merge with minimal CI when only github actions are involved
 2. remove the score card. The results cannot be viewed anyway (the link under the badge does not work)
 3. keep the github code scan (at least for some time). It is always interesting to watch if something interesting comes out of it

Overall we agree on principle: let's not allow this thing to become a work generator.

fredbi:
@casualjim  I've resolved the merge conflict and re-added 2 things that I've found interesting to continue experimenting with: the codeSQL scan and the "security score cards" (a kind of linter for "opsec best practices). The report is eventually available (it took a day to set up, but now it's there) and the findings, if not necessarily relevant to our use-case, are interesting to analyze and be reflected upon.

https://securityscorecards.dev/viewer/?uri=github.com/go-swagger/go-swagger

Maybe we'll eventually decide to disable this scorecard if it turns out to be really not suited to our project, but for the moment  I am still curious about the recommendations.

I've removed dependabot and the by-commit-sha pinning for now. I think that @mmorel-35  raised an interesting point, but the PR did not come with a complete solution, just more problems to solve in the short term and more work until we find an appropriate solution to maintain our dependencies with minimal burden. So, _statu quo_ for now, I need more experimentation on my fork.

@mmorel-35  I have to thank you for having raised an interesting point. However yesterday's experiment proved not to be very successful: as soon as the stuff was merged, we already had about 5 such update PRs, stalling all our CI for hours.

This started a fascinating conversation with @casualjim about the extra workload that comes from all those update bots...

If we eventually move to such kind of automation, it will have to be more carefully designed and discussed beforehand.

Another contributor has suggested a dependabot setup to some of the go-openapi repos. The problem to solve for those repos is different, but still, we don't want to use many different tools.

fredbi (approved):

fredbi:
NOTE: I've updated those damned actions, so the bot PR's should disappear automatically.

.github/workflows/test.yaml
```diff
@@ -230,11 +221,8 @@ jobs:
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-
-      - name: Install Go
-        uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
```
mmorel-35, line 225: Are you removing names too ?

fredbi: yes. I think it is explicit enough. Any thought?

mmorel-35: No, just ensure it’s deliberate

PR #3057, opened 2024-01-07
chore: updated dependencies for examples

fredbi:
* updated dependencies for oauth2 examples: oidc and jwt new versions
  came with breaking changes
* ci: added build check for examples

> NOTE: initially, I wanted to reduce the set of main dependencies by making the "examples" folder
>  it own module. Eventually, this did not change much (apart oidc and jwt)
> and was adding more maintainance burden. Given up on that.


examples/go.mod
```diff
@@ -0,0 +1,67 @@
+module github.com/go-swagger/go-swagger/examples
+
+go 1.21.4
```
mmorel-35, line 3: You might want align the go versions between root and examples

fredbi: yes sure. This is just a draft I have to make sure other things are not disturbed first (like go gettability etc)

.github/workflows/test.yaml
```diff
@@ -58,6 +58,12 @@ jobs:
         run: |
           go install ./cmd/swagger
 
+      - name: Verify that examples build
```
mmorel-35, line 61: Don't you think it could deserve it's own job in the workflow with a dedicated step for each these commands ?

fredbi: I don't know. This would require a "workflow_run" trigger to operate after the previous steps, but yes, essentually why not? If you feel strongly about that, please propose a follow-up PR.

youyuanwu (approved):

PR #3058, opened 2024-01-07
Update docs: choosing a principal

tikhonfedulovv:
Principal can be specified by the package path (i.e. including module name). I have not found a description of this feature anywhere.


fredbi (approved):

tikhonfedulovv:
> Thank you. That's indeed a feature.

Yeah, I was surprised when I found this.

Some time ago, I tried to specify the full path to file with principal, which didn't work. I think this should be added to the documentation.

fredbi (approved):

tikhonfedulovv:
@fredbi what about updating docs on goswagger.io? There are still no changes

fredbi:
Yes you’re correct
We have a clumsy process to update the site.
I’ve been working on a branch to fully automate that part from CI, but couldn’t complet it so far. Thanks for reminding me this overdue task… :)

Issue #2649, opened 2021-10-13
Swagger generate is using a very large amount of RAM


sodul:
## Problem statement

We have a swagger.json file that has grown to 2.2MB over time and now calling `swagger generate` is causing the process to use close to 4GB of RAM. Just over a year ago it was under 2GB. We have been allocating more RAM to our containers to mitigate the issue but we are concerns that the process seems to be using a lot more RAM than we would expect. The `swagger validate` command stays under 500MB and this is not a problem.

Unfortunately we cannot share our swagger.json file but were wondering if others experience the same problem and how to mitigate it.

## Swagger specification
N/A

## Steps to reproduce
swagger generate spec -o swagger.json

## Environment
swagger version: v0.27.0
go version: 1.17.2
OS: Ubuntu 20.04 under Docker.

ynakao:
Maybe I have the same issue. When I build [GoToSocial](https://github.com/superseriousbusiness/gotosocial) package using [AUR PGKBUILD](https://aur.archlinux.org/packages/gotosocial) on my VPS which has 2GB RAM and 4GB swap, build failed at `swagger generate spec` process due to out of RAM.

I monitored resource usage during the process, CPU usage went up to over 100% by kswapd0 and RAM usage also went up to 100%. After that, my VPS didn't accept any keyboard inputs and tmux session crashed. Also, no swap space was used as far as I can tell.

Therefore, I tried executing the swagger process alone on my local machine which has plenty of RAM(16GB), and inspected memory usage by `time` command. The result is pasted below and according to it, it seems swagger process consumed about 2.6GB memory. Considering the output swagger.yaml size is about 3500LoC or 140KB, it is a bit resource intensive, I think.

Is this a normal behavior or go-swagger problem? If it's implementation problem on GoToSocial, I'll open another issue at their issue tracker.

```shell
$ /usr/bin/time --verbose swagger generate spec --scan-models --exclude-deps -o web/assets/swagger.yaml
        Command being timed: "swagger generate spec --scan-models --exclude-deps -o web/assets/swagger.yaml"
        User time (seconds): 22.13
        System time (seconds): 2.86
        Percent of CPU this job got: 164%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:15.15
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 2644364
[...]

PR #3063, opened 2024-01-20
perf(codegen): reduced allocated memory

fredbi:
* contributes to #2649

This PR alleviates the need to duplicate specs across operations: we only need to save the state of the (mutating) spec once.


fredbi:
The impact of this single change is visible on the codegen_test CI run, with timings for the whole codegen test suite down from about 20min to 15.

youyuanwu (approved):

PR #3068, opened 2024-01-28
golangci-lint: enable testifylint linter

mmorel-35:
Helps follow best practices with testify use.


fredbi:
LGTM. Great.

fredbi (approved):

fredbi:
We have an issue with tokenless access to codecov because of rate limits on public repos, so that's the cause of CI failing. Running again...

youyuanwu (approved):

PR #3072, opened 2024-02-20
Add riscv64 & loong64 for goreleaser

znley:


fredbi:
@znley I don't think we actually use goreleaser for our release process (a pity I'd say, but that's the way it is now).
 believe tha the goreleaser config is a remnant of a past attempt to use goreleaser.

znley:
@fredbi 
Now  I get it! So what method is used now for release?

fredbi:
I believe that is the tool used by @casualjim  to cut releases: https://github.com/go-openapi/github-release

znley:
@fredbi @casualjim 
The  [github-release](https://github.com/go-openapi/github-release) seems to be used for upload binaries and [gorelease](https://github.com/goreleaser/goreleaser) is used for building, is right?

fredbi:
There is no way we can set up a CI or a build chain for loong64.

PR #3078, opened 2024-03-04
ci: reenacted codecov secret token

fredbi:
With codecov action v4, the "tokenless" functionality works but is somewhat deprecated.

With codecov action v3 or v4, even with tokenless support, our CI sometimes fails because codecov hits a rate limit on github API.

The safest way to securely send our test coverage without failures is to configure the token as a github action secret and use that token to connect to codecov.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,7 +10,7 @@ on:
     branches:
       - master
 
-  pull_request:
+  pull_request_target:
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


casualjim (approved):

fredbi:
@casualjim I've discovered that this doesn't work (yet). There are problems with forwarding repo secrets to PRs from a fork. Still investigating this stuff...

fredbi:
Need to check if events are properly fired when the workflow is already on the default master branch.

fredbi:
> Need to check if events are properly fired when the workflow is already on the default master branch.

Validated. Replaced the workflow trigger "pull_request" by "pull_request_target" so secrets are allowed in the action, even for PR's originated from a fork. Now the codecov CLI receives the configured token and it no longer has to call the flaky github API.

PR #3099, opened 2024-04-05
Adding support for s390x

dilipgb:


dilipgb:
@fredbi @casualjim please review this PR. I'm trying to enable go-swagger package for s390x platform.

hack/doc-site/hugo/github.json
casualjim, line 1: I think this file is autogenerated

casualjim (approved):

hack/doc-site/hugo/github.json
fredbi: @casualjim no it's not :) not yet at least. I've refreshed the install docs lately.

fredbi (approved):

PR #3105, opened 2024-04-21
chore: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


fredbi (approved):
Nice. Thank you for this contribution

PR #3128, opened 2024-08-12
Introduce a new command-line parameter for "generate spec"

zenovich:
Introduce a new command-line parameter for "generate spec" allowing to set "x-nullable: true" by default for each Go field having a pointer type and not having JSON 'omitempty' option set.

With this new feature, we can skip unnecessary annotations for pointer-type fields as they are almost always nullable. So instead of

```
// swagger:model Item
type Item struct {
	// Extensions:
	// ---
	// x-nullable: true
	Value *int64 `json:"value"`
}
```
we can simply write
```
// swagger:model Item
type Item struct {
	Value *int64 `json:"value"`
}
```
with the same effect when running
```
swagger generate spec --scan-models --nullable-pointers -o ./swagger.yaml && swagger validate ./swagger.yaml
```
instead of usual
```
swagger generate spec --scan-models -o ./swagger.yaml && swagger validate ./swagger.yaml
```

Although rarely needed, it is still possible to mark a field as non-nullable explicitly:
```
// swagger:model Item
type Item struct {
	// Extensions:
	// ---
	// x-nullable: false
	Value *int64 `json:"value"`
}
```

Also, `x-nullable: true` will not be set automatically for fields having the JSON 'omitempty' option set:
```
// swagger:model Item
type Item struct {
	Value *int64 `json:"value,omitempty"`
}
```

It **works** even for fields pointing to **embedded structures**.

Then, it's possible to convert the generated spec to OpenAPI 3.0 preserving the `x-nullable` tags and converting them to `nullable` tags:
```
swagger2openapi --refSiblings allOf --yaml swagger.yaml | sed 's/x-nullable:/nullable:/g' > openapi3.yaml
```

and preview the documentation:
```
redocly preview-docs openapi3.yaml
```

and build the documentation:
```
redocly build-docs openapi3.yaml
```

The field in the documentation will look as 
Screenshot 2024-08-12 at 17 33 33


casualjim (approved):

fredbi (approved):

fredbi:
Merged after rebase as #3185 
Weird that this PR could not be rebased as usual from github directly.

zenovich:
Thanks! 👍🏻

Issue #3131, opened 2024-09-02
Go v1.23.0 `panic: WARNING: can't determine refined type any (*types.Alias)`


tsmethurst:
## Problem statement

Hello hello, I can't seem to get go swagger generating a yaml output from code annotations on Go v1.23.0, where it was working fine in Go v1.22.x.

When running the following command in the GoToSocial repo with go bumped to v1.23.0:

`go run github.com/go-swagger/go-swagger/cmd/swagger generate spec --scan-models --exclude-deps --output docs/api/swagger.yaml`

The following panic occurs:

```
panic: WARNING: can't determine refined type any (*types.Alias)

goroutine 1 [running]:
github.com/go-swagger/go-swagger/codescan.(*schemaBuilder).buildFromType(0xc018e6faf8, {0x14ca240, 0xc0000b8dc0}, {0x14d30f8, 0xc052567b20})
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/go-swagger/go-swagger/codescan/schema.go:471 +0xa5e
github.com/go-swagger/go-swagger/codescan.(*schemaBuilder).buildFromType(0xc018e6faf8, {0x14c90c0, 0xc021eec720}, {0x14d30f8, 0xc052567940})
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/go-swagger/go-swagger/codescan/schema.go:315 +0x8a7
github.com/go-swagger/go-swagger/codescan.(*schemaBuilder).buildFromType(0xc018e6faf8, {0x14c8fa8, 0xc021eec6e0}, {0x14d30f8, 0xc052567940})
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/go-swagger/go-swagger/codescan/schema.go:309 +0x7ff
github.com/go-swagger/go-swagger/codescan.(*schemaBuilder).buildFromStruct(0xc018e6faf8, 0xc0548d5ac0, 0xc012ad8720, 0xc053212008, 0xc018e6f4d0)
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/go-swagger/go-swagger/codescan/schema.go:832 +0xd25
github.com/go-swagger/go-swagger/codescan.(*schemaBuilder).buildFromType(0xc018e6faf8, {0x14c8ff8, 0xc012ad8720}, {0x14d30f8, 0xc052566c30})
[...]

PR #3137, opened 2024-10-13
Ignore types.Alias to avoid panics

aeneasr:
This PR adds a no-op for `types.Alias`, which prevents the panics, see #3131.

Unfortunately, `types.Alias` are no longer supported by `go-swagger`, which would require a proper fix.


fredbi:
TODO(fred): upgrade go-swagger to go1.22 first

PR #3145, opened 2024-12-12
chore: use errors.New to replace fmt.Errorf with no parameters

RiceChuan:
use errors.New to replace fmt.Errorf with no parameters


fredbi (changes requested):
The change should affect the template, not the generated code.
On top of that, linters dogma has evolved over time and I am not longer sure that recent linting config like errors.New any longer...

PR #3156, opened 2025-02-12
fix(CI/tests): checkout the PR commits, not master

nikolaiianchuk:
The CI checkout in pull requests was checking out the master branch, which made it easy to miss the issues in tests and to push changes with test issues to master, because these issues were not seen in the PR

I added
```
        with:
           ref: ${{ github.head_ref }}
```
into every job that is lint/test/build related, not sure if we need it in the docker jobs

I will address the issues with tests in the next PR

cc @casualjim

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,6 +32,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


nikolaiianchuk:
this PR's CI will fail, as I haven't fixed any tests yet, those will come in a separate PR

PR #3158, opened 2025-02-12
fix(ci): checkout action fix & move out pr pipeline

nikolaiianchuk:
Second try, I'm moving out the PR action in it's own file and adding the solution for forks from https://github.com/actions/checkout/issues/551

The reason the previous one is failing is the fact that the commit from the fork doesn't actually exist in this repo

Reverted the changes from the previous commit in `test.yaml` that from now on should be only triggered on master/tag builds

@casualjim let's try this again, hopefully it works this time around.
And thank you for helping out with the merges

### DIFF
--- a/.github/workflows/test-pr.yaml
+++ b/.github/workflows/test-pr.yaml
@@ -1,0 +1,200 @@
+name: Run CI (PR)
+
+env:
+  GOCOVMODE: atomic
+
+on:
+  pull_request_target:
+
+#  pull_request:
+#    paths-ignore:
+#    - docs/*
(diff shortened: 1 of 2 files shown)


casualjim (approved):

nikolaiianchuk:
Tried to break things in PRs, works well enough. There's an issue with the linter annotations (they do not show up in `Files changed` tab for some reason), I'll look at that tomorrow
https://github.com/go-swagger/go-swagger/pull/3159
https://github.com/go-swagger/go-swagger/pull/3158

PR #3163, opened 2025-02-13
fix(ci/update-doc): pin hugo-book version

nikolaiianchuk:
```
WARN  Module "hugo-book" is not compatible with this Hugo version: Min 0.128.0; run "hugo mod graph" for more information.
Start building sites … 
```

We were checking out the master branch, I set it to v9 tag as v10 needs hugo `v0.124.0` to work and we have `v0.123.8`. See the tags here: https://github.com/alex-shpak/hugo-book/releases

Can't test it on my fork, but I think this should fix it. 

Also, couldn't find any logs as to which commit was it using because the last successful job ran around a year ago, and the logs are no longer present there

### DIFF
--- a/.github/workflows/update-doc.yaml
+++ b/.github/workflows/update-doc.yaml
@@ -50,7 +50,7 @@ jobs:
       -
         name: Initialize themes
         run: |
-          (cd hack/doc-site/hugo/ && git clone https://github.com/alex-shpak/hugo-book themes/hugo-book)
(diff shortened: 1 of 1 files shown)


nikolaiianchuk:
Failure in the build-doc check is expected, as it's using the workflow config from the master branch

nikolaiianchuk:
I think I've come up with something better

PR #3167, opened 2025-02-14
Ci security fix break test

nikolaiianchuk:
To be closed after the CI runs


nikolaiianchuk:
pull-request target fails as expected

PR #3135, opened 2024-09-24
fix the template layout reference

emmanuel-ferdman:
# PR Summary
Commit b44601d48a17c24dc821681ded1b6465ddb5f918 moved the location of the `template_layout.md` file. This PR adjusts sources to changes.

PR #3168, opened 2025-02-15
fix(ci/update-doc): pin hugo themes version + improvements

nikolaiianchuk:


nikolaiianchuk:
https://github.com/go-swagger/go-swagger/actions/runs/13344381895/job/37273031159?pr=3168
See this CI check, that's the fixed one

nikolaiianchuk:
@casualjim 
This should unblock #3135 and any other PRs containing changes to the docs + should fix the build-doc pipeline in the master branch

.github/workflows/update-doc.yaml
```diff
@@ -10,16 +10,14 @@ on:
     - hack/doc-site/**
     - .github/workflows/update-doc.yaml
 
-  pull_request_target:
+  pull_request:
```
nikolaiianchuk, line 13: This change is done to make it easier to change and test the workflow in the PR's (for example now, when I need to show that this change actually works without merging it to master)

The `pull_request_target` workflows always run with the yaml from the master branch, the `pull_request` ones use the ones from the PR

casualjim (approved):

PR #3186, opened 2025-04-10
Migrate to github.com/go-viper/mapstructure/v2

tklauser:
The github.com/mitchellh/mapstructure package is archived and no longer receives updates[^1]. The "blessed" fork[^2] is
github.com/go-viper/mapstructure. The API is the same, so nothing needs to be changed apart from the import path.

[^1]: https://github.com/mitchellh/mapstructure/issues/349
[^2]: https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc


fredbi (approved):
Thanks

Issue #3176, opened 2025-03-07
Can't configure TemplateOpts.SkipFormat, maybe.


seiyab:
`TemplateOpts` has `SkipFormat` Field.
https://github.com/go-swagger/go-swagger/blob/1e29a3d03ec7ea3e53aad355498996c092f90144/generator/shared.go#L283
However, it looks set nowhere.
```
[golang]user@projects/go-swagger [master] % git grep SkipFormat
generator/shared.go:    SkipFormat bool   `mapstructure:"skip_format"`
generator/shared.go:    if !t.SkipFormat {
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: true,
generator/shared_test.go:               SkipFormat: true,
generator/shared_test.go:               SkipFormat: false,
```

Probably it should be able to configure by command flag?

Adding context, I found it is bottleneck in our project to formatting code. I'm wondering whether we can skip formatting because we rarely matter format in generated code. Here is flamegraph.

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/0f3300ab-b8e1-4bb4-8970-4c97da2ff7d5" />

fredbi:
The time consuming part is actually resolving the go-import part, which is required.

seiyab:
I thought go-import is used for formatting which is optional.  Maybe I'm misunderstanding something so will investigate it.

seiyab:
Ah, maybe I understand.
Generating code with `SkipFormat: true` generates code about 2x faster. However, generated code contains unused imports so doesn't compile. So `SkipFormat` can't be true.
Is it for test, or unnecessary legacy? May I add comments or remove it?

seiyab:
Maybe we need to improve how to collect `.Imports` to fix it?
https://github.com/go-swagger/go-swagger/blob/231e896047b5dc5423e7df5ed8834fcd303f11e1/generator/model.go#L379

https://github.com/go-swagger/go-swagger/blob/4a9fb2b500b6e2643cef9f43baebada5eec10d0a/generator/support.go#L268-L280

fredbi:
@seiyab the main problem is to guess which packages any given generated source file will import, as not all generated files need the same stuff.

[...]

PR #3189, opened 2025-04-22
add comment for SkipFormat

seiyab:
Related issue: #3176

### DIFF
--- a/generator/shared.go
+++ b/generator/shared.go
@@ -280,7 +280,7 @@ type TemplateOpts struct {
 	Target     string `mapstructure:"target"`
 	FileName   string `mapstructure:"file_name"`
 	SkipExists bool   `mapstructure:"skip_exists"`
-	SkipFormat bool   `mapstructure:"skip_format"`
+	SkipFormat bool   `mapstructure:"skip_format"` // not a feature, but for debugging. generated code before formatting might not work because of unused imports.
 }
 
 // SectionOpts allows for specifying options to customize the templates used for generation


generator/shared.go
```diff
@@ -280,7 +280,7 @@ type TemplateOpts struct {
 	Target     string `mapstructure:"target"`
 	FileName   string `mapstructure:"file_name"`
 	SkipExists bool   `mapstructure:"skip_exists"`
-	SkipFormat bool   `mapstructure:"skip_format"`
+	SkipFormat bool   `mapstructure:"skip_format"` // not a feature, but for debugging. generated code before formatting might not work because of unused imports.
```
fredbi, line 283: Maybe this should be displayed in the help text for the command.

fredbi (changes requested):
Could you please add a similar not in the help text for the command line option?

generator/shared.go
seiyab: Commands doesn't look to have a `skip-format` option. Do you mean to add the option and help command? Or display anywhere anyway?

```
[golang]user@projects/go-swagger [3176-add-comment-for-SkipFormat] % git grep skip-format
[golang]user@projects/go-swagger [3176-add-comment-for-SkipFormat] % echo $?
1
```

```
[golang]user@projects/go-swagger [3176-add-comment-for-SkipFormat] % git grep SkipFormat
generator/shared.go:    SkipFormat bool   `mapstructure:"skip_format"` // not a feature, but for debugging. generated code before formatting mi
ght not work because of unused imports.
generator/shared.go:    if !t.SkipFormat {
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: false,
generator/shared_test.go:               SkipFormat: true,
generator/shared_test.go:               SkipFormat: true,
generator/shared_test.go:               SkipFormat: false,
```

```
[golang]user@projects/go-swagger [3176-add-comment-for-SkipFormat] % ~/projects/gopath/bin/swagger generate server --help
Usage:
  swagger [OPTIONS] generate server [server-OPTIONS]

generate all the files for a server application

Application Options:
  -q, --quiet                                                                                silence logs
      --log-output=LOG-FILE                                                                  redirect logs to file

Help Options:
  -h, --help                                                                                 Show this help message

[server command options]
      -s, --server-package=                                                                  the package to save the server specific code
                                                                                             (default: restapi)
          --main-package=                                                                    the location of the generated main. Defaults
                                                                                             to cmd/{name}-server
          --implementation-package=                                                          the location of the backend implementation of
                                                                                             the server, which will be autowired with api
      -P, --principal=                                                                       the model to use for the security principal
          --default-scheme=                                                                  the default scheme for this API (default:
                                                                                             http)
          --principal-is-interface                                                           the security principal provided is an
                                                                                             interface, not a struct
          --default-produces=                                                                the default mime type that API operations
                                                                                             produce (default: application/json)
          --default-consumes=                                                                the default mime type that API operations
                                                                                             consume (default: application/json)
          --skip-models                                                                      no models will be generated when this flag is
                                                                                             specified
          --skip-operations                                                                  no operations will be generated when this
                                                                                             flag is specified
          --skip-support                                                                     no supporting files will be generated when
                                                                                             this flag is specified
          --exclude-main                                                                     exclude main function, so just generate the
                                                                                             library
          --exclude-spec                                                                     don't embed the swagger specification
          --flag-strategy=[go-flags|pflag|flag]                                              the strategy to provide flags for the server
                                                                                             (default: go-flags)
          --compatibility-mode=[modern|intermediate]                                         the compatibility mode for the tls server
                                                                                             (default: modern)
          --regenerate-configureapi                                                          Force regeneration of configureapi.go
      -A, --name=                                                                            the name of the application, defaults to a
                                                                                             mangled value of info.title
          --with-context                                                                     handlers get a context as first arg
                                                                                             (deprecated)

    Options common to all code generation commands:
      -f, --spec=                                                                            the spec file to use (default
                                                                                             swagger.{json,yml,yaml})
      -t, --target=                                                                          the base directory for generating the files
                                                                                             (default: ./)
          --template=[stratoscale]                                                           load contributed templates
      -T, --template-dir=                                                                    alternative template override directory
      -C, --config-file=                                                                     configuration file to use for overriding
                                                                                             template options
      -r, --copyright-file=                                                                  copyright file used to add copyright header
          --additional-initialism=                                                           consecutive capitals that should be
                                                                                             considered intialisms
          --allow-template-override                                                          allows overriding protected templates
          --skip-validation                                                                  skips validation of spec prior to generation
          --dump-data                                                                        when present dumps the json for the template
                                                                                             generator instead of generating files
          --strict-responders                                                                Use strict type for the handler return value
          --with-expand                                                                      expands all $ref's in spec prior to
                                                                                             generation (shorthand to
                                                                                             --with-flatten=expand)
          --with-flatten=[minimal|full|expand|verbose|noverbose|remove-unused|keep-names]    flattens all $ref's in spec prior to
                                                                                             generation (default: minimal, verbose)
      -p, --template-plugin=                                                                 the template plugin to use

    Options for model generation:
      -m, --model-package=                                                                   the package to save the models (default:
                                                                                             models)
      -M, --model=                                                                           specify a model to include in generation,
                                                                                             repeat for multiple (defaults to all)
          --existing-models=                                                                 use pre-generated models e.g.
                                                                                             github.com/foobar/model
          --strict-additional-properties                                                     disallow extra properties when
                                                                                             additionalProperties is set to false
          --keep-spec-order                                                                  keep schema properties order identical to
                                                                                             spec file
          --struct-tags=                                                                     the struct tags to generate, repeat for
                                                                                             multiple (defaults to json)
          --rooted-error-path                                                                extends validation errors with the type name
                                                                                             instead of an empty path, in the case of
                                                                                             arrays and maps

    Options for operation generation:
      -O, --operation=                                                                       specify an operation to include, repeat for
                                                                                             multiple (defaults to all)
          --tags=                                                                            the tags to include, if not specified
                                                                                             defaults to all
      -a, --api-package=                                                                     the package to save the operations (default:
                                                                                             operations)
          --with-enum-ci                                                                     allow case-insensitive enumerations
          --skip-tag-packages                                                                skips the generation of tag-based operation
                                                                                             packages, resulting in a flat generation
```

fredbi: Yes you're correct. My bad. The PR is ok. thanks

fredbi (approved):

PR #3195, opened 2025-05-16
Own code format for concurrent safety

seiyab:
Discussion: #3190


generator/server_test.go
```diff
@@ -237,10 +237,11 @@ func TestServer_OperationGroups(t *testing.T) {
...
 		{
-			Name:     "opGroupTest",
-			Source:   "asset:opGroupTest",
-			Target:   "{{ joinFilePath .Target .Name }}",
-			FileName: "{{ (snakize (pascalize .Name)) }}_opgroup_test.gol",
+			Name:       "opGroupTest",
+			Source:     "asset:opGroupTest",
+			Target:     "{{ joinFilePath .Target .Name }}",
+			FileName:   "{{ (snakize (pascalize .Name)) }}_opgroup_test.gol",
+			SkipFormat: true,
```
seiyab, line 244: 【FYI】
If we want to enable format in this test, we also need following logic in `parseGoOrFragment`. I consider implementing it only for test don't make sense enough.
https://github.com/golang/tools/blob/v0.33.0/internal/imports/imports.go#L220-L246

seiyab:
I'm planning following changes. Ask me to stop if you feel I'm going a wrong way.
- Adding common libraries into import automatically on `formatGo()`
  - Candidates are standard libraries like `encoding/json` and go-openapi libraries like `github.com/go-openapi/runtime`.
    - These candidates will be hard coded in this repository.
- Sort and group imports as https://github.com/go-swagger/go-swagger/commit/944f58a1000a1ca65417bdabd1af9f1e775e35fe does.

seiyab:
I suddenly came up with concurrent safe approach. But automatic import fixing is still not straight forward.
I'm also proposing concurrent safe `LocalPrefix` option https://github.com/golang/go/issues/74486.

casualjim:
Looks like we'll need to get the golangci-lint to run again to see if this passes the build

fredbi:
@seiyab  I've finally found the time to thoroughly review your work.
I knew this would take time to test, compare performances etc.

I overall agree with your change proposal and can confirm there is a significant improvement.

I worked on a few additions to your PR in order to make the transition smoother:
1. Introduce the feature as optional
2. Adopt it in the codegen CI exerciser

you'll find my changes on top of our branch https://github.com/seiyab/go-swagger/pull/2

seiyab:
@fredbi Thank you for reviewing!
> To move forward, now please rebase and squash your commits.

Could you please let me know what level of commit granularity you’d prefer -- is just one big commit (with co-author information) acceptable?

fredbi:
> @fredbi Thank you for reviewing!
> 
> > To move forward, now please rebase and squash your commits.
> 
> Could you please let me know what level of commit granularity you’d prefer -- is just one big commit (with co-author information) acceptable?

Two commits in the end : one squashing your 40 or so commits and one for my contribution.

please reword the squashed commit details: when squashed, these long series of commit don’t look good any longer.

PR #3209, opened 2025-07-03
feature: return errors option

tikhonfedulov:
Return errors to handle it gracefully in **ServeError**


fredbi (approved):
ok. This could be a useful option. I think I understand the logic and the implementation is just fine.

LGTM. Thanks.

PR #3210, opened 2025-07-08
feat: add csv consumer and producer

thecampagnards:
Hello,
It's a PR to have csv producer and consumer to avoid this kind of block inside `configure_xxxx_api.go`:
```go
	api.CsvProducer = runtime.ProducerFunc(func(w io.Writer, data interface{}) error {
		return errors.NotImplemented("csv producer has not yet been implemented")
	})
```
and rather have
```go
	api.CsvProducer = runtime.CSVProducer()
```

Thx !


fredbi (approved):
Good addition. Thanks

PR #3215, opened 2025-07-22
feat: add cli-package option of cli generator

xpofei:
This PR provides the --cli-package option for the `swagger generate cli` command, completing the TODO left by youyuanwu@outlook.com earlier.


fredbi (approved):
LGTM.

Perhaps we could use a follow-up PR to update the documentation with this new option.

Thanks for completing this work.

Issue #3220, opened 2025-08-14
go-swagger fails to install using Go 1.25.0


vdpoora:
Installing the latest or previous release `v0.31.0` using go 1.25 fails with the following error

```
go install github.com/go-swagger/go-swagger/cmd/swagger@latest

# golang.org/x/tools/internal/tokeninternal
../../../../../pkg/mod/golang.org/x/tools@v0.21.0/internal/tokeninternal/tokeninternal.go:64:9: invalid array length -delta * delta (constant -256 of type int64)
```

Installing with GOTOOLCHAIN set to go1.24.6 or similar works.

```
GOTOOLCHAIN=go1.24.6 go install github.com/go-swagger/go-swagger/cmd/swagger@latest
```

PR #3221, opened 2025-08-18
update dependencies

cwedgwood:
update dependencies, this allows go1.25 to work but has the side-effort of requiring go1.24+

Fixes: #3220


fredbi:
@cwedgwood  unfortunately, our codescan package doesn't pass go1.22... I am currently working on a real fix to support go aliases with the new AST from go1.22. Just like you I was initially driven by the desire to upgrade, but hit severe difficulties with that part of the codebase.

fredbi:
Superseded by #3227

PR #3222, opened 2025-08-20
migration strategy

dheins:


fredbi:
I've also been working on a complete rework lately, starting with the go-openapi stuff.

PR #3224, opened 2025-08-21
Fix go 1.25 compatibility

gabriel-samfira:
This change updates `golang.org/x/tools` to `v0.24.1`.

Fixes: #3220


gabriel-samfira:
There's already a PR. My apologies!

gabriel-samfira:
for one of my projects, cloning this repo, updating golang.org/x/tools to v0.24.1, installing it and running: `swagger generate spec` and `swagger generate client` or `swagger validate` worked fine on go 1.25.0.

For now I'm going with: `GOTOOLCHAIN=go1.24.6 go generate ./...` until dependencies are updated.

fredbi:
yeah it is working fine until you have something like "type X = Y" in your code (a type alias)

fredbi:
since go1.25 is coming soon, I am wondering if it wouldn't be better to move directly to go1.24 rather than go1.23 and toolchain go1.24. What do you think?

gabriel-samfira:
Go 1.25.0 has already been released, but moving directly to 1.24.x is a good idea. The current version of `x/tools` is still compatible with go version 1.24.x. In my limited use case at least, 1.24.x has been working great for a long time. Current go-swagger tests seem to pass with 1.24.x (and even with 1.25.0 after updating x/tools), You clearly have some more complex use cases in mind (I barely use any generics), so my input in regards to whether or not you should update is not that relevant.

Are there any go-swagger tests that expose the issues you've been seeing with updating to a newer go version?

PR #3239, opened 2025-10-18
Backwards compatibility flag for alias handling

cce:
## Summary

Adds a new `--transparent-aliases` flag to `swagger generate spec` that makes type aliases completely transparent during spec generation. When enabled, aliases never create their own definitions in the swagger spec - references point directly to the underlying type instead.

This provides a backwards compatibility option for projects that depend on stable swagger specs and were affected by the alias handling changes in #3227.

## Motivation

PR #3227 improved Go 1.24 compatibility and added proper support for type aliases by creating swagger definitions for them. While this is the correct default behavior (aliases are distinct types in the API surface), it changed the generated swagger specs for existing codebases, which  be disruptive.

## Solution

The `--transparent-aliases` flag restores the pre-#3227 behavior where type aliases are treated as completely transparent to the spec generation process:
```go
type Identifier string
type UserID = Identifier  // alias

type User struct {
    ID UserID `json:"id"`
}
```

**Without flag (default):**
Generates swagger with **three** definitions:
```json
"definitions": {
  "Identifier": {"type": "string"},
  "UserID": {"$ref": "#/definitions/Identifier"},
  "User": {
    "properties": {
      "id": {"$ref": "#/definitions/UserID"}
    }
  }
}
```

**With `--transparent-aliases`:**
Generates swagger with **two** definitions (UserID omitted):
```json
"definitions": {
  "Identifier": {"type": "string"},
  "User": {
    "properties": {
      "id": {"$ref": "#/definitions/Identifier"}
    }
  }
}
```
## Example

See `examples/alias-compatibility/` for a minimal working example demonstrating the difference.

## Testing

- Added unit tests comparing output with and without the flag
- Verified existing tests continue to pass
- Tested with go-algorand's real-world API definitions

## Related

- Complements the existing `--with-expand` flag (which expands all `$ref` entries)
- Works alongside `--flatten` options
- Does not affect the `RefAliases` scanner option (which controls whether aliases are expanded or define their own model and are referred to by $ref).


fredbi (changes requested):
Another point: since you've changed the behavior for parameters and responses too, could you please add a test case that reasonably covers that part of the patch?

cce:
Thanks, I think I've addressed your feedback with some more test fixtures and updated the docs. Also, thanks for developing and maintaining this tool!

fredbi (approved):
Looks great. Thank you for this contribution.

PR #3247, opened 2025-10-24
Uppdate go-openapi/swag: v0.23.1 -> v0.25.1

winebarrel:
Update go-openapi/swag to v0.25.1 and use the yamlutils package.

- cf. https://github.com/go-openapi/swag/pull/111


fredbi:
Ok. This was a long-planned ahead operation. I am definitely not against upgrading bit by bit.

Could you please explain why yamlutils specifically should come first and not wait until we move ALL dependencies to the new swag v0.25 modules, including in generated code.

PR #3248, opened 2025-10-24
Follow-up on #3195: demonstrate codegen performance improvement

fredbi:
This PR fixes the codegen testing tool used by CI.

We want to assert the #3195 indeed boosts codegen.


fredbi:
@seiyab  I've managed to get the expected boost in our codegen CI tests. It was just some things I had omitted in the test harness. So we've got x5 boost. Thank you so much.

Issue #3151, opened 2025-01-15
analysis.New is called every time generate a model


Sczlog:
## Problem statement

it will take around 30 minutes to generate a client with go-swagger as I have a huge schema, meanwhile for other tools like openapi-generator will only use few minutes.

I add pprof profiling in code for every 5 minutes, and found `analysis.New` eat many times.

![image](https://github.com/user-attachments/assets/c3c2c7e0-6a1f-4d2d-aed5-56b7c484a871)

Then I found everytime run `makeGenDefinitionHierarchy` will call `analysis.New` to analyze the spec.
If I don't misunderstand, the spec should be static while generating and could handle only once and reuse for later?

## Environment
swagger version: v0.29.0
go version:  1.23.4
OS:  Ubuntu 24.04.1 LTS

fredbi:
Your analysis is correct. I  believe that there are a few places like this one, repeating the work already done because later operations have the bad habit of mutating the spec. The main issue here is that the spec object is full of pointers that are mutated (e.g. resolving refs or stuff like this). If I remember correctly we used to had a similar issue when generating operations, and we resorted to deep-clone the schema object _before_ altering it. Perhaps a good way could be to add a "DeepClone" capability to the spec analyzer and reuse that clone in the loop rather than re-analyzing the stuff.

PR #3279, opened 2026-01-16
perf: remove duplicated spec analysis

cjc7373:
a simple fix for https://github.com/go-swagger/go-swagger/issues/3151. I don't know if this is right. But it seems the spec would not change during the model generate process.

Another easy fix is that pretty.Sprint also has significant cost, so I removed it.


fredbi:
The spec DOES change (not always, but MAY change) during the process.
Removing the pretty print is good I agree. These debug logs are not really helpful anyway.

PR #3288, opened 2026-03-01
codescan: fix panic in SetEnum when schema type is not initialized

KingrogKDR:
## Problem

When parsing comment-based enum declarations (e.g. `// enum: [...]`) on a top-level type referenced by a `swagger:model` struct, `SetEnum` assumes that `sv.current.Type` is non-empty and accesses `Type[0]`.

In some parsing paths, the schema type has not yet been initialized, resulting in:

` panic: runtime error: index out of range [0] with length 0 `

This causes `swagger generate spec --scan-models` to crash.

## Fix

Guard access to `sv.current.Type[0]` by checking that the slice is non-empty before indexing.

If no type has been set yet, pass an empty type to `parseEnum`, preventing the panic while preserving existing behavior when the type is available.

## Tests

Added a regression test that:

- Defines a `swagger:model` struct referencing a custom string type
- Uses a comment-based enum declaration (`// enum: [...]`)
- Runs the full `Run()` path

The test reproduces the panic before this change and passes after the fix.

### DIFF
--- a/codescan/schema_test.go
+++ b/codescan/schema_test.go
@@ -2524,15 +2524,12 @@ func TestSetEnumDoesNotPanic(t *testing.T) {
 	// enum: ["created","processed"]
 	type State string
 	`
-
 	err := os.WriteFile(filepath.Join(dir, "model.go"), []byte(src), 0600)
 	require.NoError(t, err)
 
 	goMod := `
 	module failure
-
-	go 1.23
-	`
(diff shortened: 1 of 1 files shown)


fredbi (approved):
Thanks. Approved.

KingrogKDR:
I have resolved the requested changes @fredbi

PR #3357, opened 2026-06-05
fix(generator): fixed initialism issue (#3319)

uddeshsingh:
PR fixes #3319 - `x-go-name` was erroneously overridden by common initialism rules (e.g., `x-go-name: NoTls` becoming `NoTLS`).

- Updated `knownDefGoType` in `generator/types.go` to intercept `x-go-name` for internal schemas and process the name string using `Camelize` instead of `ToGoName`.
- Ignored (`isExternalType`).
- Adjusted hardcoded string expectations in `generator/typeresolver_test.go` to match the accurate camel-cased formatting outputs.

Verified locally. All tests passing:
`ok  	github.com/go-swagger/go-swagger/generator`

### DIFF
--- a/generator/model.go
+++ b/generator/model.go
@@ -546,7 +546,7 @@ func (sg *schemaGenContext) NewAdditionalItems(schema *spec.Schema) *schemaGenCo
 		pg.Path = pg.Path + "+ \".\" + strconv.Itoa(" + indexVar + mod + ")"
 	}
 	pg.IndexVar = indexVar
-	pg.ValueExpr = sg.ValueExpr + "." + sg.pascalize(sg.GoName()) + "Items[" + indexVar + "]"
+	pg.ValueExpr = sg.ValueExpr + "." + sg.GenSchema.GoName + "Items[" + indexVar + "]"
 	pg.Schema = spec.Schema{}
 	if schema != nil {
 		pg.Schema = *schema
@@ -582,7 +582,7 @@ func (sg *schemaGenContext) NewStructBranch(name string, schema spec.Schema) *sc
 		pg.Path = pg.Path + "+\".\"+" + fmt.Sprintf("%q", name)
 	}
 	pg.Name = name
-	pg.ValueExpr = pg.ValueExpr + "." + sg.pascalize(goName(&schema, name))
+	pg.ValueExpr = pg.ValueExpr + "." + schemaGoName(&schema, name, sg.mangler)
 	pg.Schema = schema
 	pg.IsProperty = true
 	if slices.Contains(sg.Schema.Required, name) {
@@ -957,7 +957,7 @@ func (sg *schemaGenContext) buildAllOf() error {
 		}
 		if comprop.GenSchema.IsMap && comprop.GenSchema.HasAdditionalProperties && comprop.GenSchema.AdditionalProperties != nil && !comprop.GenSchema.IsInterface {
 			// the anonymous branch is a map for AdditionalProperties: rewrite value expression
-			comprop.GenSchema.ValueExpression = comprop.GenSchema.ValueExpression + "." + comprop.Name
+			comprop.GenSchema.ValueExpression = comprop.GenSchema.ValueExpression + "." + comprop.GenSchema.GoName
 			comprop.GenSchema.AdditionalProperties.ValueExpression = comprop.GenSchema.ValueExpression + "[" + comprop.GenSchema.AdditionalProperties.KeyVar + "]"
 		}
 
@@ -1068,11 +1068,11 @@ func (sg *schemaGenContext) buildAdditionalProperties() error {
 			comprop.GenSchema.Required = true
 			comprop.GenSchema.HasValidations = true
 
-			comprop.GenSchema.ValueExpression = sg.GenSchema.ValueExpression + "." + sg.mangler.ToGoName(sg.GenSchema.Name) + "[" + comprop.KeyVar + "]"
+			comprop.GenSchema.ValueExpression = sg.GenSchema.ValueExpression + "." + sg.GenSchema.GoName + "[" + comprop.KeyVar + "]"
 
 			sg.GenSchema.AdditionalProperties = &comprop.GenSchema
 			sg.GenSchema.HasAdditionalProperties = true
-			sg.GenSchema.ValueExpression += "." + sg.mangler.ToGoName(sg.GenSchema.Name)
+			sg.GenSchema.ValueExpression += "." + sg.GenSchema.GoName
 
 			sg.MergeResult(comprop, false)
 
@@ -1080,7 +1080,7 @@ func (sg *schemaGenContext) buildAdditionalProperties() error {
 		}
 
 		// this is a regular named schema for AdditionalProperties
-		sg.GenSchema.ValueExpression += "." + sg.mangler.ToGoName(sg.GenSchema.Name)
+		sg.GenSchema.ValueExpression += "." + sg.GenSchema.GoName
 		comprop := sg.NewAdditionalProperty(*addp.Schema)
 		d := sg.TypeResolver.Doc
 		asch, err := analysis.Schema(analysis.SchemaOpts{
@@ -1337,6 +1337,7 @@ func (sg *schemaGenContext) buildItems() error {
 			sg.MergeResult(elProp, false)
 
 			elProp.GenSchema.Name = "p" + strconv.Itoa(i)
+			elProp.GenSchema.GoName = exportGoName("P"+strconv.Itoa(i), false, sg.mangler)
 			sg.GenSchema.Properties = append(sg.GenSchema.Properties, elProp.GenSchema)
 			sg.GenSchema.IsTuple = true
 		}
@@ -1735,7 +1736,6 @@ func (sg *schemaGenContext) makeGenSchema() error {
 	sg.GenSchema.Path = sg.Path
 	sg.GenSchema.IndexVar = sg.IndexVar
 	sg.GenSchema.Location = body
-	sg.GenSchema.ValueExpression = sg.ValueExpr
 	sg.GenSchema.KeyVar = sg.KeyVar
 	sg.GenSchema.OriginalName = sg.Name
 	sg.GenSchema.Name = sg.GoName()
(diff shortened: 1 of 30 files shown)


generator/types.go
```diff
@@ -336,11 +336,27 @@ func (t typeResolver) knownDefGoType(def string, schema spec.Schema, clearFunc f
 	ext := schema.Extensions
 	nm, hasGoName := ext.GetString(xGoName)
 
+	extType, isExternalType := t.resolveExternalType(ext)
+
 	if hasGoName {
 		debugLogf("known def type %s named from %s as %q", def, xGoName, nm)
+
```
fredbi: the name mangling is located in model.go not here.

generator/types.go
```diff
@@ -336,11 +336,27 @@ func (t typeResolver) knownDefGoType(def string, schema spec.Schema, clearFunc f
...
+		if !isExternalType {
+			safeName := t.mangler.Camelize(nm)
```
fredbi: the distinction here of external types suggests that a fix for another potential issue is being addressed here, but I don't think it is related to the issue claimed to be fixed by this PR. Please be explicit about the problem you're trying to solve with this exception.

generator/typeresolver_test.go
```diff
@@ -68,7 +68,7 @@ func schTypeVals() []struct{ Type, Format, Expected string } {
 func schRefVals() []struct{ Type, GoType, Expected string } {
 	return []struct{ Type, GoType, Expected string }{
 		{"Comment", "", "models.Comment"},
-		{"UserCard", "UserItem", "models.UserItem"},
+		{"UserCard", "UserItem", "models.Useritem"},
```
fredbi: I don't quite get how this test is related to the processing of initialism.
Anyway this is a unit test of a very narrow function that you've changed rather drastically. The impact on actually generated code remains to be seen.

fredbi (changes requested):
I don't think the problem of handling x-go-name is handled at that location. Check out model.go instead.

generator/types.go
uddeshsingh: Thanks, you're right. I think i went a bit too low broad for that fix

uddeshsingh:
### Problem

When a model property uses `x-go-name` to force a Go identifier, the generator was still running the name through initialism rules. For example, `x-go-name: NoTls` was emitted as `NoTLS`, and `x-go-name: VmId` could become `VMID`.

`x-go-name` is meant to be an explicit override. Initialism normalization should not rewrite it.


### Changes and Intent

1. `structfield.gotmpl`:

Previously every property field went through `pascalize .Name`. `pascalize` calls
`ToGoName` twice and applies common initialisms (TLS, ID, …), so an explicit
`x-go-name` was overridden at emit time.

Now, `propertyName` template uses `.Name` as-is when `x-go-name` is in
`.Extensions`, and falls back to `pascalize` for normal properties.

2. `model.go`:

Shifted name mangling for model properties to `model.go`, not `types.go`.

`NewStructBranch` builds `ValueExpr` paths used for validation. It used `pascalize(goName(...))`,
which had the same initialism problem.

3. `options.go`:

Added Go export/prefix handling for `x-go-name` values (e.g. names starting with `+`, `-`), but not initialism rewriting.

`WithInitialisms()` with no words clears the default initialism list while keeping `ToGoName` and `PrefixForName`.

4. `operation.go`:

Passing `explicitGoNameMangler` keeps body-param property naming consistent with models.

**5. Test strategy**:

Added a fixture matching the issue and a model-generation test that asserts correct requirements

### Future Scope

1. `serializer.gotmpl` and `schemavalidator.gotmpl` still call `pascalize .Name` directly for
   field access. A follow-up could expose `propertyName` via the funcmap or
   consolidate naming in one helper.

2. Documentation: clarify in docs that `x-go-name` on properties preserves author casing and skips initialisms,

generator/typeresolver_test.go
uddeshsingh: addressed, thanks!

generator/types.go
uddeshsingh: right, dropped and addressed in a new comment

generator/templates/structfield.gotmpl
```diff
@@ -1,17 +1,25 @@
...
+{{- end -}}
+{{- end }}
+
 {{ define "structfield" }}
   {{- if not $.IsBaseType -}}
     // {{ template "docstring" . }}
     {{- template "propertyValidationDocString" .}}
   {{- end}}
-{{ pascalize .Name}} {{ template "schemaType" . }} {{ .PrintTags }}
+{{ template "propertyName" . }} {{ template "schemaType" . }} {{ .PrintTags }}
```
fredbi: since this applies to GenSchema objects (for now at least), a more elegant solution could be to add this
as a method of the GenSchema struct (in structs.go).  This approach would allow to more easily generalize to all the object types that you have not covered (parameters, responses, operations, plus all the schema sub-cases not covered yet). Idea:

func (g Genschema) GoName() string {
 if ... {
    return s.GenOpts.Mangler.ToGoName(g.Name)  // would require to add a pointer to GenOpts in all GenXXX object (currently missing from Schema and Parameter)
   }
   return titleize(Name) // small helper: we don't really need to ToGoName/pascalize here: just ensure the first character is uppercased (exported)
}

generator/templates/structfield.gotmpl
```diff
@@ -1,17 +1,25 @@
+{{ define "propertyName" -}}
+{{- if (index .Extensions "x-go-name") -}}
+{{- .Name -}}
+{{- else -}}
+{{- pascalize .Name -}}
+{{- end -}}
+{{- end }}
+
```
fredbi: I get the point. model.go is not sufficient since all templates call "pascalize" (hunder the hood: ToGoName()). However, this is going to be tedious to generalize since we'd need this across the board (not just structfield.gotmpl) as many objects are subject to the x-go-name override.

generator/templates/structfield.gotmpl
```diff
@@ -1,17 +1,25 @@
...
+{{ template "propertyName" . }} {{ template "schemaType" . }} {{ .PrintTags }}
 {{ end }}
 
 {{- define "tuplefield" }}
   {{- if not $.IsBaseType -}}
     // {{ template "docstring" . }}
     {{- template "propertyValidationDocString" .}}
 {{ end }}
-{{- pascalize .Name}} {{ template "schemaType" . }} `json:"-"
+{{- template "propertyName" . }} {{ template "schemaType" . }} `json:"-"
```
fredbi: example with the idea I suggested above:

{{ .GoName }} {{ template "schemaType" . }} json "-"

generator/model.go
```diff
@@ -267,6 +267,7 @@ func makeGenDefinitionHierarchy(name, pkg, container string, schema spec.Schema,
 		StructTags:                 opts.StructTags,
 		WantsRootedErrorPath:       opts.WantsRootedErrorPath,
 		mangler:                    opts.LanguageOpts.Mangler,
+		explicitGoNameMangler:      opts.LanguageOpts.ExplicitGoNameMangler,
```
fredbi: reflecting on this, I think that we can safely ditch entirely the ToGoName() / pascalize stuff for this override: the user is in charge, they won't (and should not) enter things like "* snow white and the 7 dwarves!!!". So I'd say I simple ToUpper on the first char would be enough

generator/templates/structfield.gotmpl
fredbi: Or, we precompute a GoName property (like you've almost done it already in model.go). Your call

fredbi (dismissed):
Interesting second pass, which reveals new things.

1. In schema, there are more impacted templates than structfield.gotmpl
2. There are more objects subject to x-go-name override than schemas (parameter, response, operation)
3. I realize that a tweak in model.go is far from sufficient: templates are affected. The inlined template approach may work but is impractical if we want to generalize
4. I realize that if we want to ensure a valid exported go name with x-go-name, we can drop entirely the ToGoName sanitizer: ensuring the first letter is upper-case would be enough
5. Overall excellent progress, but I can't agree with the approach to deliver it in bits with potential future scope extensions. Changing the rules in one particular part of the generator may suit your use-case of the moment, but overall makes the situation worse by being inconsistent

uddeshsingh:
Replaced the previous partial approach with a complete solution. Took a lot!

- Naming is now centralized in name.go and applied consistently.
- exportGoName() — with x-go-name, only uppercase the first char (plus PrefixForName for +/-/etc.); no ToGoName/initialisms. Without x-go-name, behavior is unchanged.
- GoName is precomputed for GenSchema, GenParameter and GenHeader. Used by model.go value paths.
-  Major changes happened in the templates! Replaced pascalize .Name with .GoName (schemas) or .ID (params/headers) across models, validators, serializers, CLI, and client/server param/response templates.
- Increased scope of tests

Out of scope: operation/response type names, anonymous body x-go-name (existing TODO), docs.

Note: invalid x-go-name on parameters → error; on headers → silent fallback (existing behavior).

uddeshsingh:
> * Major changes happened in the templates! Replaced pascalize .Name with .GoName (schemas) or .ID (params/headers) across models, validators, serializers, CLI, and client/server param/response templates.
> 
> * Increased scope of tests

just a heads up, due to large amount of changes, had to use AI to review and help with tests and templates. I have reviewed manually, but please let me know if we don't prefer that here, will refrain from using it in the future

fredbi:
Thanks. I am going to review this thoroughly.

It is okay to use AI when there are people at the wheel we can interact with.

fyi I am preparing in parallel a pre-sanitization of extensions such as x-go-name. I’ll wait for your changes to land before introducing that change. It will reinforce your design, simplifying even further the « first letter upper-cased » path (x-go-name with invalid nothing go ident will error upstream).

Fred

fredbi (commented):
Looks great!

Given the amount of change I need to checkout 
Your branch And play a bit with it.
By the look of it it seems quite fine - just need to dive in the code not just the diffs.
If you’ve passed the full ci it is a good sign.

I’ll get back to you asap


Cheers

Fred

fredbi (approved):
Looks good to me.  Great contribution ! thank you

uddeshsingh:
Great, thanks a lot! Looking forward to learn more.

PR #3371, opened 2026-06-14
ci: add standalone golangci-lint workflow with workflow_dispatch

dashitongzhi:
## Summary

- Add a new `.github/workflows/lint.yaml` that runs `golangci-lint` as a standalone job, separate from the monolithic `test.yaml` workflow.
- Add a `workflow_dispatch` trigger so maintainers can re-run lint on demand without pushing a dummy commit.
- Pin `golangci-lint` to `v2.1.6` to match the schema version (`version: "2"`) declared in the existing `.golangci.yml`.
- Add a `concurrency` group (`lint-${{ github.workflow }}-${{ github.ref }}`) so outdated runs on the same ref are cancelled.
- Reuse the same pinned action SHAs (`actions/checkout@v6.0.2`, `actions/setup-go@v6.4.0`, `golangci/golangci-lint-action@v9.2.0`) and the "Originating repo checkout" pattern that the rest of this repo's workflows already use, so the behavior on public forks stays consistent.

## Why

The project already ships a `.golangci.yml` (v2 schema) and a golangci-lint step is embedded inside `test.yaml`, but there is no dedicated lint workflow and no `workflow_dispatch` trigger anywhere under `.github/workflows/`. A standalone lint workflow gives lint its own status check (so a lint regression shows up as a distinct check on the PR), keeps lint feedback separate from the much longer-running test jobs, and lets maintainers re-run lint on demand.

## Test Plan

- [ ] New `Lint / golangci-lint` check appears on this PR and is green.
- [ ] Triggering the workflow manually from the Actions tab works.
- [ ] Pushing a new commit to the branch cancels the in-flight run (concurrency group).
- [ ] No change in behavior for the existing `Test` workflow.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,55 @@
+name: Lint
+
+permissions:
+  contents: read
+  pull-requests: read
+
+# description: |
+#   Standalone golangci-lint runner.
+#
+#   The project ships a `.golangci.yml` (v2 config) and golangci-lint is
+#   also invoked from the monolithic test workflow, but there is no
+#   dedicated lint workflow. This workflow gives lint its own status
+#   check, lets maintainers trigger it on demand via `workflow_dispatch`,
+#   and isolates lint feedback from the longer-running test jobs.
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+# Cancel any in-flight run for the same ref when a new commit lands.
+concurrency:
+  group: lint-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  golangci:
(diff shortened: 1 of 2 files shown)


fredbi:
@dashitongzhi  you're just burning tokens for no use.

If you have ideas or suggestions regarding or CI pipelines, please engage with the maintainers on our discord channel before pushing a PR.

I truly hope that this the last of your 100% auto-generated PRs that come with no motivation, no comment, no real value added and basically drain the maintainers' time.

fredbi:
@dashitongzhi your repeated efforts to contribute are however appreciated, but as you certainly understand it, such changes can't be pushed like this right off a bare generated PR. Communication is key here. So you're welcome to file an issue where you explain your problem, or engage in a discussion on discord.

PR #3377, opened 2026-06-22
Add opt-in --with-stringer to generate model String() methods

### LINKED ISSUE #872: Improve readability of models as strings
## Problem statement

Models can be used as errors, when used for an non-200 status code. This is done by formatting the model value directly with `%#v`, which will simply format the models using their Go Syntax. This works pretty great, except when the properties of the model are of a pointer type, in which case the address of the property instead of its value get printed.

This makes that the resulted error messages aren't too great. Take for example a model which simply defines a `*string* message property. With the code as it is right now, it will simply print the address of that message string, rather than the actual message. Thus making the resulting error messages not as effective as they could be.

youdie006:
## Summary

Generated models have no `fmt.Stringer`, so printing one with `%v` / `%+v` shows pointer addresses instead of field values (#872).

This adds a **`--with-stringer`** generator option that emits a `String()` method rendering the model as JSON — the approach @GlenDC proposed and @casualjim / @fredbi were OK with in the issue thread.

It is **opt-in**: without the flag, generation output is byte-for-byte unchanged (the default example fixtures are untouched), so existing users are unaffected. With the flag, each model gets a `String()` that marshals it to JSON.

(The issue's original `%+v` client-error complaint was already fixed separately in `response.gotmpl`; this addresses the remaining ask — giving the models themselves a Stringer.)

Added a generator test (`TestIssue872`) covering the option both on and off.

Closes #872

### DIFF
--- a/cmd/swagger/commands/generate/model.go
+++ b/cmd/swagger/commands/generate/model.go
@@ -19,6 +19,7 @@ type modelOptions struct {
 	AllDefinitions             bool     `description:"generate all model definitions regardless of usage in operations"                                      hidden:"deprecated"                          long:"all-definitions"`
 	StructTags                 []string `description:"the struct tags to generate, repeat for multiple (defaults to json)"                                   long:"struct-tags"`
 	RootedErrorPath            bool     `description:"extends validation errors with the type name instead of an empty path, in the case of arrays and maps" long:"rooted-error-path"`
+	WithStringer               bool     `description:"generate a fmt.Stringer String() method on models, rendering field values as JSON (see issue #872)"    long:"with-stringer"`
 }
 
 func (mo modelOptions) apply(opts *generator.GenOpts) {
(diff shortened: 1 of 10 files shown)


fredbi (changes requested):
Yes this is a good idea, and opt-in anyway.
That's right that the desire for it emerged a long time ago, but no one really pushed for it. Glad it eventually landed.

Please check your linting (tag alignment issues may usually be auto-fixed with golangci), rebase and this should be good to go.

Thanks.

Fred

youdie006:
Thanks @fredbi! Fixed the `tagalign` lint issue (golangci-lint `--fix`) and rebased onto `master` — `golangci-lint run` is now clean and `go test ./generator/` passes. Let me know if anything else is needed.

fredbi (approved):

fredbi:
Thank you for this addition.

Issue #3171, opened 2025-02-22
zsh: killed


tikhonfedulov:
On systems with macOS sequoia 1.15.1, SIP kills the go-swagger process with CGO enabled. 

```sh
Cloning into 'go-swagger'...
remote: Enumerating objects: 53704, done.
remote: Counting objects: 100% (6783/6783), done.
remote: Compressing objects: 100% (910/910), done.
remote: Total 53704 (delta 6113), reused 6033 (delta 5852), pack-reused 46921 (from 1)
Receiving objects: 100% (53704/53704), 52.53 MiB | 968.00 KiB/s, done.
Resolving deltas: 100% (33469/33469), done.
tikhonfedulov@MacBook-Air-Tikhon Downloads % cd go-swagger/cmd/swagger 
tikhonfedulov@MacBook-Air-Tikhon swagger % go build -o swagger
go: downloading go1.22.3 (darwin/arm64)
tikhonfedulov@MacBook-Air-Tikhon swagger % ./swagger version
zsh: killed     ./swagger version
```
 
This is not obvious and I only noticed it when I started looking for the difference between the manual build and the release version. 
Spoiler:
https://github.com/go-swagger/go-swagger/blob/master/.goreleaser.yml#L10

CGO is already disabled by default for releases. As I understand, we can't disable CGO via `//go:build` in some `build.go`?

fredbi:
now a documented caveat

PR #3384, opened 2026-06-30
docs: add killed-process troubleshooting

dashitongzhi:
## Summary
- add troubleshooting guidance for `zsh: killed` / `Killed` during `swagger generate`
- document likely resource-limit and memory-pressure causes
- add macOS and Linux diagnostic commands and issue-reporting details

References #3171.

## Test plan
- Docs-only change; no code tests run.
- Reviewed the rendered Markdown diff locally.


fredbi (changes requested):
Thank you for this extra piece of documentation.

It is okay to document out-of-memory failures (how to diagnose) but I think your document is rambling on a bit about how to prevent and solve that. Our documentation does’nt aim to become a vademecum of useful linux commands.

Further, in the original issue, @tikhonfedulov mentioned a strange relationship with cgo, which remains unexplained. I am confident that @tikhonfedulov knows already quite a bit about simple out-of-memory diagnostic so this must be something trickier.

Also: please sign your commits (DCO). Agents may coauthor, but only people are responsible for their commits.

Final note: fyi the doc site is going to be transformed quite a bit later this month (new theme, reorganized content). You may still contribute additions but they may be moved or reshuffled later on.

fredbi:
@dashitongzhi  I've followed-up your proposal with some additions: https://github.com/go-swagger/go-swagger/pull/3390

Thank you for proposing this addition - just needed to get to a resolution quicker.

Issue #3394, opened 2026-07-17
Non deterministic code generation from swagger file containing `application/protobuf+json` producer


DejeroDaniel:
Hello, happy user of go-swagger here. Appreciate all of your great work in maintaining this repo.

Recently came across a bug in an older version of go-swagger, (0.26.1), and the issue persisted after updating to the newest version (0.35.0).
We use the model and server generation from a swagger design doc, and recently added a new API route with a new application type.

The official [ProtoJSON docs](https://protobuf.dev/reference/protobuf/mime-types/) mention the `application/protobuf+json` application type, and we wanted a custom producer to use the protojson Go library to encode protobuffers returned from the API routes. So `application/protobuf+json`  was added to the `produces` section of the swagger file.

However, now whenever we run the swagger command (`swagger generate server -qt $(SWAGGER_GEN_DIR) --exclude-main -f design/api.yaml`) the generator modifies the ProtobufProducer within the operations package.
No swagger changes are needed.

I hadn't had any issues like this before I added a new producer (we have only been using `application/json` previously).
I updated the swagger binary to the most recent version, (0.35.0) but the issue persisted.

I am running the swagger binary within an alpine (1.25) docker container, itself running on my debian linux laptop.

Looking at the [code](https://github.com/go-swagger/go-swagger/blob/dae350a8ebe1d6d8ff5b5453051804e2065f2904/generator/media.go#L24), I saw the regex to generate the protobuf producer wouldn't quite match the protojson application type.
I added `application/protobuf` to the produces section of the swagger document and the ProtobufProducer would be consistently created.
However, the `application/protobuf+json` line will switch between JSON and ProtobufProducer's on each run, like so:

```
// ProducersFor gets the producers for the specified media types.
//
// MIME type parameters are ignored here.
func (o *InternalAPI) ProducersFor(mediaTypes []string) map[string]runtime.Producer {
	result := make(map[string]runtime.Producer, len(mediaTypes))
	for _, mt := range mediaTypes {
		switch mt {
		case "application/json":
			result["application/json"] = o.JSONProducer
		case "application/protobuf":
			result["application/protobuf"] = o.ProtobufProducer
		case "application/protobuf+json":
			result["application/protobuf+json"] = o.ProtobufProducer
		}

		if p, ok := o.customProducers[mt]; ok {
			result[mt] = p
		}
	}

	return result
}
```

and

```
// ProducersFor gets the producers for the specified media types.
//
// MIME type parameters are ignored here.
[...]

PR #3398, opened 2026-07-20
fix(codegen): ProducersFor and ConsumersFor should prioritize custom serializers

fredbi:
* relates to #3394


fredbi:
Not necessary. I misread generated code. Closing

PR #3409, opened 2026-07-25
feat: support streaming multipart forms on generated servers

### LINKED ISSUE #3407: Proposal: support opt-in streaming for multipart file uploads
## Summary

I would like to propose an opt-in streaming mode for Swagger 2.0
`multipart/form-data` file parameters in generated go-swagger servers.

The current generated binding flow parses the complete multipart form before
invoking the application handler. As a result, the handler cannot consume or
forward file bytes while the client is still uploading them.

The existing behavior is useful and should remain the default. The proposal is
to add an explicitly enabled streaming alternative.

## Current behavior

For a parameter such as:

```yaml
- name: file
  in: formData
  required: true
  type: file
```

go-swagger v0.35.3 generates a `BindRequest` implementation that calls:

```go
runtime.BindForm(
    r,
    runtime.BindFormMaxParseMemory(...),
    runtime.BindFormMaxBody(...),
    runtime.BindFormFile("file", true, o.bindFile),
)
```

The generated file validation also seeks to the end of the parsed
`multipart.File` to determine its size.

Therefore, even though the generated handler parameter is an `io.ReadCloser`,
the application receives it only after the request body has already been
parsed.

## Reproducer

Repository:

https://github.com/fpawel/go-swagger-multipart-streaming-repro

Exact tested revision:

https://github.com/fpawel/go-swagger-multipart-streaming-repro/commit/0923884fe22522e6dde75655f54138ff731b2aa1

The repository contains:

- a committed server snapshot generated by go-swagger v0.35.3;
- a multipart client that pauses after sending its first 64 KiB;
- a deterministic in-process comparison;
- a restricted streaming prototype based on `multipart.Reader`;
- captured client and server logs.

## Reproduction with the generated server

Start the generated server:

```bash
make run-current
```

In another terminal:

```bash
make client-current
```

Relevant output:

```text
Client:
19:24:23.954257 client wrote 65536 bytes
19:24:23.954261 client paused for 3s
19:24:26.955331 client resumed
19:24:26.963224 multipart body closed

Generated server:
19:24:26.967635 application handler started
19:24:26.968255 application handler finished: 8388608 bytes
```

The application handler starts only after the client has resumed, sent the
remaining bytes, and closed the complete multipart body.

## Streaming proof of concept

Start the prototype:

```bash
make run-streaming
```

In another terminal:

```bash
make client-streaming
```

Relevant output:

```text
Client:
19:24:46.859657 client wrote 65536 bytes
19:24:46.859665 client paused for 3s
19:24:49.860783 client resumed
19:24:49.871595 multipart body closed

Streaming prototype:
19:24:46.860641 application handler received "payload.bin"
19:24:49.872391 application handler finished: 8388608 bytes
```

Here the application receives the file part while the client is still paused,
approximately three seconds before the remaining bytes are uploaded.

The deterministic comparison also reports:

```text
current:
application started while client was paused: false

streaming-poc:
application started while client was paused: true
```

## Use case

An application may want to forward an incoming upload directly to object
storage or another downstream service:

```text
HTTP request body
    -> multipart part reader
    -> application handler
    -> object storage writer
```

With the current binder, this cannot be a true end-to-end streaming pipeline
because the generated handler is invoked only after multipart parsing has
completed.

## Expected opt-in behavior

For an explicitly opted-in operation or file parameter, the generated handler
would receive a reader after the matching multipart part headers have been
parsed, while the part body is still arriving from the client.

The existing `runtime.BindForm` behavior should remain unchanged for all
operations that do not opt in.

One possible direction, only as a discussion starter, could be a vendor
extension:

```yaml
- name: file
  in: formData
  required: true
  type: file
  x-go-streaming: true
```

The exact extension name and whether it belongs at operation or parameter level
would need maintainer feedback.

## Prototype scope

The prototype is intentionally restricted and is not presented as a drop-in
replacement for `runtime.BindForm`.

It currently assumes:

- exactly one streamed file part;
- the streamed file is the first multipart part;
- no trailing form fields;
- explicit request-body and file-size limits;
- the application consumes the reader before returning;
- existing generated APIs remain unchanged without opt-in.

## Design questions

1. Should streaming be enabled at operation level or file-parameter level?
2. Should the runtime expose a separate streaming multipart binder?
3. Who owns and closes the multipart part reader?
4. How should generated `minLength` and `maxLength` validation work when the
   size is only known while the handler reads the stream?
5. How should read-time limit violations be mapped to generated responses?
6. Should form fields preceding the streamed file be supported initially?
7. Should trailing form fields or multiple streamed files be explicitly
   rejected in the first version?
8. Would maintainers prefer this proposal to be split between
   `go-swagger/go-swagger` and `go-openapi/runtime`?

## Environment

```text
Go:          go1.26.4 darwin/arm64
go-swagger:  v0.35.3
runtime:     github.com/go-openapi/runtime v0.32.6
OS:          macOS 26.5.2, arm64
```

I would be interested in implementing this upstream once the desired public API
and initial scope are agreed upon.

fpawel:
## Summary

Adds opt-in server-side streaming for `multipart/form-data` operations.

A file parameter may enable streaming with:

```yaml
parameters:
  - name: file
    in: formData
    type: file
    x-go-server-streaming: true
```

When enabled, the generated server binder passes the complete multipart request to the operation handler as a `*runtime.MultipartFormStream` instead of eagerly binding the form with `runtime.BindForm`.

Fixes #3407.

## Motivation

The existing generated server binding parses the complete multipart form before invoking the handler. This prevents applications from processing large uploads while the request body is still being received.

The streaming runtime implementation was introduced in go-openapi/runtime#507. This PR integrates that API into go-swagger code generation.

## Generated server behavior

For an operation with `x-go-server-streaming: true`:

- the extension applies to the complete multipart form;
- generated `formData` fields are not added to the server params struct;
- the params struct exposes a `*runtime.MultipartFormStream`;
- path, query and header parameters continue to be generated and validated normally;
- the handler owns multipart traversal and application-specific validation;
- the handler must call `Drain()` or `Close()` when processing is complete;
- the generated body-size limit remains configurable;
- the buffered multipart memory-limit setting is not generated.

Required fields, accepted field names, multiplicity and other form-specific rules are intentionally left to the handler because multipart parts are consumed sequentially.

Client generation remains unchanged.

Operations without the extension continue to use the existing buffered multipart binding.

## Extension validation

`x-go-server-streaming` must:

- be a boolean;
- be enabled only on a `formData` parameter of type `file`.

Invalid usage fails during generation with an operation and parameter-specific error.

## Compatibility

This is opt-in and does not change existing generated server or client behavior.

Regular multipart operations continue to use `runtime.BindForm`.

## Tests

Added coverage for:

- generated streaming server binding;
- preservation of query parameter binding and validation;
- unchanged client generation;
- unchanged buffered multipart generation;
- invalid extension types and locations;
- generated field-name conflict resolution;
- streaming-specific `configureAPI` output.

A black-box integration example that verifies the handler receives file bytes before the client completes the request body is available in go-swagger/examples#71.

## Dependency

This PR currently references the merged runtime implementation through its upstream pseudo-version:

```text
github.com/go-openapi/runtime v0.32.7-0.20260724174133-f44e6731289f
```

It can be replaced with the corresponding tagged runtime release when that release becomes available.

### DIFF
--- a/docs/reference/server.md
+++ b/docs/reference/server.md
@@ -119,3 +119,32 @@ case the PublicListTravelsError will be returned, otherwise the PublicListTravel
 
 The code generator has written the remaining code to render that response with the headers etc.
 
+## Streaming multipart forms
+
+A server operation may opt out of generated `formData` binding and hand the
+multipart request stream directly to its handler. Mark one of the operation's
+file parameters with `x-go-server-streaming: true`:
+
+```yaml
+consumes:
+  - multipart/form-data
+parameters:
+  - name: file
+    in: formData
+    type: file
+    x-go-server-streaming: true
+```
+
+For the generated server, this extension applies to the complete multipart
+form. Generated code does not pre-read the first file and does not bind or
+validate individual `formData` parameters. Instead, the operation params expose
+a `*runtime.MultipartFormStream`. The handler owns sequential traversal with
+`NextFile()`, application-specific validation, and the decision to call
+`Drain()` or `Close()`.
+
+Other parameter locations, such as path, query and header parameters, continue
+to be generated and validated normally. Client generation is unchanged.
+
+The stream records fields and file metadata discovered so far. Parts that occur
+later in the request are not visible until the handler consumes or closes the
+current file and advances the stream.
--- a/fixtures/codegen/streaming-form-invalid.yml
+++ b/fixtures/codegen/streaming-form-invalid.yml
@@ -1,0 +1,31 @@
+swagger: "2.0"
+info:
+  title: Invalid streaming multipart form fixture
+  version: "1.0.0"
+consumes:
+  - multipart/form-data
+produces:
+  - application/json
+paths:
+  /invalid-type:
+    post:
+      operationId: invalidStreamingType
+      parameters:
+        - name: file
+          in: formData
+          type: file
+          x-go-server-streaming: "yes"
+      responses:
+        "204":
+          description: uploaded
+  /invalid-parameter:
+    post:
+      operationId: invalidStreamingParameter
+      parameters:
+        - name: token
+          in: query
+          type: string
+          x-go-server-streaming: true
+      responses:
+        "204":
+          description: uploaded
--- a/fixtures/codegen/streaming-form.yml
+++ b/fixtures/codegen/streaming-form.yml
@@ -1,0 +1,46 @@
+swagger: "2.0"
+info:
+  title: Streaming multipart form fixture
+  version: "1.0.0"
+consumes:
+  - multipart/form-data
+produces:
+  - application/json
+paths:
+  /stream:
+    post:
+      operationId: streamingUpload
+      parameters:
+        - name: token
+          in: query
+          type: string
+          required: true
+        - name: description
+          in: formData
+          type: string
+          required: true
+        - name: file
+          in: formData
+          type: file
+          required: true
+          x-go-server-streaming: true
+        - name: attachment
+          in: formData
+          type: file
+      responses:
+        "204":
+          description: uploaded
+  /regular:
+    post:
+      operationId: bufferedUpload
+      parameters:
+        - name: description
+          in: formData
+          type: string
+        - name: file
+          in: formData
+          type: file
+          required: true
+      responses:
+        "204":
+          description: uploaded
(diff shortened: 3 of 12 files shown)


fredbi:
looks rather good to me. I've just cut release go-openapi/runtime@v0.33.0 which onboards your addition there.

fredbi:
@fpawel  looks good to me. You may upgrade your go.mod to runtime@v0.33.0, which onboards your feature there.

fpawel:
> @fpawel looks good to me. You may upgrade your go.mod to runtime@v0.33.0, which onboards your feature there.

done

fredbi (approved):
Looks good. This is just fine..

PR #3419, opened 2026-08-08
fix(generator): honor x-go-type novalidation hint for aliased array types

### LINKED ISSUE #3141: Validate interface requirements improvement
## Problem statement

Here's a bit of my generated code 

```golang
if m.Queues[i] != nil {
    if err := m.Queues[i].Validate(formats); err != nil {
```

which comes from `/schemavalidator.gotmpl`.

Works perfectly well for models generated by go-swagger, but when bringing my own types using `x-go-type`, to be forced to implement the `Validate` method that specifically matches that signature (`Validate(formats strfmt.Registry)`) is a bit of an anti-pattern, because I may want to be providing an implementation that is completely disjoint from OpenAPI.

I'd suggest that 

```golang
if m.Queues[i] != nil {
	var err error
	if cast, ok := any(m.Queues[i]).(interface{ Validate( strfmt.Registry ) error }); ok {
		err = cast.Validate(formats)
	}
	if cast, ok := any(m.Queues[i]).(interface{ Validate( ) error }); ok {
		err = cast.Validate()
	}

	if err != nil {

```

might be more suitable. Of course we could even be smarter and, in the templating engine, determine if the type in brought externally or not...

Background: we're using grpc-gateway and https://github.com/bufbuild/protoc-gen-validate to validate types, which provides a different interface than the one required by `go-swagger`

RubenPari:
## Summary

Fixes #3141.

`slicevalidator`'s guard for calling `.Validate(formats)` on an aliased array-kind external type nested `(not .SkipExternalValidation)` **inside** the `or` clause instead of AND-ing it:

```
{{- if and (or .IsAliased (ne .ValueExpression .ReceiverName) (not .SkipExternalValidation)) }}
```

Since `ne .ValueExpression .ReceiverName` is true for virtually any property access, the `SkipExternalValidation` check was effectively dead. Properties using `x-go-type` with `hints.kind: array` and `hints.novalidation: true` were still forced through go-swagger's `Validate(strfmt.Registry) error` interface, defeating the opt-out and forcing users to implement a signature they don't want (the exact complaint in #3141).

Fix: corrected the condition to `and (or .IsAliased (ne .ValueExpression .ReceiverName)) (not .SkipExternalValidation)`, matching the already-correct pattern used in `objectvalidator` and `mapvalidator`.

Also added the same `SkipExternalValidation` guard to `objectcontextvalidator`/`slicecontextvalidator`/`mapcontextvalidator`, which never checked it at all, for parity with their `Validate()` counterparts.

## Test plan

- [x] New fixture `fixtures/bugs/3141/fixture-3141.yaml` exercising object, array-aliased, slice-items, and map-values properties backed by `x-go-type` + `hints.novalidation: true`.
- [x] New `initFixture3141` in `generator/moreschemavalidation_fixtures_test.go`, wired into `TestMoreModelValidations`.
- [x] Verified the new test fails against the unpatched template (reproduces `m.Arr.Validate(formats)` being generated despite the opt-out) and passes with the fix.

### DIFF
--- a/fixtures/bugs/3141/fixture-3141.yaml
+++ b/fixtures/bugs/3141/fixture-3141.yaml
@@ -1,0 +1,49 @@
+swagger: "2.0"
+info:
+  title: "x-go-type novalidation hint must be honored by Validate/ContextValidate"
+  version: "0.0.1"
+paths:
+  "/":
+    get:
+      responses:
+        200:
+          description: ok
+          schema:
+            $ref: '#/definitions/Container'
+
+definitions:
+  ExternalNoValidation:
+    x-go-type:
+      type: Ext
+      import:
+        package: github.com/go-swagger/go-swagger/fixtures/bugs/3141/external
+      hints:
+        kind: object
+        novalidation: true
+    x-nullable: true
+
+  ExternalArrayNoValidation:
+    x-go-type:
+      type: HotArray
+      import:
+        package: github.com/go-swagger/go-swagger/fixtures/bugs/3141/external
+      hints:
+        kind: array
+        novalidation: true
+    x-nullable: true
+
+  Container:
+    type: object
+    properties:
+      item:
+        $ref: '#/definitions/ExternalNoValidation'
+      arr:
+        $ref: '#/definitions/ExternalArrayNoValidation'
+      items:
+        type: array
+        items:
+          $ref: '#/definitions/ExternalNoValidation'
+      itemMap:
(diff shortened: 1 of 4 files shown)


fredbi:
Thanks @RubenPari  looks good. Let's see how CI goes along with it.

fredbi (approved):
Excellent piece of sleuthing. Thank you so much

PR #3420, opened 2026-08-08
feat(generator): generate const block for array-item enum values

### LINKED ISSUE #1203: Enum array doesn't generate a constant
## When generating model, enum arrays do not generate a constant for the enum.

I expected a constant to be generated for an enum array of strings like it is when generating and enum that is not an array. 

## Swagger specification
A constant should be created for roles in the below spec:
```
  foo:
    type: object
    properties:
      name:
        type: string
      organizationId:
        format: int32
        type: integer
      roles:
        type: array
        items:
          type: string
          enum:
            - Admin
            - User
            - SuperAdmin
```

## Environment
swagger version: 0.11.0 
go version: 1.8
OS:   OSX

RubenPari:
## Summary

Fixes #1203.

A scalar property with `enum` already gets a Go `const` block generated for its values (e.g. `FooStatusPending`). An array property whose *items* declare `enum` (`type: array, items: {type: string, enum: [...]}`) never got the equivalent consts — only the `{{ .GoName }}ItemsEnum` var/`init()`/validate-function plumbing used internally for runtime validation was generated. Users referencing enum values from other code (e.g. `models.FooRolesItemsAdmin`) had nothing to reference.

## Fix

Added the same `const (...)` block generation (gated on `Items.SwaggerType == "string"`, matching the existing restriction for scalar enum consts) to the three places array-item enums are rendered in `generator/templates/schemavalidator.gotmpl`: top-level schema, property, and allOf-nested property.

Variant naming follows the existing per-property enum const convention with an added `Items` segment, matching what was proposed in the issue thread: `FooRolesItemsAdmin` for property `roles` enum value `Admin` on model `Foo`.

## Test plan

- [x] New fixture `fixtures/bugs/1203/fixture-1203.yaml` (spec from the issue).
- [x] New `initFixture1203` in `generator/moreschemavalidation_fixtures_test.go`, wired into `TestMoreModelValidations`, asserting the generated `const` block.
- [x] Verified the new test fails against the unpatched template (no `const` block generated) and passes with the fix.
- [x] Manually generated the model and ran `go build` on the output to confirm the generated Go code compiles.

### DIFF
--- a/fixtures/bugs/1203/fixture-1203.yaml
+++ b/fixtures/bugs/1203/fixture-1203.yaml
@@ -1,0 +1,25 @@
+swagger: "2.0"
+info:
+  title: "enum array items must generate a const block"
+  version: "0.0.1"
+paths:
+  "/":
+    get:
+      responses:
+        200:
+          description: ok
+          schema:
+            $ref: '#/definitions/Foo'
+
+definitions:
+  Foo:
+    type: object
+    properties:
+      roles:
+        type: array
+        items:
+          type: string
+          enum:
+            - Admin
+            - User
+            - SuperAdmin
--- a/generator/moreschemavalidation_fixtures_test.go
+++ b/generator/moreschemavalidation_fixtures_test.go
@@ -11768,3 +11768,18 @@ func initFixture3141() {
 			`.ContextValidate(ctx, formats); err != nil {`,
 		}, noLines, noLines)
 }
+
+func initFixture1203() {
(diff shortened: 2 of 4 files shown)


fredbi (approved):

fredbi:
@RubenPari  thanks for this fix. Could you please rebase?

fredbi (changes requested):
Looks all good to me. Please rebase - the model validation test suite is conflicting the previous merge

RubenPari:
Yes sure

fredbi (approved):

PR #3421, opened 2026-08-08
fix(generator): don't crash generating enum const block on null value

### LINKED ISSUE #1413: schemavalidator:421:65: executing "schemavalidator" at <.>: wrong type for value; expected string; got interface {}
## Problem statement

Not sure if this a problem with go-swagger or the swagger spec I'm trying to use.

```2018/02/09 20:37:57 rendering 1 templates for model AddLabel
2018/02/09 20:37:57 name field AddLabel
2018/02/09 20:37:57 package field models
2018/02/09 20:28:57 creating generated file "add_label.go" in "models" as definition
failed rendering template data for definition: template execution failed for template definition: template: schemavalidator:421:65: executing "schemavalidator" at <.>: wrong type for value; expected string; got interface {}```

## Swagger specification
https://quay.io/api/v1/discovery

## Steps to reproduce
`swagger generate client -f swagger.json -A client -c quay -a quay`
## Environment
swagger version: dev
go version: go version go1.9.3 linux/amd64
OS: VoidLinux

### PR DESCRIPTION
**Pull request by RubenPari**
## Summary

Fixes #1413.

A "string" enum whose value list also contains a `null` entry (typically from the tolerated-but-invalid `type: [string, null]` + `enum: [..., null]` construct some specs use to signal nullability) crashes template rendering entirely:

```
template: schemavalidator:981:96: executing "schemavalidator" at <.>: wrong type for value; expected string; got interface {}
```

`pascalize`/`cleanupEnumVariant`/`printf "%q"` all assume every `.Enum` element is a string. A `null` entry decodes to a `nil` interface{}, and Go's `text/template` refuses to pass that to functions expecting a string argument, aborting the whole model generation.

## Fix

In `generator/templates/schemavalidator.gotmpl`, the two places that build the `const (...)` block for a string enum (top-level schema and property-level) now skip `nil` entries when emitting consts. A Go typed string const has no way to represent JSON `null` alongside real string values anyway, so there's nothing meaningful to generate for that entry — the fix just stops it from aborting generation for the rest of the (valid) enum values. The `var .../init()`/enum-validation plumbing is untouched and already tolerates `null` via `json.Unmarshal`, so runtime enum validation behavior is unaffected.

## Test plan

- [x] New fixture `fixtures/bugs/1413/fixture-1413.yaml` reproducing the exact construct from the issue.
- [x] New `initFixture1413` in `generator/moreschemavalidation_fixtures_test.go`, wired into `TestMoreModelValidations`.
- [x] Verified the new test fails with the unpatched template, reproducing the exact panic message from the issue, and passes with the fix.
- [x] Manually generated the model and ran `go build` on the output to confirm the generated Go code compiles.

### DIFF
--- a/fixtures/bugs/1203/fixture-1203.yaml
+++ b/fixtures/bugs/1203/fixture-1203.yaml
@@ -1,0 +1,25 @@
+swagger: "2.0"
+info:
+  title: "enum array items must generate a const block"
+  version: "0.0.1"
+paths:
+  "/":
+    get:
+      responses:
+        200:
+          description: ok
+          schema:
+            $ref: '#/definitions/Foo'
+
+definitions:
+  Foo:
+    type: object
+    properties:
+      roles:
+        type: array
+        items:
+          type: string
+          enum:
+            - Admin
+            - User
+            - SuperAdmin
--- a/fixtures/bugs/1413/fixture-1413.yaml
+++ b/fixtures/bugs/1413/fixture-1413.yaml
@@ -1,0 +1,31 @@
+swagger: "2.0"
+info:
+  title: "enum mixing string values and null must not crash const generation"
+  version: "0.0.1"
+paths:
+  "/":
+    get:
+      responses:
+        200:
+          description: ok
+          schema:
+            $ref: '#/definitions/AddLabel'
+
+definitions:
+  AddLabel:
+    type: object
+    required:
+      - key
+      - value
+      - media_type
+    properties:
+      key:
+        type: string
+      value:
+        type: string
+      media_type:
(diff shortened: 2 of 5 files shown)


fredbi:
@RubenPari ok, the fix indeed closes the crashing issue. That's fair. The deeper underlying issues (no support
for an array type, and therefore the legit null value just being skipped, are real and not addressed here and that's understandable. Let's have it fix the crash. The rest would deserve a much deeper rework.

fredbi:
@RubenPari  hey. This is the next one to be merged. Please rebase again. Yeah these tests do not merge naturally...

fredbi (approved):

PR #3422, opened 2026-08-08
fix(generator): don't name formats/ctx params on no-op Validate/ContextValidate

### LINKED ISSUE #3112: Generated code get unused parameter: formats warning
## Problem statement

Generate code with go-swagger, get warning from vscode

image

## Swagger specification
image

## Steps to reproduce

swagger generate model -f <spec-file-name>

## Environment
swagger version:
```shell
version: v0.31.0
commit: (unknown, mod sum: "h1:H8eOYQnY2u7vNKWDNykv2xJP3pBhRG/R+SOCAmKrLlc=")
```
  
go version:
go version go1.21.4 darwin/arm64 
 
OS:   
macos

RubenPari:
## Summary

Fixes #3112.

Any model with zero validations - the common case for plain data structs - generates:

```go
func (m *Foo) Validate(formats strfmt.Registry) error {
	return nil
}
func (m *Foo) ContextValidate(ctx context.Context, formats strfmt.Registry) error {
	return nil
}
```

`gopls`/`staticcheck` flag `formats`/`ctx` as unused parameters here, which shows up as a persistent warning on essentially every generated model in a project. Three independent users reported hitting this.

## Fix

`runtime.Validatable`/`runtime.ContextValidatable` interface satisfaction only cares about parameter *types*, not names, so renaming the unused params to `_` in this no-op branch (`generator/templates/schema.gotmpl`) is a pure codegen change with no behavioral effect - just silences the warning.

Scope: this targets the guaranteed-safe no-op short-circuit only (used exactly when the model has no validations at all, i.e. `.HasValidations`/`.HasContextValidations` is false). The `schemavalidator` template's non-trivial `Validate()`/`ContextValidate()` bodies (used when there *are* validations) are untouched, since some of those don't literally reference `formats` in every statement either (e.g. a model whose only check is `minProperties`) but that's a separate, higher-risk change - not addressed here to keep this PR's blast radius contained to the reported, most common case.

## Test plan

- [x] New fixture `fixtures/bugs/3112/fixture-3112.yaml` (plain model, no validations).
- [x] New `initFixture3112` in `generator/moreschemavalidation_fixtures_test.go`, asserting both the new signatures and the absence of the old named-param ones.
- [x] Updated every existing fixture/unit-test expectation across `generator/moreschemavalidation_fixtures_test.go` and `generator/model_test.go` that asserted the old `Validate(formats strfmt.Registry) error {\n\treturn nil\n}` / `ContextValidate(ctx context.Context, formats strfmt.Registry) error {\n\treturn nil\n}` literal for no-validation models, to match the new generated output.
- [x] `go test ./generator/...` (full package, not just the schema-validation fixtures) green.
- [x] `go build ./...` and `go vet ./generator/...` clean.

### DIFF
--- a/fixtures/bugs/3112/fixture-3112.yaml
+++ b/fixtures/bugs/3112/fixture-3112.yaml
@@ -1,0 +1,21 @@
+swagger: "2.0"
+info:
+  title: "no-op Validate/ContextValidate must not declare unused formats/ctx params"
+  version: "0.0.1"
+paths:
+  "/":
+    get:
+      responses:
+        200:
+          description: ok
+          schema:
+            $ref: '#/definitions/Plain'
+
+definitions:
+  Plain:
+    type: object
+    properties:
+      name:
+        type: string
+      count:
+        type: integer
--- a/generator/model_test.go
+++ b/generator/model_test.go
@@ -274,7 +274,7 @@ func TestGenerateModel_Primitives(t *testing.T) {
 			tt.assertRender(&val, "type TheType "+exp+"\n  \n")
 			continue
 		}
-		tt.assertRender(&val, "type TheType "+exp+"\n  \n// Validate validates this the type\nfunc (o theType) Validate(formats strfmt.Registry) error {\n  return nil\n}\n// ContextValidate validates this the type based on context it is used \nfunc (o theType) ContextValidate(ctx context.Context, formats strfmt.Registry) error {\n  return nil\n}\n")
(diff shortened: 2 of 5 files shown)


fredbi:
@RubenPari ok this fixes a legit lint/govet issue in the generated code. Thanks

fredbi:
@RubenPari  now please rebase that one - next to be merged.

fredbi (changes requested):
please rebase

fredbi (approved):

fredbi:
Onboarded by #3431, which repaired merge issues. Thanks so much

PR #3423, opened 2026-08-08
feat(generator): add --no-default-omit-empty CLI flag

### LINKED ISSUE #2386: Way to not include omitempty by default?
## Problem statement

In https://github.com/go-swagger/go-swagger/pull/1189 support for adding `x-omitempty: false` to properties was added, so that the generated code for that particular property excludes `omitempty` in the `struct` defintion of the Model.

However, it's quite tedious having to add this to each and every property.

**Is there some way to disable `omitempty` globally when generating the code?**
Otherwise it would be nice it could be added.

RubenPari:
## Summary

Fixes #2386.

The only existing way to exclude a property from getting a default `omitempty` struct tag is `x-omitempty: false` on every single property - tedious for large models, as the reporter and maintainer both noted in the thread.

## Fix

Following the approach the maintainer outlined in the issue thread, added a `--no-default-omit-empty` flag (available to `generate model`/`server`/`client` via the shared `modelOptions` group) that flips the *default* used when `x-omitempty` is not set on a property: normally primitives default to `omitempty`, with this flag they default to not-`omitempty`. The per-property `x-omitempty` extension always wins either way - this only changes the fallback.

Wiring mirrors the existing `StrictAdditionalProperties` flag:
- `generator.GenOpts.NoDefaultOmitEmpty` (new field)
- threaded into `typeResolver` (constructor + `NewWithModelName` clone)
- consumed in `resolvedType.setIsEmptyOmitted`, which is the single function computing the default (`generator/types.go`)

## Test plan

- [x] New fixture `fixtures/bugs/2386/fixture-2386.yaml`: a model with two plain (non-required) properties, one required property, and one property with explicit `x-omitempty: true`.
- [x] New `initFixture2386` in `generator/moreschemavalidation_fixtures_test.go` with two runs - default flag (unchanged existing behavior) and `NoDefaultOmitEmpty: true` (asserting `omitempty` is dropped from the plain properties but the explicit `x-omitempty: true` property keeps it).
- [x] Verified the new test fails without the `types.go` changes (flag has no effect) and passes with them.
- [x] Manually generated a model with and without `--no-default-omit-empty` and diffed the struct tags to confirm the exact expected behavior.
- [x] `go test ./generator/... ./cmd/...` green.
- [x] `go build ./...`, `gofmt -l`, `go vet ./generator/...` clean.

### DIFF
--- a/cmd/swagger/commands/generate/model.go
+++ b/cmd/swagger/commands/generate/model.go
@@ -20,6 +20,7 @@ type modelOptions struct {
 	StructTags                 []string `description:"the struct tags to generate, repeat for multiple (defaults to json)"                                   long:"struct-tags"`
 	RootedErrorPath            bool     `description:"extends validation errors with the type name instead of an empty path, in the case of arrays and maps" long:"rooted-error-path"`
 	WithStringer               bool     `description:"generate a fmt.Stringer String() method on models, rendering field values as JSON (see issue #872)"    long:"with-stringer"`
+	NoDefaultOmitEmpty         bool     `description:"do not default to omitempty struct tags unless x-omitempty is explicitly set on a property (see issue #2386)" long:"no-default-omit-empty"`
 }
 
 func (mo modelOptions) apply(opts *generator.GenOpts) {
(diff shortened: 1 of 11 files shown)


fredbi (commented):
Nice addition. And long overdue... Thank you.
You'll probably need a rebase after the previous reviewed one is merged
(these moreschema* unfortunately don't merge well)

fredbi (approved):

fredbi (approved):

PR #3424, opened 2026-08-08
fix(generator): import strfmt for response headers with a strfmt format

### LINKED ISSUE #1769: Using format: uri leads to not compilable code
## Problem statement

If one uses a construct like 
```
type: string
format: uri
```
is a schema definition, the generated operation is missing `strfmt `import, hence not compilable

## Swagger specification
```
paths:
  /users:
    post:
      tags:
      - User
      summary: "an operation to register a new user"
      description: >-
        Registers a new user
      operationId: "registerUser"
      consumes:
      - "application/json"
      produces:
      - "application/json"
      parameters:
      - name: payload
        in: body
        description: >-
          Data for the user to be registered
        schema:
          $ref: "./Data.yaml"
        required: true
      responses:
        201:
          description: Created
          headers:
            Location:
              description: URI of the registered resource
              type: string
              **format: uri**
            ETag:
              description: >-
                Fingerprint of the created resource
              type: string
          schema:
            $ref: "./User.yaml"
```

Generated operation:
```
package user

// This file was generated by the swagger tool.
// Editing this file might prove futile when you re-run the swagger generate command

import (
	"net/http"

	"github.com/go-openapi/runtime"

	dto "github.wdf.sap.corp/bb-go/userservice/pkg/api/rest/gen/dto"
)

// RegisterUserCreatedCode is the HTTP code returned for type RegisterUserCreated
const RegisterUserCreatedCode int = 201

/*RegisterUserCreated Created

swagger:response registerUserCreated
*/
type RegisterUserCreated struct {
	/*Fingerprint of the created resource

	 */
	ETag string `json:"ETag"`
	/*URI of the registered resource

	 */
	Location **strfmt.URI** `json:"Location"`

	/*
	  In: Body
	*/
	Payload *dto.User `json:"body,omitempty"`
}

```
As you see the generated code uses `strfmt.URI`, while `strfmt` is not imported

## Steps to reproduce

Just generate...

## Environment
swagger version: 0.17.0 
go version: 1.11  
OS:  Windows 10 (Linux subsystem/Ubuntu 18.04)

RubenPari:
## Summary

Fixes #1769.

A response header declared with a strfmt-backed format (e.g. `type: string, format: uri`) renders as `strfmt.URI` in the generated `..._responses.go` file, but the `strfmt` import was never added - the generated operations package fails to compile with `undefined: strfmt`.

Root cause: headers are resolved via `simpleResolvedType()` directly into `GenHeader`/`GenItems`, which never goes through the schema-import collection (`findImports`) used for body/parameter schemas (that path works correctly, as noted in the issue thread - the bug is specific to headers).

## Fix

Added `codeGenOpBuilder.registerStrfmtImport(goType string)`, called from `MakeHeader` and `MakeHeaderItem` right after resolving the header's type. It registers the `strfmt` import on `b.Imports` whenever the resolved Go type has a `strfmt.` prefix. Scoped to headers only, since body/parameter schemas already get this correctly.

## Test plan

- [x] New `TestMakeHeader_StrfmtImport` in `generator/operation_test.go`, asserting `b.Imports["strfmt"]` is set after building a `format: uri` header.
- [x] Verified the new test fails without the fix and passes with it.
- [x] Manually generated a server from a spec with a `format: uri` response header and confirmed `strfmt` now appears in the generated file's imports.
- [x] `go build ./...`, `gofmt -l`, `go vet ./generator/...` clean.

### DIFF
--- a/generator/operation.go
+++ b/generator/operation.go
@@ -520,6 +520,18 @@ func (b *codeGenOpBuilder) MakeResponse(receiver, name string, isSuccess bool, r
 		}
 		res.Schema = &schema
 	}
+
+	if headersNeedStrfmt(res.Headers) {
+		// Register the strfmt import on a response-scoped copy of the imports
+		// map. Mutating the shared b.Imports would leak the import into the
+		// parameters template, which already declares strfmt on its own (#1769).
+		imports := maps.Clone(b.Imports)
+		if imports == nil {
+			imports = make(map[string]string)
+		}
+		imports["strfmt"] = "github.com/go-openapi/strfmt"
+		res.Imports = imports
+	}
 	return res, nil
 }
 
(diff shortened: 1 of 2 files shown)


fredbi (approved):
Looks good to me.

PR #3425, opened 2026-08-08
feat(generator): add --with-go-run to invoke swagger via go run in //go:generate

### LINKED ISSUE #3000: Best-practice `go:generate` command in generated server code
## Problem statement

Currently the server code which gets generated contains a line with hard-coded `//go:generate` part. Specifically it assumes that `swagger` is always pre-installed somewhere and available in the `$PATH`.

Managing tools like `go-swagger` as an external dependency however brings a number of challenges, not least one related to differences between platforms (OS/arch). This is why Go has an established common pattern for maintaining tools like these as regular dependencies via `tools.go`, see https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module for more. One possible consequence of this approach is that the tool is launched via `go run` rather than directly.

Meaning, that instead of

```go
//go:generate swagger generate ...
```
one would have
```go
//go:generate go run github.com/go-swagger/go-swagger/cmd/swagger generate ...
```

Currently, the server code only generates the top version. While I understand that the particular file involved is safe to edit I still think it would be helpful if it could be generated with best practices in the beginning, so it does not _have to_ be edited.

This could probably be achieved in two ways - either it may become another CLI flag of some kind, or it could just be detected automatically from `os.Args`.

## Swagger specification

```json
{
    "info": {
        "title": "Example",
        "version": "0.0.1"
    },
    "swagger": "2.0",
    "host": "example.com",
    "paths": {
        "/v1/products": {
            "get": {
                "summary": "List product names",
                "operationId": "listProducts",
                "responses": {
                    "200": {
                        "description": "OK",
                        "schema": {
                            "type": "array",
                            "items": {
                                "type": "string"
                            }
                        }
                    }
                }
            }
        }
    }
}
```

## Steps to reproduce

Generate server code

```sh
go run github.com/go-swagger/go-swagger/cmd/swagger generate server
```

Open `./restapi/configure_example.go`

```go
// ...

//go:generate swagger generate server --target ../../test --name Example --spec ../swagger.json --principal interface{}

// ...
```

## Environment

swagger version: v0.30.5
go version: go1.21.5
OS: darwin/arm64

RubenPari:
## Summary

Fixes #3000.

Generated server files hard-code `//go:generate swagger generate server ...`, assuming a `swagger` binary is pre-installed and available on `$PATH`. This doesn't fit the common Go pattern of tracking build tools as regular module dependencies (`tools.go`), where the tool would instead be invoked via `go run <module path>`.

## Fix

Added `GenOpts.WithGoRunGoGenerate` (wired to a new `--with-go-run` flag on `generate server`) and `GenOpts.GoGenerateCommand()`, used by `configureapi.gotmpl`/`autoconfigureapi.gotmpl` in place of the hard-coded `swagger` literal:

- default (unchanged): `//go:generate swagger generate server --target ...`
- with `--with-go-run`: `//go:generate go run github.com/go-swagger/go-swagger/cmd/swagger generate server --target ...`

Zero functional risk: this only changes a `//go:generate` comment, which is never compiled - just parsed by `go generate` when explicitly invoked.

## Test plan

- [x] New fixtures `fixtureServerGoRunGenerate3000`/`fixtureServerGoRunGenerateDefault3000` in `generator/generate_test.go`, asserting the generated `configure_*.go` content for both the default and `--with-go-run` cases.
- [x] Verified the new test fails without the fix (still emits the bare `swagger` command) and passes with it.
- [x] Manually generated a server with and without `--with-go-run` and diffed the `//go:generate` line.

### DIFF
--- a/generator/shared.go
+++ b/generator/shared.go
@@ -380,6 +380,23 @@ func (g *GenOpts) SpecPath() string {
 	return specRel
 }
 
+// GoGenerateCommand returns the command invoked by the //go:generate directive
+// emitted in generated server files.
+//
+// By default this is the bare "swagger" binary, which assumes it is
+// pre-installed and available on $PATH. When WithGoRunGoGenerate is set, the
+// tool is instead invoked via "go run", following the tools.go pattern for
+// tracking build tool dependencies (see issue #3000), so `go generate` works
+// without requiring a separately installed swagger binary.
+//
+// This method is used by templates, e.g. with {{ .GoGenerateCommand }}
+func (g *GenOpts) GoGenerateCommand() string {
+	if g.WithGoRunGoGenerate {
+		return "go run github.com/go-swagger/go-swagger/cmd/swagger"
+	}
+	return "swagger"
+}
+
 // titleOrDefault infers a name for the app from the title of the spec.
 func titleOrDefault(lang *language.Options, specDoc *loads.Document, name, defaultName string) string {
 	if strings.TrimSpace(name) == "" {
--- a/generator/shared.go
+++ b/generator/shared.go
@@ -389,7 +389,7 @@ func (g *GenOpts) SpecPath() string {
 // tracking build tool dependencies (see issue #3000), so `go generate` works
 // without requiring a separately installed swagger binary.
 //
-// This method is used by templates, e.g. with {{ .GoGenerateCommand }}
+// This method is used by templates, e.g. with {{ .GoGenerateCommand }}.
 func (g *GenOpts) GoGenerateCommand() string {
 	if g.WithGoRunGoGenerate {
 		return "go run github.com/go-swagger/go-swagger/cmd/swagger"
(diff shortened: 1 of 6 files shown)


fredbi (approved):
Looks good. Thank you for this addition.

generator/shared.go
```diff
@@ -380,6 +380,23 @@ func (g *GenOpts) SpecPath() string {
...
+// GoGenerateCommand returns the command invoked by the //go:generate directive
+// emitted in generated server files.
+//
+// By default this is the bare "swagger" binary, which assumes it is
+// pre-installed and available on $PATH. When WithGoRunGoGenerate is set, the
+// tool is instead invoked via "go run", following the tools.go pattern for
+// tracking build tool dependencies (see issue #3000), so `go generate` works
+// without requiring a separately installed swagger binary.
+//
+// This method is used by templates, e.g. with {{ .GoGenerateCommand }}
```
fredbi: Error: generator/shared.go:392:1: Comment should end in a period (godot)
  // This method is used by templates, e.g. with {{ .GoGenerateCommand }}
  ^
  1 issues:
  * godot: 1

fredbi (changes requested):
linter croaks on comment: sentence must end with dot.

fredbi:
@copilot resolve the merge conflicts in this pull request

fredbi (approved):
conflitcts resolved, linting nit fixed

PR #3426, opened 2026-08-08
fix(generator): explicitly register the yamlpc import for yaml producers/consumers

### LINKED ISSUE #1603: yamlpc is used but not imported
## Problem statement

When generate code using go-swagger against a spec that contains `produces: application/yaml`, the generated code `restapi/operations/my_service_api.go` and
# `restapi/configure_my_service.go` will use `yamlpc` without importing it. So, running `go build` will fail.

## Swagger specification

Any spec that contains `produces: application/yaml`.

## Steps to reproduce

`swagger generate -f <spec> -A my_service`

## Environment
swagger version: 0.14.0 
go version: 1.10.2  
OS:   alpine (3.7) & macos (10.13.5)

RubenPari:
## Summary

Fixes #1603.

`GenSerializer.Implementation` for the `yaml` media type is the raw code snippet `yamlpc.YAMLProducer()`/`yamlpc.YAMLConsumer()`, but nothing ever registered its import (`github.com/go-openapi/runtime/yamlpc`) anywhere in the generated server files. Compilation only succeeded when `goimports` happened to resolve it while scanning packages - several users in the issue thread reported this as flaky/non-deterministic depending on their environment (module cache state, GOPATH vs module mode, etc.), and one traced it specifically to `imports.Process`'s package-scanning behavior.

## Fix

Added `registerSerializerImports` (`generator/media.go`), called from `appGenerator.makeCodegenApp` for both the `consumes` and `produces` serializer groups, which adds the `yamlpc` import explicitly to `GenApp.Imports` whenever a serializer's implementation references it. This mirrors the always-explicit-never-implicit-via-goimports approach already used for the `strfmt` header import fix in #1769 / PR #3424.

## Test plan

- [x] New fixture `fixtures/bugs/1603/fixture-1603.yaml` (`consumes: application/x-yaml`).
- [x] New `fixtureServerYamlpcImport1603` in `generator/generate_test.go`, asserting the `yamlpc` import is present in both `restapi/configure_*.go` and `restapi/operations/*_api.go`.
- [x] Manually generated a server from the fixture spec and confirmed the import line is now present in both files.

**Note on test determinism:** because the original bug's trigger is inherently about `goimports` resolution being flaky depending on the environment, the new regression test cannot be shown to fail on unpatched code in this sandbox - `goimports` reliably resolves it here with a warm module cache. The fix is still correct and strictly more robust than relying on `goimports` for a known, well-defined import; the test asserts the explicit-import invariant going forward rather than reproducing the original flakiness.

### DIFF
--- a/generator/media.go
+++ b/generator/media.go
@@ -75,20 +75,22 @@ var knownConsumers = map[string]string{
 	multipartForm:  "runtime.ByteStreamConsumer()",
 }
 
-// registerSerializerImports adds explicit imports for well-known serializer implementations.
-//
-// This prevents code snippets such as "yamlpc.YAMLProducer()" to rely solely on goimports
-// to discover and resolve the import on the local host, which is not guaranteed (#1603).
+// registerSerializerImports adds explicit imports for well-known serializer
+// implementations that live outside the runtime package and are otherwise
+// never registered anywhere: they are referenced only as a raw code snippet
+// (e.g. "yamlpc.YAMLProducer()") in GenSerializer.Implementation, so nothing
+// ever adds their import - the generated file only compiled when goimports
+// happened to resolve it, which is not guaranteed (#1603).
 func registerSerializerImports(imports map[string]string, groups GenSerGroups) {
 	for _, group := range groups {
 		for _, ser := range group.AllSerializers {
 			if strings.HasPrefix(ser.Implementation, "yamlpc.") {
 				imports["yamlpc"] = "github.com/go-openapi/runtime/yamlpc"
 			}
-			// also register runtime import explicitly. It will be pruned if already registered with other defaults.
-			if strings.HasPrefix(ser.Implementation, "runtime.") {
-				imports["runtime"] = "github.com/go-openapi/runtime"
-			}
+			// follow-up(fredbi): we'd also like register runtime import explicitly here.
+			// It should be pruned if already registered with other defaults.
+			// Unfortunately, the way imports are currently resolved produces a duplicate import,
+			// and that issue should be fixed first.
 		}
 	}
 }
--- a/generator/media.go
+++ b/generator/media.go
@@ -75,17 +75,19 @@ var knownConsumers = map[string]string{
 	multipartForm:  "runtime.ByteStreamConsumer()",
 }
 
-// registerSerializerImports adds explicit imports for well-known serializer
-// implementations that live outside the runtime package and are otherwise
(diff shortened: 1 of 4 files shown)


fredbi:
There must be something else that explains the reported behavior. yamlpc import is not different than runtime.JSONProducer. Perhaps the fact that is is a subpackage in runtime (not a module).

That said, I like the idea of the explicit registration. We should just be consistent and add go-openapi/runtime for json

generator/media.go
```diff
@@ -75,6 +75,22 @@ var knownConsumers = map[string]string{
...
+// implementations that live outside the runtime package and are otherwise
+// never registered anywhere: they are referenced only as a raw code snippet
+// (e.g. "yamlpc.YAMLProducer()") in GenSerializer.Implementation, so nothing
+// ever adds their import - the generated file only compiled when goimports
+// happened to resolve it, which is not guaranteed (#1603).
+func registerSerializerImports(imports map[string]string, groups GenSerGroups) {
+	for _, group := range groups {
+		for _, ser := range group.AllSerializers {
+			if strings.HasPrefix(ser.Implementation, "yamlpc.") {
+				imports["yamlpc"] = "github.com/go-openapi/runtime/yamlpc"
```
fredbi, line 88: There must be something else that explains the reported behavior. yamlpc import is not different than runtime.JSONProducer. Perhaps the fact that is is a subpackage in runtime (not a module).

That said, I like the idea of the explicit registration. We should just be consistent and add go-openapi/runtime for json

fredbi: but it may be that we already have go-openapi/runtime by default. Not sure any longer.
Since this a key, it doesn' harm to be explicit here, so whatever other default stance we have on imports (or if the json producer moves to another package in runtime), we have a readable track here.

fredbi (changes requested):
Excellent idea to be explicit, even though it remains a bit unclear whether this solves the reported issue. We'll probably never know, since the information provided never allowed for a complete explanation. The explanation provided by this PR is rational, but may fail to capture to the actual cause. At any rate, the explicit import is good and should normally rule out any ambiguity with goimport.

And being explicit to the point we no longer need go import is precisely the goal I'd like to achieve. So good call overall.

fredbi (approved):
My request for change is moot, as this creates a cascade of other issues.
I left this other problem with import resolution as a follow-up, not directly tied
to the issue at hand. So no grudge held about the original PR and it can be merged.

PR #3427, opened 2026-08-08
fix(generator): fix nil-pointer unmarshal target for nullable additionalProperties/tuple items

### LINKED ISSUE #1632: Client unable to unmarshall `additionalProperties`
## Problem statement

I'm trying to use swagger to define [HAL structures](http://stateless.co/hal_specification.html) and having some problems with dynamic keys in the schemas, where the client library can not Unmarshal the additionalProperties when nested deep in a structure, as the values in `map[string]*HalHref` are null. When json.Unmarshall occurs, and the values for `additionalProperties` are unmarshalled an error is thrown:

```json
{
  "error": "json: Unmarshal(nil *models.HalHref)"
}
```

I can fix this by changing `var toadd *HalHref` to `toadd := &HalHref{}` in the `additionalProperties` functions. I tried overriding the template, but it's a protected template, so I can't override it.

If this is a bug, I'm happy to do a PR, but I'd be looking for guidance on how to write test cases around the *functionality* of the generated code, as I can't seem to find any at the moment.

My code looks like:

## Swagger specification

```yaml
definitions:
  hal-rsc-links:
    type: object
    description: Links to this resources and documentation for this resource
    properties:
      curies:
        type: array
        x-omitempty: true
        items:
          $ref: "#/definitions/hal-curie"
    additionalProperties:
      $ref: "#/definitions/hal-href"
  hal-href:
    type: object
    x-isnullable: true
    properties:
      href:
        type: string
  hal-curie:
    type: object
    properties:
      name:
        type: string
      href:
        type: string
        format: uri
      templated:
        type: boolean
```
https://github.com/getlunaform/lunaform/blob/6055306666424ced718514d90c0f27ea6e59665e/swagger.yml#L779

This generates a model like:

```go
// UnmarshalJSON unmarshals this object with additional properties from JSON
func (m *HalRscLinks) UnmarshalJSON(data []byte) error {
        //
	// ... Remove code for brevity
        //

	// stage 3, add additional properties values
	if len(stage2) > 0 {
		result := make(map[string]*HalHref)
		for k, v := range stage2 {
			var toadd *HalHref
			if err := json.Unmarshal(v, toadd); err != nil {
				return err
			}
			result[k] = toadd
		}
		m.HalRscLinks = result
	}

	return nil
}
```

and the generic resource parameters passed in from the client look like:

```go
// NewCreateModuleParams creates a new CreateModuleParams object
// with the default values initialized.
func NewCreateModuleParams() *CreateModuleParams {
	var ()
	return &CreateModuleParams{

		timeout: cr.DefaultTimeout,
	}
}
```

## Steps to reproduce

If possible, this can be reproduced in my project repo.

1. `cd $GO/src/getlunaform/lunaform`
2. `make generate`
3. `dep ensure`
4. `go run cmd/lunaform-server/main.go --port 8080 --scheme=http --api-key=dev-key`
5.  Generate a config file for the client
```bash
mkdir -p ~/.config/
cat > ~/.config/lunaform.yaml <<EOF
---
host: localhost
port: 8080
schemes:
  - http
log:
  level: error
apikey: dev-key
EOF
```
6. Run the command to get the error
```bash
go run cli/main.go terraform \
    module \
    create \
    --name tf-vpc  \
    --source github.com/drewsonne/tf-vpc \
    --type git
```

## Environment
swagger version: 0.15.0 - 84da4f998602ef20ebd60a576756996162a1ca66 
go version: 1.10.3 
OS:   darwin/amd64

RubenPari:
## Summary

Fixes #1632.

Generated `UnmarshalJSON` for `additionalProperties`/`additionalItems` whose value type resolves to a pointer (`IsNullable`, e.g. a `$ref` to a model) declared:

```go
var toadd *SomeModel
if err := json.Unmarshal(v, toadd); err != nil { ... }
```

Passing `toadd` (a nil pointer) directly to `json.Unmarshal` is an invalid unmarshal target - it always fails at runtime with `json: Unmarshal(nil *models.SomeModel)`, since `json.Unmarshal` needs the *address* of the variable to populate, not its current (nil) value.

The `{{if not .IsNullable}}&{{end}}` conditional in the templates had the logic backwards: value types correctly got `&toadd`, but pointer types - which need it just as much, so `json` can allocate through the pointer - did not.

## Fix

Fixed in all three templates that share this exact pattern:
- `generator/templates/serializers/additionalpropertiesserializer.gotmpl`
- `generator/templates/serializers/subtypeserializer.gotmpl` (the `allOf` variant)
- `generator/templates/serializers/tupleserializer.gotmpl` (`additionalItems`)

by always taking the address regardless of nullability - the address-of a local variable is what `json.Unmarshal`/`Decoder.Decode` always need, whatever the variable's own type is.

Also fixed `generator/model_test.go`'s `TestGenerateModel_TupleWithComplex`, which had (unintentionally) pinned the buggy `dec.Decode(toadd)` (missing `&`) as the expected generated output.

## Test plan

- [x] New fixture `fixtures/bugs/1632/fixture-1632.yaml` reproducing the exact spec shape from the issue (`additionalProperties: $ref`).
- [x] New `initFixture1632` in `generator/moreschemavalidation_fixtures_test.go`, asserting `json.Unmarshal(v, &toadd)`.
- [x] Verified the new test fails without the fix (asserts the buggy line is absent) and passes with it.
- [x] Wrote a standalone Go test against the actually-generated model package and confirmed: unpatched code panics with the exact `json: Unmarshal(nil *models.HalHref)` error from the issue; patched code unmarshals correctly and round-trips the value.
- [x] `go test ./generator/...` (full package) green - this also caught and fixed a pre-existing test (`TestGenerateModel_TupleWithComplex`) that had pinned the bug as expected behavior.


generator/templates/serializers/additionalpropertiesserializer.gotmpl
```diff
@@ -28,7 +28,7 @@ func ({{.ReceiverName}} *{{ .GoName }}) UnmarshalJSON(data []byte) error {
     result := make(map[string]{{ template "schemaType" .AdditionalProperties }})
     for k, v := range stage2 {
       var toadd {{ template "schemaType" .AdditionalProperties }}
```
fredbi, line 30: I have a reasonable doubt here: what if .AdditionalProperties _is_ nullable and rendered as a pointer?
and what if the value then that enters the function is nil?

just being a bit speculative here because it super hard to ascertain if we have or don't have a pointer type that is populated there. So in the new fixtures, this is a struct, so no pointer.

fredbi (changes requested):
need clarification about what happens if the AdditionalPoperty resolves as a pointer

generator/templates/serializers/additionalpropertiesserializer.gotmpl
RubenPari: Good catch, checked it end to end.

`.AdditionalProperties` being nullable is already handled upstream in `schemaType`: it prefixes the type with `*` when `.IsNullable` is true (see `schematype.gotmpl`), so `var toadd {{ template "schemaType" .AdditionalProperties }}` already declares `toadd` as e.g. `*HalHref`, not `HalHref`.

That's exactly why the old conditional `&` was the bug: when nullable, it skipped taking the address and passed `toadd` (a nil `*HalHref`) straight to `json.Unmarshal`. `encoding/json` requires a non-nil pointer target, so that panics/errors on any nullable additionalProperties value — issue #1632.

Taking `&toadd` unconditionally passes `**HalHref` instead. `encoding/json` unmarshals through arbitrary pointer depth correctly: it allocates the inner `*HalHref` as needed (or leaves it nil on JSON `null`), and the result stored in the map is still the expected single-level `*HalHref` — no double pointer leaks into the generated struct's field type.

Added a regression test (`TestGenModel_Issue1632` in `generator/model_test.go`) using the `1632` fixture, which has `additionalProperties: $ref: '#/definitions/HalHref'` (nullable, pointer-rendered). It asserts:
- `var toadd *HalHref` is generated (confirming the nullable/pointer case)
- `json.Unmarshal(v, &toadd)` is generated (address always taken)
- the field stays `map[string]*HalHref` (no double pointer)

Full generator test suite passes locally with this change. Pushed.

fredbi: ok thanks going to review it again today. As a rule of thumb, all things pointer-related are **hard** in go-swagger. We suffer from a decade old design issue there. So patching is fine but some issues that I've labeled "future/maybe" are likely hitting this architecture wall and would need a major change (which I am working on aside).

fredbi (approved):
ok @RubenPari got your explanations. Going to merge this: this is good enough for the pursued objective.
I'll have a few more edge cases to cover in the same zone, and will add those as a follow-up PR.

fredbi:
thanks  @RubenPari. Now for the follow-up (for discriminated types, "any" types, etc...)

PR #3433, opened 2026-08-10
fix(generator): guard against nil params being passed into client method

### LINKED ISSUE #3432: nil pointer deref in client template
We have users who have been passing `nil` as input to a client's method (relying on the behavior of previous code which generated the params quietly under the covers; now since https://github.com/go-swagger/go-swagger/commit/01d8f1358f8ab94c4b7c3a1caf50e1d0b11b26c6 this will panic on a nil pointer deref.

https://github.com/go-swagger/go-swagger/blob/bd444bd687d42b599a6bd17f488b4267b0c694a3/generator/templates/client/client.gotmpl#L147

i think this just needs a simple guard `if params != nil && params.inner.ctx != nil {` added, since in the fallthrough to the `Context()` method they do get recreated as before.

bobcallaway:
Fixes: #3432

### DIFF
--- a/generator/client_test.go
+++ b/generator/client_test.go
@@ -1044,6 +1044,40 @@ func TestGenClient_2773(t *testing.T) {
 	})
 }
 
+func TestGenClient_3432(t *testing.T) {
+	t.Parallel()
+	defer discardOutput()()
+
+	root := t.TempDir()
+	opts := testClientGenOpts()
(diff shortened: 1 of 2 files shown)


cpanato (approved):

fredbi (approved):

fredbi:
Thanks @bobcallaway

PR #3471, opened 2026-08-18
Generate optional getters for all fields

pyrohedgehog:
the --generate-getters param lets us generate getters on all fields. That way interfaces can be written to accept multiple generated models.

Proposal mentioned in Discord and [this examples PR](https://github.com/go-swagger/examples/pull/94)


fredbi (commented):
Looks good to me. Nice execution. I am not sure that the "if not .IsBaseType" covers all the edge cases one might imagine, but i think this pretty good for a first version.
It remains extremely difficult to anticipate all the cases at hand in the combinatorial nightmare of jsonschema.

fredbi (approved):
thank you

PR #3474, opened 2026-08-22
fix(generator): handle discriminator allOf models

### LINKED ISSUE #3133: Generating incorrcect model with discriminator , x-discriminator-value and all-of
While generating a model from a JSON with polymorphic types, Swagger generates an incorrect model, with an empty definition of the Set method, and  empty Validate and ContextValidate functions. Additionally, the struct contains incorrect field name "nameField" , it should be "name". Why it is generating struct and interface both?

## Swagger specification
"swagger": "2.0",
"definitions": {
  "Datasource": {
    "type": "object",
    "example": "isResource",
    "required": [
        "collectorAttribute",
    ],
"properties": {
"collectorAttribute": {
            "description": "Data collector's attributes to collector data. e.g. a ping data source has a ping collector attribute. \n PingCollectorAttributeV1 has two fields. the ip to ping, the data size send to ping",
            "$ref": "#/definitions/CollectorAttribute"
        }
   }
}
"CollectorAttribute": {
    "type": "object",
    "required": [
        "name"
    ],
    "properties": {
        "name": {
            "description": "The collector name",
            "type": "string"
        }
    },
    "discriminator": "name"
},
"SNMPCollectorAttribute": {
            "allOf": [
                {
                    "$ref": "#/definitions/CollectorAttribute"
                },
                {
                    "type": "object",
                    "properties": {
                        "ip": {
                            "readOnly": true,
                            "type": "string"
                        }
                    }
                }
            ],
            "x-discriminator-value": "snmp"
        },
"ScriptCollectorAttribute": {
            "allOf": [
                {
                    "$ref": "#/definitions/CollectorAttribute"
                },
                {
                    "type": "object",
                    "properties": {
                        "groovyScript": {
                            "description": "groovy script",
                            "type": "string"
                        },
                        "scriptType": {
                            "description": "script type: embed | file | powershell",
                            "type": "string"
                        },
                        "windowsScript": {
                            "description": "windows script",
                            "type": "string"
                        },
                        "linuxCmdline": {
                            "description": "linux script command line",
                            "type": "string"
                        },
                        "linuxScript": {
                            "description": "linux script",
                            "type": "string"
                        },
                        "windowsCmdline": {
                            "description": "windows script command line",
                            "type": "string"
                        },
                        "properties": {
                            "description": "property based",
                            "type": "string"
                        }
                    }
                }
            ],
            "x-discriminator-value": "script"
        }
}
## Steps to reproduce
type CollectorAttribute interface {
	runtime.Validatable
	runtime.ContextValidatable

	// The data collector's name
	// Required: true
	Name() string
	SetName(string)

	// AdditionalProperties in base type shoud be handled just like regular properties
	// At this moment, the base type property is pushed down to the subtype
}

type collectorAttribute struct {
	nameField string
}

// Name gets the name of this polymorphic type
func (m *collectorAttribute) Name() string {
	return "CollectorAttribute"
}

// SetName sets the name of this polymorphic type
func (m *collectorAttribute) SetName(val string) {
}

// UnmarshalCollectorAttributeSlice unmarshals polymorphic slices of CollectorAttribute
func UnmarshalCollectorAttributeSlice(reader io.Reader, consumer runtime.Consumer) ([]CollectorAttribute, error) {
	var elements []json.RawMessage
	if err := consumer.Consume(reader, &elements); err != nil {
		return nil, err
	}

	var result []CollectorAttribute
	for _, element := range elements {
		obj, err := unmarshalCollectorAttribute(element, consumer)
		if err != nil {
			return nil, err
		}
		result = append(result, obj)
	}
	return result, nil
}

// UnmarshalCollectorAttribute unmarshals polymorphic CollectorAttribute
func UnmarshalCollectorAttribute(reader io.Reader, consumer runtime.Consumer) (CollectorAttribute, error) {
	// we need to read this twice, so first into a buffer
	data, err := ioutil.ReadAll(reader)
	if err != nil {
		return nil, err
	}
	return unmarshalCollectorAttribute(data, consumer)
}

func unmarshalCollectorAttribute(data []byte, consumer runtime.Consumer) (CollectorAttribute, error) {
	buf := bytes.NewBuffer(data)
	buf2 := bytes.NewBuffer(data)

	// the first time this is read is to fetch the value of the name property.
	var getType struct {
		Name string `json:"name"`
	}
	if err := consumer.Consume(buf, &getType); err != nil {
		return nil, err
	}

	if err := validate.RequiredString("name", "body", getType.Name); err != nil {
		return nil, err
	}

	// The value of name is used to determine which type to create and unmarshal the data into
	switch getType.Name {
	case "CollectorAttribute":
		var result collectorAttribute
		if err := consumer.Consume(buf2, &result); err != nil {
			return nil, err
		}
		return &result, nil
	case "SNMPCollectorAttribute":
		var result SNMPCollectorAttribute
		if err := consumer.Consume(buf2, &result); err != nil {
			return nil, err
		}
		return &result, nil
	case "ScriptCollectorAttribute":
		var result ScriptCollectorAttribute
		if err := consumer.Consume(buf2, &result); err != nil {
			return nil, err
		}
		return &result, nil
	}
	return nil, errors.New(422, "invalid name value: %q", getType.Name)
}

// Validate validates this collector attribute
func (m *collectorAttribute) Validate(formats strfmt.Registry) error {
	return nil
}

// ContextValidate validates this collector attribute based on context it is used
func (m *collectorAttribute) ContextValidate(ctx context.Context, formats strfmt.Registry) error {
	return nil
}
## Environment
swagger version: v0.29.0
go version: go1.20.3 darwin/arm64

014-code:
## Summary

Fix model generation for polymorphic schemas using `discriminator`, `allOf`, and `x-discriminator-value`.

The generated code previously used inconsistent field names between the model body and discriminated serializers, producing code that failed to compile. This change also honors explicit `x-discriminator-value` values while preserving existing fallbacks.

## Related Issue

Fixes #3133

## Tests

- `TestBuildDiscriminatorMap_UsesDiscriminatorValue`
- `TestGenerateModel_DiscriminatorAllOfBaseProperties`
- Existing discriminator and allOf regression tests
- Generated #3133 reproduction package: `go build ./models`

All focused tests and the generated reproduction build pass.

The full `go test ./generator` run still reports unrelated existing failures involving `TestGenParameter_Issue909` and dependency resolution in temporary generated modules.


fredbi:
Thank you for this proposal. Looking into it today. Could please rebase with the merge conflict resolved? thanks

014-code:
> Thank you for this proposal. Looking into it today. Could please rebase with the merge conflict resolved? thanks

Thank you, I have resolved the merge conflicts

fredbi (approved):
This looks good. Thank you for this fix.

Issue #3485, opened 2026-09-03
Params not supported by --generate-getters


pyrohedgehog:
For some reason, whoever wrote the `--generate-getters` flag, forgot to include the parameters, which is the main helpful case! 
It'd be great if that could be supported, so things like common query flags could be parsed in a common way. EG, for OData support, accepting any param that meets: 
```go
type ODataPagination interface{
    GetDollarTop() int64
    GetDollarSkip() int64
}
```

fredbi:
hey. If I remember correctly the new feature is for schemas. So that should normally work for parameters in body. Isn't it the case?

pyrohedgehog:
yeah, but for things like OData query string params it doesn't. I forgot to check their generation. I'll raise a PR later today

pyrohedgehog:
https://github.com/go-swagger/go-swagger/pull/3491

PR #3491, opened 2026-09-04
fix:#3485 support params with getters generation.

pyrohedgehog:
does what it says on the tin. I updated the test model to highlight what this will likely get used for, common OData parsing. Now you can have getters on the params as well.


fredbi (approved):
Looks good. Thanks or this addition

Issue #3278, opened 2025-12-29
Potential Bug: Context validation early return in generated code


pavece:
## Problem statement

I believe there may be a bug in the generated `ContextValidate` methods where validation appears to stop early when encountering zero values in arrays, potentially preventing validation of subsequent array elements. While I'm not entirely certain this is unintended behavior, it seems problematic to me and I'd like to get your feedback.

### Early Return Instead of Continue

When generating the code for an API with a similar request / response schema that the one I provided under the 'Swagger specification & steps to reproduce' section, the generated code for the `contextValidateOrders` method includes the following code:

```go
func (m *BatchOrderRequest) contextValidateOrders(ctx context.Context, formats strfmt.Registry) error {
    for i := 0; i < len(m.Orders); i++ {
        if swag.IsZero(m.Orders[i]) { // not required
            return nil  // BUG: Shouldn't this be continue?
        }

        if err := m.Orders[i].ContextValidate(ctx, formats); err != nil {
            // ...
        }
    }
    return nil
}
```

In my opinion, this appears to be problematic because it seems to lead to some items not getting context validated when an empty item is present in the array. However, I'm not certain if this is the intended behavior or if there's a specific reason for this design choice.

## Swagger specification & steps to reproduce

Generate client for this swagger specification. Go into `batch_order_request.go` and `batch_order_response.go` and look for the `contextValidateOrders` method.

```yaml
swagger: "2.0"
info:
  title: Orders API
  version: 1.0.0

host: api.example.com
basePath: /v1
schemes:
  - https

paths:
  /orders:
    post:
      operationId: createOrders
      summary: "Create multiple orders"
      parameters:
        - name: body
          in: body
          required: true
          schema:
            $ref: "#/definitions/BatchOrderRequest"
      responses:
        201:
          description: Created
          schema:
            $ref: "#/definitions/BatchOrderResponse"

definitions:
  BatchOrderRequest:
    type: object
    required:
      - orders
    properties:
      orders:
        type: array
        items:
          $ref: "#/definitions/Order"

  Order:
    type: object
    x-nullable: false
    required:
      - sku
      - quantity
    properties:
      id:
        type: string
        readOnly: true
      sku:
        type: string
        minLength: 1
      quantity:
        type: integer
        minimum: 1
      status:
        type: string
        readOnly: true
      details:
        $ref: "#/definitions/OrderDetails"

  OrderDetails:
    type: object
    properties:
      notes:
        type: string

[...]

PR #3495, opened 2026-09-10
fix(generator): continue array context validation after zero item

ghost:
## Summary

Fixes the collection-item `ContextValidate` early-return regression reported by @pavece in #3278.

The regression was introduced in [`a4b8a2de`](https://github.com/go-swagger/go-swagger/commit/a4b8a2de152c56ed58d299d402a4a8e523689035) (`fix: ensure that the expressionValue isn't Zero before context validation`) and first released in `v0.30.5`. `v0.30.4` is unaffected; the bug remains present in current releases.

For an optional zero-valued object in an array or map, generated code returns from the entire collection-validation helper:

```go
if typeutils.IsZero(m.Orders[i]) { // not required
    return nil
}
```

This skips validation of every following item. Generate `continue` for collection elements instead, while retaining `return nil` for ordinary optional object properties.

## Regression test

Adds a generated-server consumer test based on the #3278 Orders API example. It validates this generated request model in request context:

```go
Orders: []Order{
    {},
    {ID: "client-supplied value"},
}
```

The first order is an allowed zero-valued array item. The second supplies the read-only `ID` field and must be rejected. Before this fix, the first item caused an early return and the invalid second item was accepted.

## Validation

```sh
golangci-lint run --new-from-rev master
go test ./...
```

Both pass.


fredbi:
Good catch. Thank you. Let's see now in details how it goes

fredbi (approved):
Looks good. Thanks

